PR #2, opened 2026-06-03
chore: stop caching .lake in CI; rely on Mathlib's cache server

kim-em:
This PR sets `use-github-cache: false` on `lean-action` in the CI workflow. By default `lean-action` wraps the ~2.4 GB `./.lake` directory in `actions/cache`; in our last run the save and restore of that archive cost roughly 40s combined, far more than the ~6s `lake exe cache get` it is meant to save. The Mathlib cache server (`use-mathlib-cache: true`) still supplies the prebuilt oleans, so nothing is compiled from Mathlib, this only drops the redundant GitHub-side caching. Expected effect: the build step falls from ~1m45s to roughly ~45s.

🤖 Prepared with Claude Code


kim-em:
Closing: this breaks the build (`error: ... bad imports`). `lake exe cache get` only fetches Mathlib's own oleans; the non-Mathlib dependency oleans (batteries, aesop, proofwidgets, Qq, …) were being supplied by the `actions/cache` of `./.lake` that this disabled, so the build can no longer resolve Mathlib's imports. The `.lake` cache is load-bearing. Leaving CI as-is.

PR #163, opened 2026-06-15
feat: bump-guard — machine-validate forward-only Lake-pin / toolchain bumps

kim-em:
This PR adds a trusted CI guard, `bump-guard`, so that a PR which changes only `TauCeti/` plus the two Lake *pins* (`lake-manifest.json` and `lean-toolchain`) can be sandbox-built and — once review approves — auto-merged without a human, while the lakefile itself stays human-owned. Today every Lake-config change is routed to a human; this makes routine forward Mathlib/toolchain bumps machine-validated instead.

`Scripts/check-bump.sh` reduces the whole manifest to one machine-checkable fact: Mathlib moved *forward* on the branch nominated in `lakefile.toml` (checked via the GitHub compare API — the new rev descends from the old one and is on the nominated branch). It then ties every transitive pin to Mathlib's own `lake-manifest.json` at that rev, so a dependency cannot be re-pointed independently of the trusted Mathlib, and requires `lean-toolchain` to move monotonically forward on the `leanprover/lean4` channel and to match Mathlib's toolchain. It executes none of the PR's code — it only reads two text files and queries the trusted Mathlib repo — and it fails outright if `lakefile.toml`/`lakefile.lean` changed.

`.github/workflows/bump-guard.yml` runs that validator in a trusted, base-defined `pull_request_target` workflow and posts a `bump-guard` commit status (a PR touching neither pin file passes trivially, so the check is uniformly present). `pr-build.yml` now permits the two pin files, validates them inline with the same trusted validator — run inline rather than by reading bump-guard's status, to avoid a race with that concurrently-running workflow — overlays the validated pins onto the trusted base, and builds against them under landrun. An invalid or backward bump still sets `INFRA=1` and is routed to a human. `AGENTS.md` documents the forward-only exception.

This is the foundational piece of a three-repo change; the matching TauCetiReview change (let the auto-merge arbiter accept PRs touching only `TauCeti/` + these two pins) and the TauCetiWorker change (a proactive bump task) follow in their own PRs. [FormalFrontier/TauCeti#162](https://github.com/FormalFrontier/TauCeti/pull/162) is a real forward bump that validates green under `check-bump.sh` locally and is the natural first end-to-end test.

🤖 Prepared with Claude Code

### DIFF
--- a/.github/workflows/bump-guard.yml
+++ b/.github/workflows/bump-guard.yml
@@ -1,0 +1,86 @@
+name: bump-guard
+
+# Validates that a PR's change to lake-manifest.json / lean-toolchain is a safe,
+# machine-checkable *forward* bump (mathlib moved forward on the branch nominated
+# in lakefile.toml; transitive pins derived from mathlib; toolchain monotonic and
+# consistent), and posts a `bump-guard` commit status. This status is what lets
+# pr-build build, and the review auto-merge accept, a manifest/toolchain-only PR
+# without a human — see Scripts/check-bump.sh for the full contract.
+#
+# Runs in a TRUSTED, base-defined workflow (pull_request_target): the validator is
+# always the base's Scripts/check-bump.sh, run against the base config and the PR's
+# (text-only) lean-toolchain/lake-manifest.json/lakefile. It executes NONE of the
+# PR's code — only reads two files and queries the trusted mathlib repo via gh api.
+# A PR that changes neither file passes trivially, so the check is uniformly present.
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+  workflow_dispatch:
+    inputs:
+      pr: { description: "PR number to validate", required: true, type: string }
+
+permissions:
+  contents: read
+  statuses: write
+
+concurrency:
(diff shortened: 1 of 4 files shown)


kim-em:
Got a second opinion from Codex before merging. It flagged three genuine structural gaps in `check-bump.sh`, now fixed in ff2c8e7:
- **Critical:** the validator only iterated git packages, so a non-git (e.g. `path`) dep could slip through unchecked. Now every PR package must be git and the package set (minus mathlib) must *exactly* equal mathlib's own manifest at the new rev — field for field, both directions.
- **High:** duplicate package names (incl. a second `mathlib`) are now rejected.
- **High:** every git `rev` must be an immutable 40-hex SHA, so `rev: master` (a moving ref) can't pass the compare checks then resolve a different commit at build time.
- **Medium:** toolchain failures now emit the `::error::bump-guard` annotation.

Validator now passes 9/9 unit cases (3 valid, 6 adversarial incl. the new ones). The one accepted item: `lake exe cache get` runs the new Mathlib's cache exe unsandboxed — but bump-guard proves it's a forward commit on Mathlib's own `master` (never a fork), the step carries no token, and this matches the existing base behavior under the track-master policy; documented inline. The red `build` check here is the intended infra → human-review signal.

PR #170, opened 2026-06-15
feat: auto-sync the TauCeti.lean import aggregator (mk_all)

kim-em:
This PR keeps the root aggregator `TauCeti.lean` importing every `TauCeti/` module in sorted order, automatically, so nobody has to hand-maintain it — it had already drifted (8 modules missing from the list, plus a couple of mis-orderings), all fixed here.

`Scripts/mk_all.sh` is a tiny self-contained version of Mathlib's `lake exe mk_all`: it regenerates `TauCeti.lean` (a fixed doc header followed by every module under `TauCeti/`, `LC_ALL=C`-sorted) and has a `--check` mode. I evaluated reusing `lake exe mk_all` directly (TauCeti gets it transitively from mathlib), but it rewrites the whole file — dropping our doc header — and requires building the mathlib executable in CI; the script avoids both and runs no module code (it only lists filenames).

`.github/workflows/mk-all.yml` runs the **trusted base** script over each PR's files on `pull_request_target`. If the aggregator drifted: a **same-repo** PR (the worker's branches) gets the fix auto-committed and pushed; a **fork** PR fails the `mk-all` status check with instructions, since we can't push to a fork. The push uses the GitHub App token rather than `GITHUB_TOKEN` (whose pushes don't re-trigger CI), so the corrected head gets a fresh build/review and the next `mk-all` run finds it in sync and stops — no loop. `ci.yml` runs `--check` on `main` as a backstop against direct pushes.

Touches `Scripts/` and `.github/` (human-owned), so it needs a human review.

🤖 Prepared with Claude Code


kim-em:
Got a second opinion from Codex (this is a write-capable `pull_request_target` workflow, so worth the scrutiny). It found a real **Critical** plus several lower issues, all fixed in 49e4d0c:

- **Critical — script injection:** PR-controlled `head.ref`/`head.repo.full_name` were spliced into `run:` shell via `${{ }}` (a crafted branch name could run code in the privileged job). They now flow only through `env`, are validated (sha 40-hex, repo + branch allowlists, `git check-ref-format`, no `refs/` prefix), and only validated forms are reused.
- **High — symlink write:** refuse to write through a symlinked `TauCeti.lean` (temp + rename; `find -type f` already skips symlinked modules).
- **Medium:** reject newline/space/metacharacter module paths (NUL-delimited + allowlist) so a crafted filename can't inject an `import` line; push to the exact `refs/heads/$REF` via checkout-managed credentials (no token in any URL); `set -euo pipefail` + atomic write; post the status to the new head SHA after an auto-commit.

Verified locally: regenerate/idempotence/`--check` pass+fail, and the symlink / space / newline cases are all rejected. Things Codex confirmed fine: the workflow runs the trusted base script (not the PR's), lists filenames only, stages only `TauCeti.lean`, and uses a non-force push.

PR #174, opened 2026-06-16
fix: post mk-all status with GITHUB_TOKEN, not the App token

kim-em:
This PR fixes the `sync` (mk-all) check failing on essentially every same-repo PR even when the `TauCeti.lean` aggregator is already in sync. The job regenerates the aggregator and then posts a commit status, but it posted that status with the GitHub App token, which lacks `statuses: write`; the POST returned `403 Resource not accessible by integration`, and because the runner executes the step under `bash -e`, that 403 failed the whole job — independently of whether the aggregator had actually drifted.

The diagnosis on a recent run: `SAME: 1` (correctly detected as same-repo), the `git diff` was empty (in sync), and the only error was the 403 on `POST /repos/.../statuses`. So contributors were seeing a red required check for a file that was correct.

The fix posts the status with `GITHUB_TOKEN`, which the workflow already grants `statuses: write`, and makes the post non-fatal so a status hiccup can never red an otherwise-correct sync. The auto-sync git push is untouched: it still uses the GitHub App token via the checkout-managed credentials (whose pushes re-trigger CI), since a status POST never triggers CI and so never needed the App token. Net effect: an in-sync same-repo PR now goes green, and a drifted one auto-commits, pushes, and reports success as originally intended — with no manual `Scripts/mk_all.sh` step required of contributors.

🤖 Prepared with Claude Code

### DIFF
--- a/.github/workflows/mk-all.yml
+++ b/.github/workflows/mk-all.yml
@@ -7,6 +7,11 @@ name: mk-all
 #   - fork PR              -> fail the `mk-all` check with instructions (we can't push to a fork).
 # The push uses the GitHub App token (not GITHUB_TOKEN, whose pushes don't re-trigger CI), so
 # the corrected head gets a fresh build/review; the next mk-all run finds it in sync and stops.
+# The commit-status POST, by contrast, uses GITHUB_TOKEN: it already carries `statuses: write`
+# here, whereas the App installation does not, so posting with the App token 403s ("Resource not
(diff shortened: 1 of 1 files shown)


kim-em:
Note for review: this `sync` check on **this** PR will itself stay red. `pull_request_target` workflows always run from the base branch (`main`), so #174 is still evaluated by the old, buggy `mk-all.yml` — the fix only takes effect for PRs opened/updated **after** it lands on `main`. The red `sync` here is the very bug being fixed, not a regression; build/guard pass.

PR #692, opened 2026-07-03
feat(Analysis/Contour): model-sector winding numbers

CBirkbeck:
## What

Discharges the **Layer 1 model-sector winding-number targets** from the contour-integration roadmap (`TauCetiRoadmap/ContourIntegration/Targets.lean`, lines 117–148), stated **verbatim** on the raw `circleMap` interval integral per the roadmap's function-based convention:

| theorem | arc | value |
|---|---|---|
| `windingNumber_modelSector` | opening angle `α` | `α / 2π` |
| `windingNumber_modelSector_pi` | `α = π` (semicircle) | `½` |
| `windingNumber_modelSector_pi_div_three` | `α = π/3` | `1/6` |

## Math

For a counterclockwise arc `γ θ = z₀ + r·e^{iθ}` about its centre `z₀` (Hungerbühler–Wasem, arXiv:1808.00997 §2, (2.4)), the centre is *off* the arc, so the generalized winding number is an ordinary interval integral whose integrand `γ̇ / (γ − z₀)` is constantly `i`; the value is then elementary.

## Two notes for reviewers

- **Full circle.** The roadmap's fourth Layer 1 target, `n_c(circle) = 1` (the `α = 2π` case), is *subsumed* by the general theorem and recorded in its docstring rather than added as a separate public lemma — as a standalone theorem it would only restate `circleIntegral.integral_sub_center_inv` in normalized form.
- **Naming.** The specializations are named by their opening angle (`_pi`, `_pi_div_three`) — a conclusion-based name — rather than the roadmap's motivating fundamental-domain points (`_at_i`, `_at_rho`); the statements are generic arbitrary-centre arc computations with no `i` or `ρ` in them.

## Provenance

Adapted from the AINTLIB `LeanModularForms` project (sector geometry of `ForMathlib/HungerbuhlerWasem/Crossing.lean`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #695, opened 2026-07-04
feat(Analysis/Contour): generalized winding number definition

CBirkbeck:
## What

Defines the **generalized winding number** (Hungerbühler–Wasem Def 2.1) — the core Layer-0 object of the contour-integration roadmap (`TauCetiRoadmap/ContourIntegration/Targets.lean:77`):

`windingNumber (γ : ℝ → ℂ) (a b : ℝ) (z₀ : ℂ) = (2πi)⁻¹ ∫ₐᵇ γ'(t) / (γ(t) − z₀) dt`

together with `IsNullHomologous γ a b Ω` (winding zero about every point outside `Ω`), the hypothesis of the homology Cauchy theorem and of HW Thm 3.3.

## Notes

- **Closedness of `γ` is not part of the definition** — it is carried on the theorems for which this value is a genuine winding number (integer index off the curve, geometric angle `α/2π` on it).
- `windingNumber` is the plain interval integral; the Cauchy principal value (for a singularity *on* the contour) is layered on separately via `HasCauchyPV`. For a circular arc about its centre (centre off the arc) the integral converges and equals `α/2π`, matching the model-sector values in #692 — same `(2πi)⁻¹` normalization.

## Provenance

Adapted from the AINTLIB `LeanModularForms` project.

### DIFF
--- a/TauCeti/Analysis/Contour/WindingNumber.lean
+++ b/TauCeti/Analysis/Contour/WindingNumber.lean
@@ -1,0 +1,63 @@
+/-
+Copyright (c) 2026 The Tau Ceti contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Chris Birkbeck
+-/
+module
+
+public import Mathlib.MeasureTheory.Integral.IntervalIntegral.Basic
+public import Mathlib.Analysis.Calculus.Deriv.Basic
+public import Mathlib.Analysis.SpecialFunctions.Trigonometric.Basic
+
+/-!
+# The generalized winding number (Hungerbühler–Wasem Def 2.1)
+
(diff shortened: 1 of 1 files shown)


CBirkbeck:
Closing: the local rubric review correctly flagged (correctness) that this definition is too weak. The roadmap's `windingNumber` is the *principal value* `(2πi)⁻¹·PV ∮_γ (z−z₀)⁻¹ dz` — giving the geometric angle α/2π when z₀ lies *on* the curve — whereas the plain interval integral used here loses that content at on-contour points. It depends on the Cauchy-principal-value construction (`HasCauchyPV`/`cauchyPV`), which I'll land first and then rebuild `windingNumber` on top of. Superseding PR to follow.

PR #696, opened 2026-07-04
feat(Analysis/Contour): residue at a point (simple-pole form)

CBirkbeck:
## What

Defines the **residue** of `f : ℂ → ℂ` at `z₀` (Hungerbühler–Wasem) — the Layer-0 object at `TauCetiRoadmap/ContourIntegration/Targets.lean:82`:

`residue f z₀ = lim_{z → z₀} (z − z₀) · f z`

the order-`(−1)` Laurent coefficient in its simple-pole form.

## Design note (flagging scope explicitly)

The roadmap phrases the target as the general "order-`(−1)` Laurent coefficient." Mathlib provides **no general residue** — `meromorphicTrailingCoeffAt` is the *leading* Laurent coefficient (equal to the residue only when the pole order is exactly `−1`), and there is no arbitrary-Laurent-coefficient API. I used the roadmap's own simple-pole characterization `lim_{z→z₀} (z − z₀) · f z`, which is:

- **correct for every simple pole** (the case HW 3.3 and the valence formula use) and for `f` analytic at `z₀` (residue `0`);
- a junk value for poles of order `> 1` (via `limUnder`).

It avoids a bespoke order-of-vanishing notion, deferring order theory to Mathlib's `meromorphicOrderAt`. If the general Laurent coefficient is wanted, that is real extra machinery — happy to extend it, but the simple-pole form suffices for every roadmap target that consumes `residue`.

## Provenance

Adapted from the AINTLIB `LeanModularForms` project.

### DIFF
--- a/TauCeti/Analysis/Contour/Residue.lean
+++ b/TauCeti/Analysis/Contour/Residue.lean
@@ -1,0 +1,51 @@
+/-
+Copyright (c) 2026 The Tau Ceti contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Chris Birkbeck
+-/
+module
+
+public import Mathlib.Analysis.Complex.Basic
+
+/-!
+# The residue of a function at a point
+
+For `f : ℂ → ℂ` with an isolated singularity at `z₀`, the **residue** is the order-`(−1)` Laurent
+coefficient of `f` at `z₀`. For a simple pole `f z = c / (z − z₀) + g z` with `g` analytic at `z₀`,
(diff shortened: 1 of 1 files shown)


CBirkbeck:
Closing: the local rubric review correctly flagged (correctness) that this is only the simple-pole limit `lim (z−z₀)f`, which diverges to a junk value for poles of order > 1 — not the roadmap's general order-(−1) Laurent coefficient. Mathlib exposes no general residue (only `meromorphicTrailingCoeffAt`, the leading coefficient). I'll define it correctly once the principal-value foundation lands. Superseding PR to follow.

PR #791, opened 2026-07-08
feat: add per-roadmap lines-of-Lean chart to the Statistics page

kim-em:
This PR adds a per-roadmap "lines of Lean" chart to the Statistics page: a stacked area of the cumulative net lines each roadmap has accrued over time, so a visitor can see not just how large the library is but which roadmap all that Lean is serving.

The chart is built from the `roadmap/<Area>` PR labels (added in https://github.com/TauCetiProject/TauCeti/pull/786, "feat: auto-label PRs with the roadmap they advance") rather than from git, because roadmap attribution lives in the labels, not in the tree. `scripts/loc_roadmap_graph.py` sums each merged PR's net diff (additions minus deletions) into its roadmap's running total by merge date, and renders a stacked SVG with a colour-coded legend, styled to match the two existing charts. It is a churn measure, not a `wc -l` of the tree: a line later rewritten counts under both PRs, so the figure is work landed per roadmap. Infrastructure and refactor PRs (`roadmap/none`) advance no roadmap and are excluded.

Like the existing charts, it is regenerated at every deploy and daily, so it needs no committed state; the SVG under `web/static_files/` is only the fallback snapshot a local `lake exe web` renders. Because the generator reads labels through `gh`, the `pages` workflow gains `pull-requests: read` and a generation step that keeps the committed fallback in place if the query fails, so a transient `gh` error cannot sink a deploy.

🤖 Prepared with Claude Code


kim-em:
Preview of the chart this PR adds to the Statistics page (GitHub renders the committed SVG in the browser):

**[View the rendered chart →](https://github.com/TauCetiProject/TauCeti/blob/roadmap-loc-chart/web/static_files/loc-per-roadmap.svg)**

As of Jul 8: **14 roadmaps, 85,338 net lines**, Jun 4 – Jul 8. Stacked largest-first, legend gives each roadmap's colour and net-lines total:

- ReductiveGroups 16,117 · HeegaardFloer 8,833 · OneParameterSemigroups 8,287 · CombinatorialHeegaardFloer 7,554 · UniversalCovers 7,047
- JacobianChallenge 6,519 · Multiquadratic 6,444 · Exchangeability 5,962 · PDE 5,160 · GeometricTopology 3,909
- ContourIntegration 3,680 · EffectiveBounds 3,348 · OrthogonalL2Bases 1,300 · ConformalMapping 1,178

`roadmap/none` (infra/refactor) is excluded. The metric is net lines added or refactored (summed PR diffs), not a `wc -l` snapshot.

PR #1169, opened 2026-07-24
refactor(Contour): share exists_Icc_mem_avoiding instead of duplicating it

weiscadence:
## What

`exists_Icc_mem_avoiding` was stated and proved twice, character for character, as a `private theorem` in both `Analysis/Contour/PiecewiseC1On.lean` and `Analysis/Contour/Crossing/Finiteness.lean` (only the docstring wording differed). This drops `private` on the `PiecewiseC1On` copy and deletes the duplicate from `Finiteness`.

Both call sites are unchanged. No mathematics changes; the statement and proof are the ones already on `main`.

## Why it resolves cleanly

`Finiteness` already imports `PiecewiseC1On` publicly and transitively (via `PwC1ImmersionOn`), and both files sit in `namespace TauCeti.Contour` with compatible `variable` binders. Only the `private` modifier forced the copy, so sharing needs no move or restatement.

## On the exposure

`api-design` asks that an implementation helper stay `private` "when it has no use outside the proof or file it serves". This one has a use outside the file it serves: two independent files need it, which is what produced the copy. `reuse` asks that near-clones be "factored into a shared construction". Since Lean's `private` is module-scoped, dropping it is the only way to share the single proof.

I searched Mathlib for a lemma this duplicates (interval/neighbourhood avoiding a finite set, `Icc _ _ ⊆ uIcc _ _` with a `Disjoint` clause, `*_avoiding` names) and did not find one, so this is not a restatement of existing library material.

If you would rather not widen the surface here, the alternative is relocating the lemma to a shared low-level file; happy to do that instead.

## Verification

`lake build` green: 5250 jobs, exit 0, no errors, matching the pre-change baseline on the same commit.

Diff: `+1 / -18` over 2 files.

### DIFF
--- a/TauCeti/Analysis/Contour/PiecewiseC1On.lean
+++ b/TauCeti/Analysis/Contour/PiecewiseC1On.lean
@@ -148,15 +148,13 @@ theorem isPiecewiseC1On_comm : IsPiecewiseC1On γ a b ↔ IsPiecewiseC1On γ b a
 theorem IsPiecewiseC1On.symm (h : IsPiecewiseC1On γ a b) : IsPiecewiseC1On γ b a :=
   isPiecewiseC1On_comm.mp h
 
-/-- Around any non-breakpoint interior parameter there is a closed subinterval of `[[a, b]]` with
-`t` in its interior and interior disjoint from the breakpoints, so the piecewise-`C¹` clause
-applies to it. -/
-theorem exists_Icc_mem_avoiding {p : Finset ℝ} {t : ℝ}
-    (ht : t ∈ Ioo (min a b) (max a b)) (htp : t ∉ (↑p : Set ℝ)) :
-    ∃ c d : ℝ, t ∈ Ioo c d ∧ Icc c d ⊆ uIcc a b ∧ Disjoint (↑p : Set ℝ) (Ioo c d) := by
-  have hopen : IsOpen (Ioo (min a b) (max a b) \ ↑p) :=
-    isOpen_Ioo.sdiff p.finite_toSet.isClosed
-  obtain ⟨ε, hε, hball⟩ := Metric.isOpen_iff.mp hopen t ⟨ht, htp⟩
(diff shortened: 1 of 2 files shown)


TauCeti/Analysis/Contour/PiecewiseC1On.lean
kim-em, line 1: ### ✅ generality — now passing on `e334575`.

kim-em, line 1: ### ✅ naming — now passing on `e334575`.

kim-em, line 1: ### 🟡 reuse — request_changes  `codex/gpt-5.6-sol`

The shared theorem still reimplements Mathlib’s closed-interval neighborhood construction.

- `TauCeti/Analysis/Contour/PiecewiseC1On.lean:157` — The `nhds_basis_Icc_pos` extraction and explicit symmetric endpoints duplicate `exists_Icc_mem_subset_of_mem_nhds`. _Fix:_ Apply `exists_Icc_mem_subset_of_mem_nhds hmem`, derive `t ∈ Ioo c d` via `Icc_mem_nhds_iff.mp`, and use its `Icc c d ⊆ Ioo (min a b) (max a b) \ s` inclusion for both remaining clauses.

<sub>`codex/gpt-5.6-sol` · 33s · 131.5k in / 1.2k out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/2c9d9f799fcdf9494e0131cdb9f754b48686ba6c...66fb3f20d7e5e75cb4db187d0c0718d7bfb49e9a) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/b9539a39556d34b3e3bf6199d0da7b4e7d5d4c27/rubrics/reuse.md)</sub>

kim-em, line 1: ### ✅ proof-quality — now passing on `66fb3f2`.

PR #372, opened 2026-06-23
refactor: dedupe the prime-radicand Galois corollaries via section variables

kim-em:
This PR removes the repeated instantiation boilerplate in `PrimeGaloisGroup.lean`. Each of the four prime-radicand corollaries re-spelled the same four-argument instantiation of the field-generic multiquadratic Galois lemmas — `(d := fun i => (p i : ℚ)) (root := fun i => Real.sqrt (p i)) (fun i => sq_sqrt_natCast (p i)) (not_isSquare_prod_primes_of_injective p hp hinj)`. It moves `p`, `hp`, `hinj` into section variables, introduces a private `sqrtPrime_sq` lemma whose type already pins the radicand data `d` and the generators `root` so the explicit `(d := …) (root := …)` arguments are inferred and dropped, and switches the two `@[simp]` apply-lemmas from `by exact` to term mode. The public statement forms, lemma names, and exported argument lists are unchanged, and the full library builds with `warningAsError` on.

This is maintenance of existing code (always in scope per `AGENTS.md`); the file's corollaries serve the prime-radicand multiquadratic Galois-group target under `TauCetiRoadmap/Multiquadratic`, Layer 0 (multiquadratic field degree / Galois group).

Roadmap: Multiquadratic

🤖 Prepared with Claude Code

PR #1176, opened 2026-07-25
refactor(Multiquadratic): share sqrtPrime_sq instead of duplicating it

weiscadence:
`sqrtPrime_sq` was a byte-identical `private theorem` in two files:

- `TauCeti/NumberTheory/Multiquadratic/Prime/GaloisGroup.lean`
- `TauCeti/NumberTheory/Multiquadratic/Prime/Subfield/Lattice.lean`

Same namespace (`TauCeti.Multiquadratic`), same section binders, same one-line proof term `sq_sqrt_natCast (p i)`. This PR keeps one copy, in a file both already `public import`, and deletes the other two. No call site changes.

## Why share rather than inline

The helper is deliberate, not accidental redundancy, and it's worth being explicit about that because it determined the shape of this PR.

#372 (*"dedupe the prime-radicand Galois corollaries via section variables"*) introduced it on purpose:

> introduce a private `sqrtPrime_sq` whose type already pins the radicand data `d` and the generators `root` (so the explicit `(d := …) (root := …)` are inferred and dropped)

So the obvious-looking alternative — delete both copies and pass `(fun i => sq_sqrt_natCast (p i))` directly, which is already the spelling used in `Prime/Subfield/Count.lean`, `Prime/Subfield/Degree.lean` (4×) and `Prime/Radicands.lean` itself — is a regression against a considered decision, even though it does compile. I implemented that version first and discarded it after reading #372.

What #372 did *not* do is duplicate the helper: it touched a single file. The second copy appeared later, so the **duplication** is accidental even though the **helper** is not. Hence: share it, don't remove it.

## Placement

`Prime/Radicands.lean` is the natural home:

- both files already `public import` it, so nothing new enters the import graph;
- it already contains `sq_sqrt_natCast`, the lemma that *is* this proof term, so the two now sit together;
- its module docstring already describes supplying the `hroot` hypothesis, which is precisely this helper's role.

The binders are generalised from the prime-family section variables to `{ι : Type*} (p : ι → ℕ)`. Primality is never used in the statement or the proof, and the two consumers pass a prime family regardless, so nothing at the call sites changes. #372's rationale is carried over into the docstring so the type-pinning purpose is not lost now that the declaration lives away from its consumers.

## Verification

- Full `lake build` green **before** the change (baseline) and **after**: 5250 jobs, exit 0 both times.
- Canary, to establish the green build is actually sensitive to this change rather than a blind pass: renaming the shared declaration makes all four `GaloisGroup.lean` call sites fail with ``Unknown identifier `sqrtPrime_sq` ``. Restored, rebuilt green. So the call sites demonstrably resolve to the new shared declaration.
- All touched lines ≤ 100 characters.

Net `+8 / -12`.

## One thing I deliberately left out of scope

The name says `Prime`, but primality is not used — the statement holds for any `p : ι → ℕ`, which is how it is now stated. A name like `sq_sqrt_natCast_apply` (or similar, following the `sq_sqrt_natCast` sibling) would describe it more honestly.

I left the name as #372 chose it, since renaming a now-public declaration is a separate topic from removing a duplicate, and bundling them would make this diff harder to judge. Happy to fold a rename into this PR if you'd prefer it in one go.

Roadmap: Multiquadratic

### DIFF
--- a/TauCeti/NumberTheory/Multiquadratic/Prime/GaloisGroup.lean
+++ b/TauCeti/NumberTheory/Multiquadratic/Prime/GaloisGroup.lean
@@ -43,13 +43,6 @@ section PrimeFamily
 
 variable [Finite ι] (p : ι → ℕ) (hp : ∀ i, (p i).Prime) (hinj : Function.Injective p)
 
-omit [Finite ι] in
-/-- The square roots `√(p i)` of a prime family satisfy the radicand equation `√(p i)² = p i`,
-the `hroot` hypothesis the field-generic multiquadratic results consume. This also pins the
-radicand data `d i = p i` and the generators `root i = √(p i)` for those results. -/
-private theorem sqrtPrime_sq (i : ι) : Real.sqrt (p i) ^ 2 = algebraMap ℚ ℝ (p i : ℚ) :=
-  sq_sqrt_natCast (p i)
-
 include hp hinj
 
 /-- **The Galois group of a prime-radicand multiquadratic field is `(ℤ/2)ⁿ`.** For a finite family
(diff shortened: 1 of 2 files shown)


TauCeti/NumberTheory/Multiquadratic/Prime/Radicands.lean
kim-em, line 1: ### ✅ reuse — now passing on `72e8187`.

PR #1180, opened 2026-07-25
fix(AdditiveFrobeniusKernel): drop dead @[simp] on mk_ι_pow_eq_zero

weiscadence:
## What

Removes the `@[simp]` attribute on `mk_ι_pow_eq_zero` and deletes its
grandfathered entry from `scripts/lint-baseline.txt`. Purely subtractive
(−2 lines, no new declarations, statement unchanged).

## Why it's dead

The lemma's LHS mentions `(hopfIdeal p).toIdeal`. The sibling simp lemma
`hopfIdeal_toIdeal` — declared immediately above and also `@[simp]` —
rewrites that to `Ideal.span {ι R R 1 ^ p}`. So on any normal-form goal
simp rewrites the ideal first and this lemma's LHS can never match. That
is exactly what `simpNF` reports ("Left-hand side simplifies … using
`hopfIdeal_toIdeal`"), so the attribute never fires.

## Why remove rather than restate

The `simpNF` hint suggests restating the LHS to the simplified term. I
tried that and rejected it: the lemma's sole consumer passes it as an
`IsNilpotent` witness for `x := (Ideal.Quotient.mk _ (ι R R 1))`, where
the ideal is inferred as `(hopfIdeal p).toIdeal`. A restated LHS over
`Ideal.span {…}` no longer unifies at that call site. Since the only use
is an explicit term (not a simp call), dropping the attribute is the
minimal correct fix and leaves the statement intact.

## Verification

- Full `lake build` green (5251 jobs) — nothing downstream relied on it
  firing via simp.
- `scripts/lint-env.sh` passes (exit 0); the removed baseline line no
  longer corresponds to a violation.

### DIFF
--- a/TauCeti/Algebra/AlgebraicGroup/AdditiveFrobeniusKernel/Basic.lean
+++ b/TauCeti/Algebra/AlgebraicGroup/AdditiveFrobeniusKernel/Basic.lean
@@ -185,7 +185,6 @@ noncomputable abbrev CoordinateRing : Type u :=
   SymmetricAlgebra R R ⧸ (hopfIdeal (R := R) p).toIdeal
 
 /-- The class of `x` in the coordinate ring of `αₚ` is `p`-nilpotent: its `p`-th power is `0`. -/
-@[simp]
 theorem mk_ι_pow_eq_zero :
     (Ideal.Quotient.mk (hopfIdeal (R := R) p).toIdeal
         (ι R R 1 : SymmetricAlgebra R R)) ^ p = 0 := by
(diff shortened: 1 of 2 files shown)


TauCeti/Algebra/AlgebraicGroup/AdditiveFrobeniusKernel/Basic.lean
kim-em, line 1: ### ✅ api-design — now passing on `501167a`.

kim-em, line 1: ### ✅ reuse — now passing on `501167a`.

PR #1201, opened 2026-07-26
feat(L2Product): product Hilbert basis of L² on a product measure (roadmap B3)

weiscadence:
Completes **Part B3 of the `OrthogonalL2Bases` roadmap**: the product Hilbert basis of `L²(μ ⊗ ν)`.

`TauCeti/Analysis/InnerProductSpace/L2Product.lean` already proved the orthonormality half, and its
module docstring stated that the completeness half "is left to a separate construction". This PR
supplies that construction, in the same file, and delivers the roadmap targets:

- `prodHilbertBasis` — matches `TauCetiRoadmap/OrthogonalL2Bases/Suggested.lean` verbatim.
- `coeFn_prodHilbertBasis` — the roadmap's `prodHilbertBasis_apply`, renamed to the `coeFn_`
  convention this file and Mathlib already use for a.e.-representative lemmas. Statement identical.

This unblocks the roadmap's `piHilbertBasis` and `gaussianHermitePiBasis`.

## The completeness argument

The obvious route is a Fubini slice: fix `i`, show `y ↦ ∫ conj (b₁ i x) · h (x, y) ∂μ` vanishes a.e.
That needs a null set per index, so it silently requires `ι₁` countable — which the statement does
not assume. Instead the proof splits so that no null-set bookkeeping is indexed by `ι₁` or `ι₂` at
all:

1. `inner_L2prodMul_eq_zero_of_forall_basis` — orthogonality to the *basis* tensors upgrades to
   orthogonality to *every* elementary tensor `F ⊗ G`. Pure Hilbert-space geometry: expand along
   the basis with `HilbertBasis.hasSum_repr` and push the sum through the continuous linear map
   `L2prodMulRightL`, then again in the other slot. No countability, no measure theory.
2. `setIntegral_prod_eq_zero_of_forall_inner` — with `F`, `G` now arbitrary, take indicators:
   `1ₐ ⊗ 1_b` is the indicator of `a ×ˢ b`, so every rectangle integral vanishes.
3. `setIntegral_eq_zero_of_forall_prod` — the Dynkin (π-λ) step over `isPiSystem_prod` /
   `generateFrom_prod`, run inside a finite box, then a monotone exhaustion along
   `spanningSets μ n ×ˢ spanningSets ν n` to reach every finite-measure set. `SigmaFinite` is used
   exactly here, and is what the roadmap signature already assumes.

Conclusion via `Lp.ae_eq_zero_of_forall_setIntegral_eq_zero`.

## Reuse

[...]

PR #1202, opened 2026-07-26
feat(L2Pi): Fintype-indexed product Hilbert basis of L² on a pi measure (roadmap B3/D)

weiscadence:
Completes **Part B3/D of the `OrthogonalL2Bases` roadmap**: the `Fintype`-indexed product Hilbert
basis of `L²(Measure.pi μ)`.

Roadmap targets delivered (`TauCetiRoadmap/OrthogonalL2Bases/Suggested.lean`):

- `piHilbertBasis` — signature matches the roadmap verbatim.
- `coeFn_piHilbertBasis` — the anti-vacuity pin: the `k`-th basis vector really is a.e.
  `∏ i, b i (k i)`, not merely something of the right type. (Named for the `coeFn_` convention
  Mathlib uses for a.e.-representative lemmas.)

This is the last input `gaussianHermitePiBasis` needs on the pi side.

## Why this is not an iterate of the binary product basis

Mathlib already supplies both Fubini inputs at this arity — `Integrable.fintype_prod_dep` and
`integral_fintype_prod_eq_prod` (the latter needs no integrability side conditions) — so the
orthonormality half is a short direct construction rather than an induction over `Fintype.card ι`
with a `MeasurePreserving` transport of Hilbert bases through `piEquivPiSubtypeProd`. The direct
route also avoids introducing a `HilbertBasis`-transport API that nothing else would use.

## The completeness argument

The slice route would need one null set per basis index, i.e. countability of each `κ i`, which the
statement does not assume. This proof instead never indexes a null set by `κ i` at all:

1. `inner_L2piMul_eq_zero_of_forall_basis` — orthogonality to the *basis* tensors upgrades to
   orthogonality to *every* elementary tensor. Coordinates are generalized one at a time by `Finset`
   induction: each step expands a single slot along its basis via `HilbertBasis.hasSum_repr` and
   pushes the sum through the continuous linear map `L2piMulSlot`. Pure Hilbert-space geometry.
2. `setIntegral_pi_eq_zero_of_forall_inner` — testing against indicators; a tensor of indicators is
   the indicator of the box.
3. `setIntegral_eq_zero_of_isPiSystem` — the Dynkin (π-λ) step, then a monotone exhaustion along
   `∏ i, spanningSets (μ i) n`.

Conclusion via `Lp.ae_eq_zero_of_forall_setIntegral_eq_zero`.

## Note for the `reuse` rubric — deliberate generality, and a follow-up I am offering

`setIntegral_eq_zero_of_isPiSystem` is stated for an **arbitrary** π-system generating the
σ-algebra, not for boxes. It is the Bochner analogue of Mathlib's existing
`lintegral_eq_lintegral_of_isPiSystem` (which is `ℝ≥0∞`-valued and so cannot be used here).

Stating it generally is deliberate: the sibling PR #1201 (binary product basis, currently in review)
contains `setIntegral_eq_zero_of_forall_prod`, which is the *rectangle* case of this same lemma. I
did not fold the refactor into either PR because they touch a common file and would conflict while
both are open. **Once #1201 merges I will follow up with a PR deleting
`setIntegral_eq_zero_of_forall_prod` and deriving the rectangle case from
`setIntegral_eq_zero_of_isPiSystem`.** Flagging it here rather than leaving it to be discovered.

## Verification

- `lake build TauCeti.Analysis.InnerProductSpace.L2Pi` — succeeds, no `sorry`.
- `#print axioms` on `piHilbertBasis`, `coeFn_piHilbertBasis`,
  `orthogonal_span_range_L2piMul_eq_bot`, `inner_L2piMul_eq_zero_of_forall_basis`:
  `[propext, Classical.choice, Quot.sound]` only.
- One new file, entirely under `TauCeti/`; no baseline, script, or root-module changes.


TauCeti/Analysis/InnerProductSpace/L2/Pi.lean
kim-em, line 1: ### ✅ reuse — now passing on `18d763b`.

kim-em, line 1: ### ✅ api-design — now passing on `18d763b`.

kim-em, line 1: ### ✅ generality — now passing on `18d763b`.

kim-em, line 1: ### ✅ placement — now passing on `18d763b`.

kim-em, line 1: ### ✅ documentation — now passing on `18d763b`.

kim-em, line 1: ### ✅ proof-quality — now passing on `18d763b`.

kim-em: **Re: your reply on `placement` —** re-reviewed on `e02bdf0`; the finding stands — The new L2 directory still leaves the existing prefix-sharing L2 module flat; sequencing concerns do not satisfy the required same-PR migration.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3651520830.</sub>

PR #832, opened 2026-07-09
refactor: correct PiecewiseC1On docstring on ArcFTC

kim-em:
This PR corrects the module docstring of `Analysis/Contour/PiecewiseC1On.lean`, which claimed that the raw contour-integral lemmas — "for example the fundamental theorem of calculus along a contour in `Contour.ArcFTC`" — consume `IsPiecewiseC1On` as the hypothesis a regularity package supplies. That is false: `ArcFTC` is deliberately stated on the integrand (continuity, pointwise `HasDerivAt`, and integrability hypotheses), takes no such predicate, and its own docstring notes it "can be used with any regularity package that supplies it"; nothing in the repository currently consumes `IsPiecewiseC1On`. The revised text says instead that the predicate is the regularity a package will use to discharge those integrand-level hypotheses, and that the raw lemmas do not consume the predicate themselves. This discharges the refactor lead in task 3 of https://github.com/TauCetiProject/TauCeti/issues/804 (Analysis/Contour: make the on-curve winding-number layer real).

Per `AGENTS.md`, improving code that already exists is always in scope; this PR adds no new mathematical scope. `lake build` completes all 4618 jobs, `lake exe axioms` reports all 8741 declarations within the allowlist [propext, Classical.choice, Quot.sound], and `lake exe module-system` confirms all 441 modules opt into the module system.

🤖 Prepared with Claude Code

Issue #804, opened 2026-07-08
Analysis/Contour: make the on-curve winding-number layer real (HW Prop 2.2 and regularity-condition witnesses)


kim-em:
From a whole-repo quality audit (2026-07-08, at d76e9fa). Verify each claim against current `main` before acting, and comment here if a claim no longer holds.

The ContourIntegration roadmap's distinctive layer, non-integer winding numbers for singularities **on** the contour (Hungerbühler-Wasem), exists only as definitions that nothing exercises:

- `Analysis/Contour/RegularityConditions.lean`'s `ConditionAprime`, `ConditionB`, `SectorCompatible`, `crossingAngle`, `basepointAngle`, `FlatOfOrder` each appear in exactly one file (their own): zero consumers, zero instantiation lemmas, so their faithfulness to the paper is currently unfalsifiable. The only computed value is `crossingAngle = π` at a smooth crossing.
- `ModelSectorWinding.lean`'s ½ and 1/6 "winding number" theorems evaluate `(2πi)⁻¹ ∫ arc` at the arc's *centre*, an off-curve point where the principal value is the ordinary integral; nothing yet computes a winding number to a non-integer value at an on-curve point. (These match the pinned `Suggested.lean` signatures, so this is a sequencing observation, not a defect.)
- `PiecewiseC1On.lean`'s `IsPiecewiseC1On` (136 lines with `mono`/`symm`/`of_contDiffOn` API) is orphaned, and its docstring's claim that `Contour.ArcFTC` consumes it is false: `ArcFTC.lean` takes raw `HasDerivAt`/`ContinuousOn`/`IntervalIntegrable` hypotheses and never references it.

Tasks (new mathematics except task 3; cite `TauCetiRoadmap/ContourIntegration/README.md` and carry an authoring claim + `tauceti-target` marker per `TauCeti/Coordination.lean`):

1. Prove HW Proposition 2.2 (the finite-crossing decomposition) and the lemma equating an on-curve corner's `windingNumber` with `crossingAngle / 2π`, making `crossingAngle`, the model-sector computations, and `WindingNumberConcat` load-bearing.
2. Add at least one witness lemma per regularity condition (a circle around a simple pole satisfies `ConditionAprime`; a concrete order-2 pole with compatible angle satisfies `ConditionB`), plus the roadmap's `hasCauchyPV_half_residue` acceptance test when reachable.
[...]

PR #1205, opened 2026-07-26
refactor(Contour): export the simple-pole witnesses for regularity conditions (A') and (B)

weiscadence:
Addresses part of #804 (task 2: witness lemmas for the regularity conditions).

## Roadmap

This advances `TauCetiRoadmap/ContourIntegration/README.md` — the Hungerbühler–Wasem layer whose regularity conditions these two lemmas instantiate, and the subject of #804.

## The finding

#804 asks for instantiation lemmas for `ConditionAprime` and `ConditionB`, on the grounds that they have *"zero consumers, zero instantiation lemmas, so their faithfulness to the paper is currently unfalsifiable."*

Two such lemmas already exist on current `main`. They are `private`:

| lemma | file |
|---|---|
| `conditionAprime_of_simple_poles` | `TauCeti/Analysis/Contour/HungerbuhlerWasem.lean` |
| `conditionB_of_simple_poles` | `TauCeti/Analysis/Contour/HungerbuhlerWasem.lean` |

Both are used at two call sites each, inside that one file, to make `hungerbuhlerWasem_residueTheorem_of_simple_poles` unconditional.

Crucially they are **not vacuous witnesses**. `conditionAprime_of_simple_poles` discharges the interior flatness clause with content: at an interior crossing it forces the pole order to `1` (via `neg_one_le_meromorphicOrderAt` plus the simple-pole hypothesis) and then supplies first-order flatness from `IsPwC1ImmersionOn.flatOfOrder_one`. That is a crossing the curve actually makes, with `FlatOfOrder` genuinely exercised.

So the gap #804 describes is not that witnesses are missing. It is that the ones that exist cannot be reached from outside the file that consumes them.

## The change

Visibility only. `private` is dropped on both lemmas, and their docstrings are expanded to state which hypothesis set each discharges and why the flatness clause is non-vacuous. No statement, no proof, and no call site is modified.

## On the api-design question this raises

The obvious objection is that a declaration with no consumer outside its own file is exactly what `private` is for. I think the roadmap issue is the justification that overrides it here: #804 asks specifically for these conditions to be instantiable so their faithfulness to Hungerbühler–Wasem becomes falsifiable, and `private` is the single thing preventing that. The export is the fix, not a side effect of one.

## What I deliberately did not include

I first wrote three new lemmas for this task — an off-curve `ConditionAprime` witness (curve avoids the singular set), its circle instance, and a `ConditionB` witness for integrands with no higher-order on-curve poles. I dropped all three before opening this PR. They discharge the flatness and sector clauses **vacuously** (no crossing exists to constrain, so `FlatOfOrder` and `SectorCompatible` are never exercised), which does not address what #804 is actually asking for, and they are strictly weaker than the two lemmas above. Adding them alongside would also have made this diff harder to judge. I can open them separately if a consumer ever wants the off-curve case.

## Verification

- `lake build TauCeti.Analysis.Contour.HungerbuhlerWasem` — 2792 jobs, `rc=0`
- `lake build` (full library) — 5352 jobs, `rc=0`
- `scripts/lint-env.sh` — `LINT-ENV: PASS`, no new violations (55 grandfathered, 2 ratchetable)

## Note on #804's premise

While scoping this I found three claims in #804 that no longer hold on current `main`: the task-3 docstring correction was already discharged by #832, `IsPiecewiseC1On` now has three consumers rather than zero, and the "zero instantiation lemmas" claim above. I have commented the first two on the issue; I will add the third once this lands so the correction arrives with the diff attached.

Roadmap: ContourIntegration

### DIFF
--- a/TauCeti/Analysis/Contour/HungerbuhlerWasem.lean
+++ b/TauCeti/Analysis/Contour/HungerbuhlerWasem.lean
@@ -183,8 +183,13 @@ private theorem neg_one_le_meromorphicOrderAt {f : ℂ → ℂ} {U : Set ℂ} {S
       h_an.meromorphicOrderAt_nonneg
 
 /-- **Condition (A′) is automatic at simple poles.** The only pole order the interior clause can
-meet is `1`, discharged by the first-order flatness of the immersion, and the basepoint of
-the closed curve is off the singularities.
+meet is `1`, discharged by the first-order flatness of the immersion, and the basepoint
+`γ (min a b)` is off the singularities.
+
+The curve need not be closed: `ConditionAprime` constrains the basepoint `γ (min a b)`, so
+`hymin` is the whole premise the proof consumes. A closed contour supplies it from
+`γ a = γ b` together with `γ a ∉ S`, which is what
+`hungerbuhlerWasem_residueTheorem_of_simple_poles` does at its call site.
 
 This is an instantiation lemma for `TauCeti.Contour.ConditionAprime`: it exhibits a hypothesis
 set a caller can actually supply — a piecewise-`C¹` immersion, closed, rooted off `S`, with `f`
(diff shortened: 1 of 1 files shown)


TauCeti/Analysis/Contour/HungerbuhlerWasem.lean
eohjelle, line 1: ### 🟡 generality — request_changes  `codex/gpt-5.6-sol`

The exported Condition (A′) witness unnecessarily restricts callers to closed curves, although the proof only needs the distinguished endpoint outside the singular set.

- `TauCeti/Analysis/Contour/HungerbuhlerWasem.lean:194` — `conditionAprime_of_simple_poles` assumes both `hclosed : γ a = γ b` and `hγa`, solely to establish `γ (min a b) ∉ S`; ConditionAprime itself does not require a closed curve. _Fix:_ State the general lemma with `hγmin : γ (min a b) ∉ (S : Set ℂ)` instead, and derive that premise from `hclosed` and `hγa` at the closed-contour call site.

<sub>`codex/gpt-5.6-sol` · 63s · 153.2k in / 878 out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/3096557b3362bc9d7a6ecbc0d253072ec750683a...982bb931786ea5ce1af42cc9c6c61c4b28b97ac8) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/b9539a39556d34b3e3bf6199d0da7b4e7d5d4c27/rubrics/generality.md)</sub>

kim-em, line 1: ### ✅ api-design — now passing on `707f7b9`.

kim-em, line 1: ### ✅ documentation — now passing on `707f7b9`.

PR #1313, opened 2026-07-28
feat(Contour): expose the piece/window decomposition of a principal value

CBirkbeck:
Re-scoped after the dry-run blocked the previous shape on five rubrics. The findings agreed on
one thing: the headline `windingNumber_eq_windowPieceSum` was a composition of two existing
lemmas with no consumer, and its docstring claimed an integrality property the statement did not
have. That theorem and its file are **removed**; what is left is the API that was genuinely
missing.

## What this PR now does

* Exposes `hasCauchyPVAt_of_perWindow` and `windowPieceSum` in `Crossing/PVAggregation.lean` —
  the general piece/window decomposition of a principal value. The induction already existed as
  the private `hasCauchyPVAt_along_sorted`, but its only public consumer,
  `hasCauchyPVAt_of_perWindow_boundary_tendsto`, assumes a curve-antiderivative `Φ` and telescopes
  to `Φ (γ b) - Φ (γ a)`. The Cauchy kernel `(z - z₀)⁻¹` has no such global antiderivative — `log`
  exists only off a branch cut — so the untelescoped form is the one a winding-number consumer
  needs, and nothing cancels.

* Moves `hasCauchyPVAt_of_dist_lower_bound` to `Cauchy/PrincipalValue/Basic.lean` and weakens it.
  It is a general fact about `HasCauchyPVAt` — away from the pole the principal value is the
  ordinary integral — with nothing to do with crossings or window aggregation, so it belongs with
  the predicate rather than in a specialised module. Its integrability hypothesis was `∀ ε > 0`;
  `HasCauchyPVAt` itself only asks for `∀ᶠ ε in 𝓝[>] 0`, so it now asks for that too.

* Weakens the window-placement hypotheses throughout from strict to non-strict: `a ≤ t - r`
  rather than `a < t - r`, and `2 * r ≤ |t - t'|` rather than `2 * r < |t - t'|`. Concatenation
  never needed the strictness — a first window touching `a`, or consecutive windows touching at
  an endpoint, simply produce a degenerate arc, on which the principal value holds trivially. The
  private induction is weakened with them, so the two pre-existing public aggregation theorems
  gain the generality too.

## What was dropped and why

`windingNumber_eq_windowPieceSum`. Two rubrics were right that it added nothing over composing
`hasCauchyPVAt_of_perWindow` with `windingNumber_eq_of_hasCauchyPVAt` at a call site, and
correctness was right that its docstring's claim — that the crossings carry all the
non-integrality — is false: the arcs there are *open* arcs, whose integrals are not multiples of
`2πi`. The integrality in Hungerbühler–Wasem is a statement about the *closed* curve `Λ̃` formed by
joining the arcs with detours, which nothing here constructs.

It is better re-introduced once the geometric half of Proposition 2.2 exists and gives it a real
consumer; #1314 and #1318 are the first two steps of that.

## Scope

Roadmap-gated: Layer 1 of `TauCetiRoadmap/ContourIntegration/README.md`. No new mathematics — two
declarations made public, one relocated with a weaker hypothesis, one removed.

## Gates

`lake build` · `lake exe axioms` · `lake exe module-system` · `scripts/lint-env.sh` — all green.


CBirkbeck:
🔍 Review in progress — `codex` reviewing `bce1fb87bc54`.

CBirkbeck:
Closing as superseded — every part of this PR has either landed elsewhere or been overtaken.

**Already on main.** The `PrincipalValue/Basic.lean` half of this diff is `hasCauchyPVAt_of_dist_lower_bound`. That landed via #1357 as `HasCauchyPVAt.of_dist_lower_bound` (with an existence form `cauchyPVExistsAt_of_dist_lower_bound` alongside it), and main now has four call sites. The `hr_nonneg` relaxation and the `_of_interiorDisjoint` forms this PR also carried landed in the same PR.

**Overtaken.** The remaining content is publicising `windowPieceSum` and the piece/window decomposition. #1365 has since rewritten `PVAggregation.lean` around Mathlib's `Finset.intervalGapsWithin` and `Finset.sum_intervalGapsWithin_add_sum_eq_sub`, in response to the `reuse` finding that blocked *this* PR for seven rounds. Exposing `windowPieceSum` now works against that direction: the `reuse` rubric's position was that the piece/window aggregation should be expressed through the Mathlib gap API rather than through a bespoke recursive definition.

**Not rebaseable in any meaningful sense.** The base is 29 commits behind, and an attempted rebase conflicts in five hunks, every one of which resolves to "main already does this, better". What would survive is a publicisation with no consumer.

If a consumer for the decomposition appears later, the right move is a fresh, minimal PR stated over the gap API that main now uses — not a revival of this branch.

PR #1161, opened 2026-07-24
refactor(Analysis/PDE): remove unconsumed ellipticity patching API

fraware:
## Summary
- Remove the unconsumed domain-patching API from `TauCeti/Analysis/PDE/Uniform/Ellipticity.lean`: `UniformlyEllipticOn.union` / `iUnion` / `biUnion` and `uniformlyEllipticOn_union_iff` / `uniformlyEllipticOn_iUnion_iff` / `uniformlyEllipticOn_biUnion_iff`.
- Drop the corresponding module-doc bullets and unused variables (`ι`, `Ωs`, and post-namespace `Ω'`).
- Keep `mono_set` / `mono_constants` / `congr`; no deprecation aliases; no other mathematical or API changes.

Partially addresses #793 (independent single-file follow-up from the Task-2 decomposition; does not close the issue).

Scope: `AGENTS.md` — improving existing code (refactor / dead-code removal) is always in scope.

## Test plan
- [x] Confirmed the six declarations existed only in the target file (no other TauCeti consumers)
- [x] `lake env lean TauCeti/Analysis/PDE/Uniform/Ellipticity.lean` (pre-edit, post-edit, and post-rebase): exit 0
- [x] `lake build`: `Build completed successfully (5250 jobs).`
- [x] `lake exe axioms`: `axioms: audited 11252 TauCeti declaration(s); all within the allowlist [propext, Classical.choice, Quot.sound].`
- [x] `lake exe module-system`: `module-system: all 630 TauCeti module(s) opt into the module system.`
- [x] `git diff --check` clean on the changed file; zero remaining name matches for the six removed declarations
- [ ] CI green on this PR

Local `scripts/lint-env.sh` reported baseline ratchet noise unrelated to this file (no Ellipticity hits); treating CI lint as authoritative after rebase onto current `main`.

## AI disclosure
Prepared with Cursor (Composer / Auto agent).

Roadmap: PDE

PR #1047, opened 2026-07-20
chore: drop the stale ShiftInvariantSigma simpNF baseline entry

cameronfreer:
## Summary

Follow-up to #1043: delete the now-stale grandfathered entry

```
simpNF TauCeti.Probability.isShiftInvariant.preimage_shift_iterate_eq
```

from `scripts/lint-baseline.txt`. The declaration was removed in #1043 (the
`ShiftInvariantSigma` adapter module deleted per the merged roadmap change
[TauCetiRoadmap#73](https://github.com/TauCetiProject/TauCetiRoadmap/pull/73)), so since
that merge `scripts/lint-env.sh` prints this entry as a non-failing RATCHET reminder. This
PR performs exactly the cleanup the script's header prescribes ("delete it from the
baseline in a follow-up PR").

One-line deletion; no `.lean` changes. `scripts/` is human-owned, so this is intentionally
separate from #1043 rather than bundled into it.

*Directed by Cameron Freer, written by Fable 5 high, reviewed by GPT 5.6 Sol xhigh.*

PR #1450, opened 2026-07-30
refactor(ConditionallyIID): name the symmetry consequences of the conditional predicate

cameronfreer:
## What this is

Four named symmetry consequences of the conditional predicate, replacing a composition that
`DeFinetti/Theorem.lean` inlined at three sites.

```lean
ConditionallyIIDWith.exchangeable   ConditionallyIIDWith.contractable   -- at a named directing measure
ConditionallyIID.exchangeable       ConditionallyIID.contractable       -- existential corollaries
```

They live in a new `Exchangeability/ConditionallyIID/Implications.lean`, mirroring
`MixedIID/Implications.lean`: the predicate and its projection stay in `Basic`, the implications out
of the predicate sit one layer above, so `Basic` keeps its original import set. `Exchangeable` and
`Contractable` already arrive through `MixedIID.Basic`, so stating them needs no new *public*
import; `MixedIID.Implications` enters proof-only for the two mixture-side lemmas.

The layering matches the mixture side exactly. `ConditionallyIIDWith.exchangeable` composes
`mixedIIDWith_of_conditionallyIIDWith` with `MixedIIDWith.exchangeable`, preserving the witness;
`ConditionallyIID.exchangeable` destructs `exists_directing` and applies it — the same shape as
`MixedIIDWith.exchangeable` / `MixedIID.exchangeable`. A caller holding a named directing measure
therefore never has to weaken to the existential to obtain exchangeability.

`Theorem.lean`'s three inlined `(mixedIID_of_conditionallyIID _).exchangeable` / `.contractable`
call sites now cite the named lemmas.

## History, for anyone reading the thread

This branch began as a duplicate: #1444 landed the conditional equivalence API while it was in
progress, so that work was dropped and intention #104 closed as superseded. What survives is a
refactor of already-merged code, which `AGENTS.md` puts in scope without a roadmap intention — hence
`roadmap/none`.

Two things were tried and withdrawn, and should not be read as still present in the diff:

* **A `deFinetti` argument-order swap.** Dropped in `2750602`. I contested the finding on the
  grounds that the theorem had no call sites and was thirty minutes old; both were true, but the
  roadmap pins the order explicitly as
  `deFinetti (X) (hX_meas : ∀ i, Measurable (X i)) (hX_exch : Exchangeable μ X)`, so the rubric had
  specification precedent and I was wrong on the merits. `deFinetti` is byte-for-byte main's.
* **A module-docstring "refresh".** The stale text was this branch's, not main's, and the edit
[...]

PR #1453, opened 2026-07-30
chore: ratchet the three stale simpNF baseline entries

cameronfreer:
## Why

`scripts/lint-env.sh` on current `main` reports:

```
lint-env: RATCHET — 3 baseline entr(y/ies) no longer violate.
Please delete these lines from scripts/lint-baseline.txt (a follow-up PR is fine):
  simpNF TauCeti.AlphaP.mk_ι_pow_eq_zero
  simpNF TauCeti.PDE.uniformlyEllipticOn_biUnion_iff
  simpNF TauCeti.Probability.permReindex_apply
LINT-ENV: PASS — no new violations (54 grandfathered, 3 ratchetable).
```

This is the follow-up it asks for, and the same shape as #1047.

## What

Delete those three lines. Nothing else changes.

`TauCeti.PDE.uniformlyEllipticOn_biUnion_iff` no longer exists in the library at all — #1161
removed the unconsumed ellipticity patching API it belonged to. `AlphaP.mk_ι_pow_eq_zero` and
`Probability.permReindex_apply` still exist and are now in simp normal form.

Leaving them grandfathered is not merely untidy: a baseline entry suppresses the linter for that
declaration, so a future change that pushed one of these back out of simp normal form would pass CI
silently.

## Verification

Re-running the script on this branch:

```
lint-env: linted 860 modules; 54 (linter, declaration) violation(s).
LINT-ENV: PASS — no new violations (54 grandfathered, 0 ratchetable).
```

The baseline is now exact. Note the `structureInType TauCeti.BCRPoint✝` entry is **not** included —
that declaration still exists as a private structure and still violates, so it stays.

## Review path

This touches `scripts/`, which `AGENTS.md` designates human-owned, so it needs a human review and
merge rather than the auto-merge path. Flagging that deliberately: the three-line baseline change *is* the
deliverable, and it should not be reshaped to look auto-mergeable.


cameronfreer:
The red `build` and `scope` checks here are the designed routing, not a defect to fix.

The workflow run sets `OUT_OF_SCOPE=1`, skips the sandboxed build entirely (the job finished in 12s
without compiling anything), and posts `scope=failure` with *"changes outside TauCeti/ — human review
+ merge required"* plus a companion `build=failure`. That is the gate deliberately steering a
human-owned diff away from auto-merge.

Per `CLAUDE.md`, the correct response is to wait for a human review and merge — not to reshape the
PR until it looks auto-mergeable. The three deleted `scripts/lint-baseline.txt` lines *are* the
deliverable, so there is nothing here to strip. Verification therefore has to come from the local
run quoted in the description: `LINT-ENV: PASS — no new violations (54 grandfathered, 0
ratchetable)`, against `main` at `360baae8`.

Nothing in `TauCeti/` changes, so no Lean rebuild is implicated either way.

cameronfreer:
Both description corrections applied; no code change.

* **`CLAUDE.md` → `AGENTS.md`.** Worth recording *why*, since the file does exist: `.claude/CLAUDE.md`
  is tracked, but it is a **symlink** to `../AGENTS.md`. So my link resolved to a symlink blob, which
  GitHub renders as the literal path text rather than the document. `AGENTS.md` is the real file, and
  the link now points there. I have made the same substitution on my other two open PRs that cited it
  (#1450, #1479); #1482 and #1485 never referenced it.
* **"the single-line change" → "the three-line baseline change."** The diff removes three lines.

**Correcting my own earlier claim in this comment.** I originally wrote that an independent run had
reproduced `PASS — 54 grandfathered, 0 ratchetable` "from a clean tree", and that this excluded
contamination my local run could not. That was an overstatement on my part: the independent check
used an isolated source archive but symlinked its `.lake` to the existing workspace build. It
therefore verified the PR diff and the ratchet logic independently, but it did **not** rule out
stale-olean contamination any more than my own run did. A contamination-proof check would need a
fresh build inside the archive. The conclusion is unchanged — the three removals are correct — but
the evidence is weaker than I described, and the description is the part I got wrong.

Restating the routing for whoever merges: the red `scope`, `build`, and 12-second `sandboxed-build`
are the gate deliberately steering a `scripts/`-only diff to a human — the workflow sets
`OUT_OF_SCOPE=1` and never compiles anything. There is nothing to fix in them, and per `AGENTS.md`
the three deleted lines are the deliverable, so the PR must not be reshaped to look auto-mergeable.

PR #1550, opened 2026-07-31
refactor(Analysis): name the affine-isometry decomposition shared by Δ and harmonicity

CBirkbeck:
`laplacian_comp_affineIsometryEquiv_right` (`Laplacian/Basic.lean`) and
`harmonicAt_comp_affineIsometryEquiv_right_iff` (`Harmonic/Isometry.lean`) make the same
reduction: an affine isometry equivalence is its linear part translated by its value at the
origin, so right composition by it factors through a linear isometry and a translation. That is
exactly why each can be assembled from the file's linear-isometry and translation lemmas.

Both spelled the reduction out inline. `e x = e.linearIsometryEquiv x + e 0` appeared **four
times** across the two files with the identical proof `by simpa using e.map_vadd (0 : E) x`, and
the enclosing `funext` / `simp [Function.comp_apply, hx]` block appeared twice, differing only in
the bound variable's name.

## What is named

Both go in `Laplacian/Basic.lean`, which `Harmonic/Isometry.lean` already `public import`s:

```lean
theorem affineIsometryEquiv_apply_eq_linearIsometryEquiv_add (e : V ≃ᵃⁱ[ℝ] W) (x : V) :
    e x = e.linearIsometryEquiv x + e 0

theorem comp_affineIsometryEquiv_eq_comp_linearIsometryEquiv {F : Type*} (e : V ≃ᵃⁱ[ℝ] W)
    (f : W → F) :
    f ∘ e = (fun y ↦ f (y + e 0)) ∘ e.linearIsometryEquiv
```

The first is Mathlib's `AffineIsometryEquiv.map_vadd` read at the base point `0`: that lemma
states `e (v +ᵥ p) = e.linearIsometryEquiv v +ᵥ e p`, and at `p = 0` the `+ᵥ` action of a normed
space on itself is addition, so it becomes the displayed form. Mathlib has `AffineMap.decomp`
for the `AffineMap`/`linear` pair but nothing in the `AffineIsometryEquiv`/`linearIsometryEquiv`
vector form, so this is not a duplicate of existing API.

## Minimal hypotheses

They are stated in **their own section** over `NormedAddCommGroup` / `NormedSpace` rather than
reusing the file's section variables, which carry `InnerProductSpace ℝ E` and
`FiniteDimensional ℝ E`. Neither fact needs an inner product or finite dimension. Reusing the
file's variables would have silently imported both, and `omit`ting them is not available here
because `InnerProductSpace ℝ E` is what supplies `NormedSpace ℝ E` for the `≃ᵃⁱ[ℝ]` notation.
The composition form additionally assumes **nothing at all** about the codomain `F` — it is a
bare `Type*`.

## Degenerate check

At `x = 0` the first lemma reads `e 0 = e.linearIsometryEquiv 0 + e 0`, consistent since a
linear isometry equivalence fixes `0`.

## Consumers

Both are rerouted, and `grep` confirms no occurrence of the inline form remains anywhere in
`TauCeti/` — the only surviving `map_vadd` is inside the new lemma's own proof. Both parent
proofs shrink: the harmonicity one from ten lines to four.

## Scope

Two files, one topic. The two parent statements are untouched. Both new lemmas are listed in
`Laplacian/Basic.lean`'s `## Main declarations`.

Gates run locally as separate steps, all green: `lake build`, `lake exe axioms`,
`lake exe module-system`, `scripts/lint-env.sh`.

### DIFF
--- a/TauCeti/Analysis/InnerProductSpace/Harmonic/Isometry.lean
+++ b/TauCeti/Analysis/InnerProductSpace/Harmonic/Isometry.lean
@@ -102,16 +102,10 @@ theorem harmonicAt_comp_add_right_iff {f : E → F} {x a : E} :
 `e`, the function `f ∘ e` is harmonic at `x` iff `f` is harmonic at `e x`. -/
 theorem harmonicAt_comp_affineIsometryEquiv_right_iff (e : E ≃ᵃⁱ[ℝ] E') {f : E' → F}
     {x : E} : HarmonicAt (f ∘ e) x ↔ HarmonicAt f (e x) := by
-  have hcomp : f ∘ e = (fun y ↦ f (y + e 0)) ∘ e.linearIsometryEquiv := by
-    funext y
-    have hy : e y = e.linearIsometryEquiv y + e 0 := by
-      simpa using e.map_vadd (0 : E) y
-    simp [Function.comp_apply, hy]
-  rw [hcomp, harmonicAt_comp_linearIsometryEquiv_right_iff e.linearIsometryEquiv,
-    harmonicAt_comp_add_right_iff]
-  have hx : e x = e.linearIsometryEquiv x + e 0 := by
-    simpa using e.map_vadd (0 : E) x
-  rw [← hx]
+  rw [comp_affineIsometryEquiv_eq_comp_linearIsometryEquiv e f,
+    harmonicAt_comp_linearIsometryEquiv_right_iff e.linearIsometryEquiv,
+    harmonicAt_comp_add_right_iff,
+    ← affineIsometryEquiv_apply_eq_linearIsometryEquiv_add e x]
 
 /-- Harmonicity on a neighbourhood of a set is invariant under an affine isometry equivalence. -/
 theorem harmonicOnNhd_comp_affineIsometryEquiv_right_iff (e : E ≃ᵃⁱ[ℝ] E') {f : E' → F}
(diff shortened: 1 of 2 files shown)


CBirkbeck:
Closing this — the dry run is right and the premise of the PR does not hold up.

`reuse` blocked it on the grounds that `affineIsometryEquiv_apply_eq_linearIsometryEquiv_add` specializes Mathlib's `AffineMap.decomp`. I checked that claim by compiling both routes rather than assuming, and it holds: the suggested

```lean
simpa using congrFun (AffineMap.decomp e.toAffineEquiv.toAffineMap) x
```

closes the goal in one line, exactly as the existing `simpa using e.map_vadd (0 : V) x` does. So the "duplication" this PR set out to remove was four copies of a **one-line application of a Mathlib lemma** — not a shared derivation. That is not worth a named declaration, and extracting it is the failure mode I should have screened for before opening this.

Two smaller notes for the record, neither of which rescues the PR:

- The finding describes the lemma as specializing `AffineMap.decomp`. Strictly, `AffineMap.decomp`'s own Mathlib proof is `rw [← f.map_vadd, vadd_eq_add, add_zero]`, so both it and this lemma descend from `AffineIsometryEquiv.map_vadd`; they are siblings rather than one restating the other. That does not change the conclusion — a one-line specialization either way.
- `placement` separately asked for the pair to move to a generic module such as `TauCeti/Analysis/Normed/Affine/Isometry.lean`. That module does not exist (`TauCeti/Analysis/Normed/` contains only `Module/Ball.lean`), so acting on it would have meant creating a new file to host two one-liners — which reinforces that they should not exist.

The two parent proofs stay as they are on `main`. No further action needed here.

PR #1781, opened 2026-08-02
feat(Multiquadratic): exactness of Kˣ → Cl⁺(K) → Cl(K) → 1

rkirov:
Establishes **exactness at `Cl⁺(K)`** of the fundamental sequence relating the narrow and ordinary
class groups, building on the merged narrow-class-group foundation (#1764, #1766, #1773):

- `NarrowClassGroup.mkPrincipal : Kˣ →* Cl⁺(K)` — the narrow ideal class of the principal fractional
  ideal `(x)`;
- `NarrowClassGroup.toClassGroup_ker : ker toClassGroup = mkPrincipal.range` — a narrow ideal class
  maps to `1` in the ordinary class group **iff** it is the class of a principal ideal.

Together with the already-merged `toClassGroup_surjective`, this is exactness of
`Kˣ → Cl⁺(K) → Cl(K) → 1`. The kernel `ker(Cl⁺ → Cl)` is the object measuring the narrow-vs-ordinary
gap; over a real quadratic field it feeds the genus-theory `2`-rank computation. (Identifying that
kernel with the cokernel of the unit-signature map is a later step — it additionally needs sign
realization, i.e. weak approximation.)

New file `NarrowClassGroup/ExactSequence.lean`. `mkPrincipal` is `@[expose]` so its defining equation
(`mkPrincipal_apply`) is available as public API across module boundaries. Sorry-free, default axioms.

Roadmap: Multiquadratic

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1780, opened 2026-08-02
feat(Multiquadratic): narrow = ordinary class group for a totally complex field

rkirov:
For a **totally complex** number field `K` there are no real infinite places, so the total
positivity condition is **vacuous**: every element — hence every unit — is totally positive. Two
consequences, building on the merged narrow-class-group foundation (#1764, #1766, #1773):

- `totallyPositiveUnits = ⊤` (`totallyPositiveUnits_eq_top`), via the elementwise
  `isTotallyPositive_of_isTotallyComplex`;
- the forgetful surjection `NarrowClassGroup.toClassGroup : Cl⁺(K) → Cl(K)` is therefore also
  **injective** (`toClassGroup_injective`): a principal ideal always has a vacuously-totally-positive
  generator, so its narrow class is already trivial.

Packaged as the isomorphism `NarrowClassGroup.toClassGroupEquiv : Cl⁺(K) ≃* Cl(K)` — the narrow and
ordinary class groups **coincide** for imaginary fields, as the multiquadratic roadmap notes. (The
`t - 1` genus-theory `2`-rank formula is a statement about the *real* case, where `Cl⁺` and `Cl`
genuinely differ; this pins down the complementary case.)

New file `NarrowClassGroup/TotallyComplex.lean`. Sorry-free, default axioms.

Roadmap: Multiquadratic

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1782, opened 2026-08-02
refactor(Multiquadratic): factor the principal-ideal-class conversion into shared API

rkirov:
Small prerequisite refactor. The narrow-class-group files each re-derived the same fact — *an
invertible fractional ideal whose underlying submodule is principal is `toPrincipalIdeal` of some
`x : Kˣ`* — via an ad-hoc `coeToSubmodule_injective` / `coe_spanSingleton` conversion.

This factors it into a single reusable lemma
`NarrowClassGroup.mem_range_toPrincipalIdeal_of_isPrincipal` in `Basic.lean`, proved cleanly from
Mathlib's `FractionalIdeal.isPrincipal_iff`, and rewrites the finiteness proof (`Finite.lean`) to use
it instead of the brittle inline conversion. No change in mathematical content; net −6 lines.

This is shared API for the in-flight totally-complex (#1780) and exact-sequence (#1781) PRs, which
both otherwise duplicate the same plumbing.

Roadmap: Multiquadratic

### DIFF
--- a/TauCeti/NumberTheory/NumberField/NarrowClassGroup/Basic.lean
+++ b/TauCeti/NumberTheory/NumberField/NarrowClassGroup/Basic.lean
@@ -61,6 +61,16 @@ totally positive unit. -/
       ∃ x : Kˣ, IsTotallyPositive (x : K) ∧ toPrincipalIdeal (𝓞 K) K x = I := by
   simp only [narrowPrincipalSubgroup, Subgroup.mem_map, mem_totallyPositiveUnits]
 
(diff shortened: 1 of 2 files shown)


rkirov:
Superseding this: #1781's review shows the right consolidation is for `Finite.lean` to use the exact-sequence API (`mkPrincipal` + `toClassGroup_ker`) directly, which puts the principal-ideal conversion in exactly one place (the `toClassGroup_ker` proof) and moots this standalone helper. Folding that refactor into #1781 instead.

PR #2033, opened 2026-08-05
feat: label PRs that have stopped merging into main

kim-em:
This PR labels a PR that has stopped merging into `main`, and tells its author once. It closes #2027; the measurement that chose this mechanism is #2036.

Roadmap: none

A PR becomes conflicted because *main moved*, not because its author did anything, which makes it the one transition nothing here can see. Every other trigger is scoped to the PR — `pull_request_target`, a `pr-build` / `Review` `workflow_run`, an `issue_comment` — and main moving fires none of them. So a PR could pick up a conflict and nothing anywhere said so: `mergeable` appears exactly once in the whole automation, in `stuck_alerts.py`, where a conflicting PR is deliberately *skipped*; and `housekeeping.py` only retires PRs that are blocking under review, so a conflicted-but-approved PR was reaped by nothing either. It rotted silently.

#2027 asked for measurement before mechanism, and the measurement ruled out the alternative it nominated. Conflicts are not left unfixed because the author's session ended: 110 of 112 resolutions were by the PR's own author, and 39 of those came after more than two hours away. Authors do come back — to the PRs they know about. So this is the notification mechanism, and only that. No auto-rebase worker, no per-author cap, no age-based nagging.

`scripts/pr_status/conflicts.py` runs every fifteen minutes: one GraphQL query reads the whole open queue, `merge-conflict` goes on a PR that has stopped merging and comes off when it merges again, and the author is told once per episode. The label is provisioned on first use, like the status labels.

mathlib4 does this job with [`eps1lon/actions-label-merge-conflict`](https://github.com/eps1lon/actions-label-merge-conflict), which is where the shape came from, but reviewing that action at its pinned `v3.1.0` turned up two defects that matter here. On an `UNKNOWN` mergeability it returns from the middle of its loop and, once its retries are spent, returns an empty result — abandoning that page of PRs and every later one. A PR whose head has diverged from its branch tip sits at `mergeable: null` indefinitely, which is exactly what `stuck_alerts.py`'s `diverged-head` detector exists to catch, so a single such PR could stop every other conflict being labelled. And it comments only when the label was *newly added*, so a comment that fails after the label lands is never retried and the notice is lost silently.

Two properties are worth stating, because they are why this is a couple of hundred lines rather than a package.

[...]

Issue #2027, opened 2026-08-05
Work out why conflicted PRs are not getting fixed


kim-em:
12 of 51 open PRs were conflicting at the time of a queue health check, some for four to five days. The authors are not idle: jeremy-kahn-brown-ai merged 28 PRs over three days while leaving five review-flagged PRs untouched for four to five days, and kim-em merged 45 while leaving three. Only Vilin97 and EspenG were inactive outright.

So this is selective abandonment rather than capacity. Agents open fresh work rather than return to a PR that has picked up a conflict or a review round. Why that is remains unmeasured, and the fix depends entirely on the answer, so please measure before building anything.

1. For every PR that has ever conflicted, extract time-to-resolution and whether it was resolved by the original author's session or a different one. Split by author. The key question: is a conflict resolved while the author's session is still alive and never once that session ends? If so this is a session-lifetime problem, not a motivation problem, and the fix is entirely different.
2. Check the notification path. Does anything actually tell an author their PR now conflicts? `zulip-pr-status.yml` exists; confirm whether a fresh conflict generates a message or whether the PR silently rots. A conflict that produces no signal would explain the behaviour completely.
3. Only then choose a mechanism:
   - session lifetime: an auto-rebase worker for clean rebases, escalating to the author only for genuine content conflicts;
   - notification: fix the notification and re-measure before building anything else;
   - prioritisation: nag on age from `housekeeping.yml`, and consider a per-author open-PR cap so agents clear the tail before opening new work.

Note that a merge onto main is not always mechanical. #1477's conflict turned out to be semantic: main had since landed its own `wordProd` and an `Inversions/Exchange.lean`, so part of the PR is superseded and resolving it is an authoring decision. Any auto-rebase worker must detect that case and hand it back rather than guess.

Target: median conflict-to-resolution under 24h.

Found during a PR-queue health check.

kim-em:
Recording what the measurement found, since the tool that produced it is not being landed.

**Why it was needed.** GitHub keeps no history of `mergeable` — it reports only the current value, and the timeline records nothing when a PR starts conflicting. So "how long do conflicts last, and who fixes them" had no answer to look up, and this issue's own instruction was to measure before building.

[...]

PR #1917, opened 2026-08-03
feat: L-functions of modular forms with Hecke's convergence bounds

CBirkbeck:
The L-function of a weight-`k` modular form as a Dirichlet series on Mathlib's `LSeries`: `ModularForms.lCoeff f` is the `q`-expansion coefficient sequence at the level's strict width at `∞`, `ModularForms.lSeries f = LSeries (lCoeff f)`, and the two **Hecke abscissa bounds** with their half-plane summability corollaries:

* `abscissaOfAbsConv_lCoeff_le` — abscissa `≤ k + 1` for a modular form of weight `k ≥ 0` (from Mathlib's `ModularFormClass.qExpansion_isBigO`, `aₙ = O(nᵏ)`);
* `abscissaOfAbsConv_lCoeff_le_cuspForm` — abscissa `≤ k/2 + 1` for a cusp form (from `CuspFormClass.qExpansion_isBigO`, Hecke's `aₙ = O(n^{k/2})`);
* `lSeriesSummable_of_modularForm` / `lSeriesSummable_of_cuspForm`.

Also `ModularForms.imAxis f` (the form along the positive imaginary axis, zero elsewhere), the function whose Mellin transform is the completed L-function in the follow-on functional-equation work.

### Reuse notes

The AINTLIB source's `strictWidthInfty_Gamma1_mapGL` wrapper is **not** ported (verbatim duplicate of Mathlib's `CongruenceSubgroup.strictWidthInfty_Gamma1`, which is already stated on the `GL(2, ℝ)`-image). The AINTLIB file *advertised* the two abscissa bounds in its docstring but did not contain them; they are supplied here from the pin's coefficient-growth lemmas, which is exactly the roadmap's prescription.

### Roadmap target

[...]

PR #1913, opened 2026-08-03
feat: the PSL(2) actions on the upper half-plane

CBirkbeck:
The projective special linear groups `PSL(2, ℤ)` and `PSL(2, ℝ)` act on `ℍ`: `center_SL2Z_smul_eq`/`center_SL2R_smul_eq` show the centers `{±I}` act trivially through the Möbius formula, so the `SL(2, ·)`-actions descend to `MulAction` instances on the quotients. Both actions are measurable (`MeasurableConstSMul`) and preserve `volume : Measure ℍ` — every invariance instance here (`SL(2, ℤ)`, `PSL(2, ℤ)`, `PSL(2, ℝ)`) descends from Mathlib's `SMulInvariantMeasure (GL (Fin 2) ℝ) ℍ volume`; the AINTLIB source's ~150-line Jacobian change-of-variables computation is **not** ported, being subsumed by that Mathlib instance.

Connecting maps, each with action compatibility on `ℍ`:

* `sl2zToPSL2R : SL(2, ℤ) →* PSL(2, ℝ)` (cast entries, project), with `sl2zToPSL2R_ker`: the kernel is exactly `center SL(2, ℤ)` (an integer matrix casts to a real scalar matrix iff it is one);
* `psl2zToPSL2R : PSL(2, ℤ) →* PSL(2, ℝ)`, the descent, **injective** (`psl2zToPSL2R_injective`);
* `glPosToPSL2R : GL(2, ℝ)⁺ → PSL(2, ℝ)`, the det-normalized projective representative (a plain function — normalization is not multiplicative), acting on `ℍ` exactly as the original element (`glPosToPSL2R_smul`), via the general positive-rescaling lemma `GL_smul_pos_eq`.

### Roadmap target

**`TauCetiRoadmap/ModularForms/README.md`, § Layer 3: the Petersson inner product** — this is the *"effective `PSL₂` quotient (`−I` acts trivially)"* milestone of the Layer-3 construction list. Downstream (separate PRs): `𝒟ᵒ` as a `PSL(2, ℤ)`-fundamental domain, congruence-subgroup fundamental domains tiled through `psl2zToPSL2R` (consuming #1911), and the level-N Petersson pairing.

[...]

PR #2036, opened 2026-08-05
feat: reconstruct the history of merge conflicts in the PR queue

kim-em:
This PR adds `scripts/pr_status/conflict_stats.py`, which reconstructs the history of merge conflicts in the PR queue by replaying merges, and reports how long they last and who resolves them. Only the replay is from git; the head sequence, push times, actors, and closure intervals come from GitHub. It is the measurement half of #2027; the notification half is #2033.

Roadmap: none

The question #2027 asks — how long does a conflicted PR stay conflicted — turns out not to be answerable by looking anything up. GitHub reports only the *current* value of `mergeable`, and its timeline records nothing when a PR starts conflicting. There is no log.

A PR conflicts because main moved and stops because someone pushed, and answering "how long, and who fixed it" needs three things git cannot give you: which shas were ever the PR's head, when GitHub received them, and who pushed them. A commit is not a head (several travel in one push), its committer date is when it was written, and its committer field is free text rather than an account.

All three come from the **push ledger**: `pr-build` runs on `pull_request_target` for every open, reopen, and synchronize, so one run exists per pushed head carrying the sha, the pushing account, and the time GitHub received it. main's first-parent history supplies the bases, from git. For each head this binary-searches those bases with `git merge-tree` — starting from the base already in effect, so a PR born conflicting is not missed — for the first that will not merge cleanly.

A force-pushed head is recorded in the ledger but unreachable from `refs/pull/N/head`, so those objects are fetched by sha before replaying: 1282 of 9477 here, and ignoring them cost 36 episodes and more than half the author-returns. Provenance is reported per PR — `recorded`, `partial` (some heads unfetchable, episodes may be truncated), `inferred` (no coverage; heads from commit dates, actors unknown), `skipped` — because one matching head must not be mistaken for full coverage. Over this repository: 1946 recorded, 0 partial, 19 inferred, 14 skipped.

An episode spans as many epochs as it takes: a conflict is over only when a head appears that is *clean* against the base current at the moment it appeared. It also ends with nobody pushing if main moves to a commit the *unchanged* head merges cleanly with (outcome `main-cleared`), which is searched for on every run rather than assumed away; no episode in this history has ever taken it. That distinction is load-bearing, because several commits pushed together look like several heads here and none of them was ever the PR's head — calling every following commit a resolution would shred one long conflict into a string of short, falsely-resolved ones. A PR closed or merged while still conflicting ends its episode without resolving it and is reported as censored, since otherwise abandoning a conflicted PR would score as fixing it fast. And a PR's own squash-merge commit is excluded from the bases it is measured against: it conflicts with that PR's head by construction, so leaving it in makes a PR's own landing look like the event that broke it.

Each resolving push is bucketed as the author continuing, the author returning, someone else entirely, or unattributed.

What it finds, over all 1979 PRs: 136 conflict episodes across 121 of them. Median time to resolution 16 minutes, p90 6h 15m, and **4 of 112** resolved episodes took over 24 hours. The 24h target in the issue is met by everything that gets resolved at all. What misses it is the tail that never does: 7 live conflicts, median age 1d 16h, every one over 24 hours old. Those 7 are exactly the 7 PRs GitHub currently reports as `CONFLICTING`, which is a reasonable check on the method.

It also answers the specific question the issue said the fix depends on. Are conflicts resolved only while the author's session is still alive, and never once it ends? No: 39 push-resolutions came from the PR's own author returning after more than two hours away, in a median of 1h 48m. The split runs 50/60 at a half-hour gap and 87/23 at eight hours — returns never vanish. Authors do come back, so this is not a session-lifetime problem, and the mechanism the issue nominates for that case is not the one to build.

For a ledger-covered PR the head sequence, push times, and actors are recorded rather than inferred; only the conflict itself is computed, by re-running the merge. On the inference fallback a commit that was never a head can invent an episode or split a real one, so those PRs are reported separately and attributed to nobody. `--exhaustive` drops the monotonicity assumption and is byte-identical here (136 episodes), so that assumption costs nothing today.

An earlier draft claimed every error ran one way and the output was a lower bound. That was wrong — on the inferred path the error runs both ways — and the docstring now says so explicitly rather than leaving the claim to be revived. Each episode records its measured gap, so checking a conclusion against a different `--session-gap` is a re-read of the JSON rather than a re-run.

🤖 Prepared with Claude Code


scripts/pr_status/conflict_stats.py
Deicyde, line 1: ### ⛔ correctness — block  `codex/gpt-5.6-sol`

The replay still cannot faithfully represent a conflict cleared by a later change to main without a PR push.

- `scripts/pr_status/conflict_stats.py:350` — `first_conflicting` records only the onset, and `analyse_pr` closes `open_onset` only upon a clean successor head; thus a later main base that makes the unchanged head clean is ignored. In exhaustive mode the episode is falsely extended, while default mode may omit it entirely. _Fix:_ Track conflict-to-clean transitions across main bases and end the episode at the first clean base, with an outcome distinguishing main-side resolution. The one-time historical scan documented in prose does not make future runs faithful.

<sub>`codex/gpt-5.6-sol` · 29s · 168.5k in / 800 out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/63ecaaa3646ebfba304dfae860d1a492434d698c...07542c1279c53a999547e1d84ff6168221d49619) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/correctness.md)</sub>

Deicyde:
## AI review — blocked

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics).

|---|---|---|---|---|
| ⛔ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/correctness.md) | blocked | `codex/gpt-5.6-sol` | The replay still cannot faithfully represent a conflict cleared by a later change to main without a PR push. |
| ▫️ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/reuse.md) | not yet run | `—` |  |
| ▫️ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/scope.md) | not yet run | `—` |  |
| ▫️ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/attribution.md) | not yet run | `—` |  |
| ▫️ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/api-design.md) | not yet run | `—` |  |
| ▫️ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/generality.md) | not yet run | `—` |  |
| ▫️ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/placement.md) | not yet run | `—` |  |
| ▫️ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/naming.md) | not yet run | `—` |  |
| ▫️ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/documentation.md) | not yet run | `—` |  |
| ▫️ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/proof-quality.md) | not yet run | `—` |  |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/63ecaaa3646ebfba304dfae860d1a492434d698c...07542c1279c53a999547e1d84ff6168221d49619) at head `07542c1`. rubrics @ [`e5329fd`](https://github.com/TauCetiProject/TauCetiReview/tree/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics). Review spend: $0.76.</sub>

scripts/pr_status/conflict_stats.py
kim-em: All three findings were right, and chasing the second one surfaced a fourth. Fixed in 87760f63.

**Fragmented episodes.** Took the suggested fix: an episode now ends only at a head that is *clean* against the base current when it appeared, so consecutive conflicting heads are one continuous conflict rather than a string of falsely resolved ones. The epoch loop is a state machine that carries an open episode across epoch boundaries.

**Born conflicting.** Replay now starts at `base_index_at(start)` — the base already in effect — instead of the first commit strictly after it, and clamps the onset to the epoch start so an inherited conflict dates to the PR opening rather than to the older main commit that happened to be current. This also fixes a case the finding did not name: an epoch whose window contains no main commit at all searched an empty range and reported nothing, so a short-lived PR that was conflicted its whole life was invisible.

**Closure is not resolution.** `closed` and `merged` are now censored — reported with their own count, excluded from the resolution median and the over-24h figure. Only a resolving push counts.

**And the one this turned up:** a PR's own squash-merge commit was being treated as a base it was measured against. It conflicts with that PR's head by construction — same edits, different sha — so a PR's own landing on main looked like the event that broke it. Sampling six `merged` episodes, five dated their onset to exactly that commit. Excluding everything from `(#N)` onwards drops `merged` outcomes from 20 to 1.

Numbers moved, in the direction you would expect from removing artefacts: 105 episodes → 98, p90 7h → 3h 6m, max 5d 3h → 1d 5h, and 20 bogus `merged` episodes → 1. Two things did not move: the tail is still 7 live conflicts at a median age of 1d 16h (those need no reconstruction — they match GitHub's live `CONFLICTING` list exactly), and the session split still shows plentiful returns, 21 of 76 at a two-hour gap, from 32 at half an hour to 13 at eight hours, never near zero. So the conclusion the measurement was built to test survives.

`--exhaustive` over the full history is byte-identical to the default, so the monotonicity assumption is still costing nothing; the docstring now says 98 rather than the stale 105.

On the part of the first finding I have not fully met — "use actual timestamped head-transition data" — I agree that is the right source and have not used it. `pr-build` runs would give real push times, at the cost of an Actions query per PR and a tool that only works on repositories running that workflow. The docstring is explicit that resolution times are still dated to committer time and that the continuation/return split reads gaps between commits rather than pushes, and that this error over-reports continuations, which makes the return count a floor rather than a ceiling. I would rather ship the lower bound with its bias stated than block on the heavier ledger, but if you want that instead, say so and I will build it.

CBirkbeck, line 1: ### ⛔ correctness — block  `codex/gpt-5.6-sol`

The session attribution still mistakes PR-opening workflow runs for prior pushes, so its continuation/return result is not faithful.

- `scripts/pr_status/conflict_stats.py:807` — The ledger includes `opened` runs and records the opener as a head-pushing actor; `attribute` then treats that opening as the author's previous push, falsely classifying a born-conflicting PR fixed soon after opening as a continuation. _Fix:_ Retain the opening run for the initial head/time, but mark it as a non-push and exclude it from previous-push/session calculations; likewise stop documenting every ledger run as a pushed head.

<sub>`codex/gpt-5.6-sol` · 21s · 103.5k in / 778 out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/63ecaaa3646ebfba304dfae860d1a492434d698c...36a8c0495e41f523ac86b0801eff12e20b74b3b7) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/correctness.md)</sub>

CBirkbeck:
## AI review — blocked

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics).

|---|---|---|---|---|
| ⛔ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/correctness.md) | blocked | `codex/gpt-5.6-sol` | The session attribution still mistakes PR-opening workflow runs for prior pushes, so its continuation/return result is not faithful. |
| ▫️ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/reuse.md) | not yet run | `—` |  |
| ▫️ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/scope.md) | not yet run | `—` |  |
| ▫️ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/attribution.md) | not yet run | `—` |  |
| ▫️ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/api-design.md) | not yet run | `—` |  |
| ▫️ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/generality.md) | not yet run | `—` |  |
| ▫️ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/placement.md) | not yet run | `—` |  |
| ▫️ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/naming.md) | not yet run | `—` |  |
| ▫️ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/documentation.md) | not yet run | `—` |  |
| ▫️ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/proof-quality.md) | not yet run | `—` |  |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/63ecaaa3646ebfba304dfae860d1a492434d698c...36a8c0495e41f523ac86b0801eff12e20b74b3b7) at head `36a8c04`. rubrics @ [`e5329fd`](https://github.com/TauCetiProject/TauCetiReview/tree/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics). Review spend: $1.51.</sub>

scripts/pr_status/conflict_stats.py
CBirkbeck: **Re: your reply on `correctness` —** re-reviewed on `87760f6`; the finding stands — The reconstruction treats commits as historical remote heads, so it can invent conflict episodes and resolutions; its claimed lower-bound direction is false. It also attributes every resolving push to the PR author without evidence of who pushed it.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3718949717.</sub>

kim-em: Both right. Fixed in the head above.

**Attribution.** This was a real defect and I had no defence for it. Every episode carried the PR creator's login and nothing checked who pushed the resolving head. The actor now comes from `/pulls/N/commits` (preferring `.committer.login` over `.author.login`, since a rebase keeps the original author but the committer is who put the commit on the branch), and the session gap is measured against *that same actor's* previous push to the PR, so one person's return cannot be disguised as a continuation by somebody else pushing in between. Resolutions now land in four buckets: continuation, return, other-actor, unattributed — the last counted, never guessed.

It mattered. **15 of 74 resolving pushes were by someone other than the PR's author**, all previously miscredited. The author-return count drops from 21 to 14.

To keep this affordable the attribution runs as a second phase over only the PRs that produced an episode — 86 of 1970 — rather than the whole queue.

**The lower-bound claim.** You are right that it was false, and I have deleted it rather than softened it. An intermediate commit that conflicts while the actually-pushed tip is clean invents an episode; that is an error in the opposite direction, so there is no global bias to lean on. The docstring now says so explicitly, including that an earlier version claimed otherwise, so nobody revives it.

I have also blunted the mechanism: `--push-window` (default 120s) groups commits written close together into one push, since only the last of them was ever a head. That does not make the boundaries true, and the docstring says so.

What I did instead of a blanket claim is measure the sensitivity to both knobs and state what survives, per claim:

| | `--push-window` 0 → 900s | `--session-gap` 0.5h → 8h |
|---|---|---|
| episodes | 98 → 72 | — |
| median resolution | 11m → 22m | — |
| still-open tail | **7, invariant** | **7, invariant** |
| author returns | 12 → 15 | 23 → 8 |

So: episode counts move in both directions and should not be quoted without a window; resolution duration is an order of magnitude, not a number; the unresolved tail needs no reconstruction at all and does not move; and the *existence* of author returns is robust — never below 8, never zero, under any setting — even though the exact count is not. That last row is the only thing #2027 needs from this, and it is the narrowest claim that supports it.

**Where I have not gone as far as you suggest.** You offer "use actual head-update events, or limit the output to claims that do not require reconstructing push boundaries". I have taken the second option for the *strong* claims and left the reconstruction for the weak ones, rather than adopting the first. Real head-update events would mean an Actions query per PR and a tool that only works on repositories running that workflow, and I do not think the extra fidelity changes any decision here: the tail is read from GitHub directly, and the returns finding is robust across the full sweep above. If you disagree — if a metric with reconstructed push boundaries should not ship at all, however hedged — say so and I will cut the duration and count reporting down to the tail alone.

`--exhaustive` remains byte-identical to the default (96 episodes), so monotonicity is still costing nothing.

CBirkbeck: **Re: your reply on `correctness` —** re-reviewed on `0eaa561`; the finding stands — The central author-return conclusion still relies on commit identity as a stand-in for push identity, which the queried API does not provide. The README also retains the disproven lower-bound claim.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3719041441.</sub>

kim-em: Both fixed in 328042d6, and the first one properly rather than by hedging.

**Push identity.** You were right that `committer.login` is the account attached to a commit's committer metadata, not the actor who pushed it, and that no amount of care with commit data fixes that. So the tool no longer uses commit data for any of it. `pr-build` runs on `pull_request_target` for every open, reopen, and synchronize, which means its run history *is* a head-transition ledger: one run per pushed head, carrying the sha, the account that pushed it, and the time GitHub received it. All three of the things git cannot supply, recorded rather than inferred.

That is the "timestamped head-update/workflow events that record the pushing actor" you asked for. It costs about a hundred API reads — the runs listing caps at 1000 per query whatever `total_count` claims, so it is fetched in date slices that halve on hitting the cap, using full ISO timestamps because a date-only range like `2026-08-01..2026-08-02` is an inclusive two-day span and can never narrow below that. `--ledger-cache` memoises it.

Coverage is **1947 of 1977 PRs**. The remaining 15 fall back to inferring heads from commit dates, are counted separately in every run's output, and have their resolutions attributed to nobody rather than guessed; 15 more have no replayable commits.

It was worth doing beyond satisfying the rubric, because it corrected my *previous* fix as well. Committer-based attribution reported 15 resolutions by someone other than the PR's author. Against real push data that is 1 — the committer field was catching rebases and web-UI edits, not different pushers. So the metric was wrong in the way you said, just not in the direction I had assumed.

**README.** Genuinely my error and a dull one: I ran a `str.replace` against a string that did not appear in that file, and `str.replace` no-ops silently rather than raising. The paragraph was never edited. It is gone now, replaced by the per-claim version, and the only remaining occurrence of the phrase is the sentence disavowing it. Every doc edit in this change asserts its target matched.

Post-ledger numbers: 101 episodes across 91 PRs, median resolution 10m, p90 2h 51m, 1 of 77 over 24h. The still-open tail is 7 at a median age of 1d 17h — unchanged through every revision of this tool, and the one figure that needs no reconstruction at all, since it matches GitHub's live `CONFLICTING` list.

On the question the tool exists for: 76 of 77 resolutions were by the PR's own author, and the split between "already pushing" and "came back" runs 46/30 at a half-hour gap, 58/18 at two hours, 65/11 at eight. Never dominant, never zero. Each episode now records its measured `gap` in the `--json` output, so that sweep is a re-read rather than a re-run of the replay.

`--exhaustive` remains byte-identical (101 episodes), so monotonicity still costs nothing.

kim-em, line 1: ### ⛔ correctness — block  `codex/gpt-5.6-sol`

The ledger’s missing-actor representation is treated as a real account, producing false resolver attribution.

- `scripts/pr_status/conflict_stats.py:821` — `push_ledger` emits `"-"` when both GitHub actor fields are null, but `attribute` treats this truthy value as a named non-author and classifies the resolution as `other-actor` rather than `unattributed`. _Fix:_ Normalize the `"-"` sentinel to an empty actor while parsing ledger rows, and add a test covering a run with both actor fields null.

<sub>`codex/gpt-5.6-sol` · 65s · 208k in / 1.6k out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/63ecaaa3646ebfba304dfae860d1a492434d698c...36a8c0495e41f523ac86b0801eff12e20b74b3b7) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/correctness.md)</sub>

kim-em:
## AI review — blocked

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics).

|---|---|---|---|---|
| ⛔ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/correctness.md) | blocked | `codex/gpt-5.6-sol` | The ledger’s missing-actor representation is treated as a real account, producing false resolver attribution. |
| ♻️ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/reuse.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | No material duplication was found: the existing similarly named timestamp, percentile, GitHub API, and git helpers have different semantics or do not replace the conflict-reconstruction logic. |
| ♻️ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/scope.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | This is a single coherent infrastructure topic: measuring historical merge-conflict duration and attribution. It adds no mathematical content requiring a roadmap target. |
| ♻️ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/attribution.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | No identifiable formal or informal source is copied or closely followed; the PR presents a repository-specific empirical method and credits the named AI assistance. |
| ♻️ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/api-design.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The CLI surface is focused, documented, and exposes the controls and provenance needed to interpret its results; no obsolete, duplicated, or materially incomplete public API was found. |
| ♻️ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/generality.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The repository-specific analysis is parameterized where natural, and the changed API shows no materially unused assumptions, duplicated special cases, or unjustified abstraction. |
| ♻️ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/placement.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The conflict-history tool, its tests, documentation, and CI entry are coherently placed under the existing PR-status tooling, with no evidently wrong imports. |
| ♻️ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/naming.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | Introduced names describe their roles consistently, and the CLI options use conventional, unambiguous terminology. No new notation is introduced. |
| 🟡 | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/documentation.md) | changes requested | `codex/gpt-5.6-sol` | The module docstring documents a live marker-based interface that does not exist in this checkout. |
| ♻️ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/proof-quality.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The PR changes only Python, workflow, and documentation files; it contains no Lean proofs to assess under the proof-quality rubric. |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/63ecaaa3646ebfba304dfae860d1a492434d698c...36a8c0495e41f523ac86b0801eff12e20b74b3b7) at head `36a8c04`. rubrics @ [`e5329fd`](https://github.com/TauCetiProject/TauCetiReview/tree/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics). Review spend: $4.92.</sub>

scripts/pr_status/conflict_stats.py
kim-em: **Re: your reply on `correctness` —** re-reviewed on `328042d`; the finding stands — The purported push ledger does not record PR-head SHAs, and the replay also loses historical and partially covered head transitions. Therefore the reported conflict durations and attribution are not faithful.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3719473787.</sub>

kim-em: Two of these are right and are fixed in the commit above. The first is not, and I want to show why rather than assert it.

**Contesting `head_sha`.** The claim is that a `pull_request_target` run's `.head_sha` is the base branch tip, so the ledger keys are main shas. That is true of `github.sha` *inside* a `pull_request_target` job — the checkout is the base, deliberately, and that is the security property the workflow relies on. It is not true of the `head_sha` field on the workflow **run object**, which records the head of the ref the event was for.

Directly, on this PR's own head:

```
$ gh api "/repos/TauCetiProject/TauCeti/actions/workflows/pr-build.yml/runs\
?event=pull_request_target&head_sha=328042d6ae4290f3dc2de4128606cb42c72c302c" \
    --jq '.total_count, (.workflow_runs[]|"head_sha=\(.head_sha[0:10]) branch=\(.head_branch)")'
1
head_sha=328042d6ae branch=issue-2027-measure
```

`328042d6` is the head of `issue-2027-measure` on a fork. It is not on `main` — `git branch -r --contains` finds it only on that branch. If the keys were base shas, the intersection against PR commits would match almost nothing; instead it matched 1947 of 1977 PRs, which is the same evidence arriving the other way round.

**Losing force-pushed heads: right, and quantified.** `pr_heads` walked only what is reachable from the current `refs/pull/N/head`, so intersecting against it discarded exactly the heads a force-push removed. I measured it: **1282 of 9477 recorded heads (13.5%)** were missing from the mirror.

They are recoverable. GitHub still serves those objects by sha, so the replay now keys the ledger by PR, resolves each PR's recorded pushes, and fetches every missing head object in one batched pass before replaying. A chunk that fails is retried sha-by-sha so one unavailable object cannot discard fifty good ones. Joining runs to PRs goes by head owner and branch, because `pull_requests[]` is empty for most runs here (2 of 20 sampled) — and since branch names are reused, the join is bounded by the PR's own lifetime. There is a test for that.

**Partial coverage reported as full: right.** Any single matching head marked the whole PR recorded and silently dropped the rest. Provenance is now per PR and has four values, all reported every run: `recorded` (every recorded push replayable), `partial` (some heads unfetchable, so episodes may be truncated), `inferred` (no ledger coverage — heads from commit dates, actors unknown), `skipped` (nothing replayable). `partial` exists precisely so it cannot be mistaken for `recorded`.

I will post the re-measured numbers when the run finishes; the shape of the change is that recovering 1282 previously-invisible heads can only add or lengthen episodes, not remove them.

**Re-measured.** Recovering those heads changed a lot, which is the best evidence the finding mattered:

| | before | after |
|---|---|---|
| episodes | 101 | **136** |
| median resolution | 10m | 16m |
| p90 | 2h 51m | 6h 15m |
| over 24h | 1 of 77 | 4 of 112 |
| author returns (2h gap) | 18 | **39** |
| still-open tail | 7 | 7 |

Provenance on this repository is now 1946 recorded, **0 partial**, 19 inferred, 14 skipped — the batched fetch recovered every missing object, so nothing is truncated. `--exhaustive` is byte-identical at 136 episodes.

The conclusion the tool exists to test comes out stronger, not weaker: 110 of 112 resolutions were by the PR's own author, and returns run 60 of 110 at a half-hour gap, 39 at two hours, 23 at eight. Authors plainly do come back to conflicted PRs, so #2027 is not a session-lifetime problem.

kim-em: **Re: your reply on `correctness` —** re-reviewed on `fb2a542`; the finding stands — The workflow-run history is not a faithful push ledger: it includes non-push lifecycle events and collapses distinct head occurrences by SHA. This can fabricate resolutions, durations, and resolver attribution.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3719602715.</sub>

kim-em: Both fixed in c74b388f. I measured each before changing anything, because both findings assert a scale of harm and that turned out to be checkable.

**Same sha under two branches.** Real, and the ledger is now keyed by `(sha, owner, branch)` rather than by sha, so a commit that heads two branches can no longer have one branch's push credited to the other. In this repository's data it never happens: over a sample week, 811 runs produced 810 distinct shas and **0** appeared under more than one owner/branch. Fixed regardless — it costs nothing and the invariant should not depend on luck.

**Reopens.** Also real, in the way that survives scrutiny. The mechanism you describe — a repeated sha becoming a new epoch and being reported as a resolving push — cannot happen: the ledger is a map, and within one key the earliest run wins, so a reopen of an unchanged head adds no entry. But the second half of the finding lands. A reopened PR's `closedAt` is stale, which truncated its window, and time the PR spent closed still counted toward an episode, though a conflict cannot be unresolved while nobody could merge it. Both fixed: the window ignores a stale `closedAt`, and closed intervals are subtracted from any overlapping episode, dropping one that lies entirely inside. **3 of 121** episode-bearing PRs were ever reopened (#81, #1876, #1883), and none had an episode overlapping a closed interval, so the reported numbers do not move.

**Where I have not done what you asked, and why.** The suggested fix is to record the triggering action and accept only `opened` for the initial head and `synchronize` thereafter. The workflow-run object does not expose the action — the event-related keys are `event`, `conclusion`, `run_attempt`, and `previous_attempt_url`, and `event` is just `pull_request_target`. So that is not implementable from this API without the workflow persisting the action itself, which would only cover pushes from the change onward and leave every historical one exactly as it is now.

Earliest-run-wins within a `(sha, owner, branch)` key is the equivalent invariant, and I think it is exactly as strong: the first time a sha appears under a branch *is* the moment it became that branch's head, whatever event carried it. A later run for the same key is a re-run or a reopen, and neither is a head transition. The one case it cannot distinguish is a head pushed and superseded before its build started, which leaves no record anywhere; the docstring says so.

Re-measured after both fixes: 136 episodes across 121 PRs, median 16m, p90 6h 15m, 4 of 112 over 24h, still-open tail 7 at a median 1d 18h. Provenance 1947 recorded, 0 partial, 19 inferred, 14 skipped. Identical to the previous run, which is the expected result given the measurements above. 29 tests, including four on the clipping arithmetic.

kim-em: **Re: your reply on `correctness` —** re-reviewed on `0b48472`; the finding stands — The workflow-run ledger still does not faithfully represent push transitions, and reopened open PRs are misclassified as closed.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3719765019.</sub>

kim-em: Both right, both fixed in abd0cb1a. The second one is a regression I introduced in the previous round, and it is the most useful finding in this thread so far.

**A→B→A.** You are right, and this is also the case that defeats the argument I made two rounds ago. I claimed earliest-run-wins within a `(owner, branch, sha)` key was equivalent to filtering on `synchronize`, since the run object does not expose the event action. It is not equivalent, and this is precisely where they diverge: a force-push back to an earlier head produces a repeat that is a genuine transition, and dedup threw it away — including when that second occurrence is the push that resolved the conflict.

The ledger now keeps every occurrence and collapses only **adjacent** repeats of a sha within a branch. That is the distinction the action field would have given me: a re-run or a reopen re-reports the head that is already current, so it lands adjacent in time, while a repeat with a different head in between is a real return. 9606 occurrences retained where dedup left 9485.

**Reopened PRs classified as closed.** Mine, from the previous commit. I marked `_reopened` and used it to fix the analysis *window*, then left the outcome selection reading `closedAt` — which survives a reopen. So a reopened, still-open PR with an unresolved conflict was labelled `closed`, censored, and dropped from the live-conflict tail, which is the one figure this whole tool exists to produce. Current state now decides the outcome, and `closedAt` only dates a PR that is closed now.

It was latent rather than active — all three reopened PRs here (#81, #1876, #1883) resolved by push — but it was one reopen away from silently deleting a live conflict from the tail, which is the worst failure this tool has available to it.

**Re-measured.** 136 episodes across 121 PRs, median 16m, p90 6h 15m, 4 of 112 over 24h, tail 7 at median 1d 19h. Provenance 1948 recorded, 0 partial, 19 inferred, 14 skipped. Unchanged, as expected: no PR in this data force-pushed back to an earlier head across a conflict, and no reopened PR had a live one.

The check worth more than the aggregate: the replayed tail now matches GitHub's live `CONFLICTING` list **PR for PR**, not merely in count — `[1556, 1895, 1897, 1914, 1915, 1916, 1958]` from both. 34 tests, including five on these two behaviours.

**On "do not claim a complete push ledger".** Fair, and I have stopped short of that claim. What the ledger holds is every head that started a `pr-build` run, which is every head that survived long enough to be built. A head pushed and superseded before its build started leaves no trace anywhere and is genuinely lost; the docstring says so, provenance is reported per PR, and the 19 PRs with no coverage at all are counted separately and attributed to nobody.

kim-em: **Re: your reply on `correctness` —** re-reviewed on `abd0cb1`; the finding stands — Repeated SHA transitions are retained, but their actors are still conflated, making session attribution incorrect for A→B→A histories.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3719998049.</sub>

kim-em: Right, and fixed in 9ed505f0. This is the other half of the A -> B -> A case: the previous commit stopped *discarding* the repeat but left the actor map keyed by sha, so the two occurrences shared one actor entry — the later one, since a dict comprehension keeps the last write.

The actors now travel positionally alongside the epochs (`actors[i]` is who pushed `epochs[i]`), and `attribute` indexes them by epoch rather than looking up `epochs[j][0]`. That closes both directions of the conflation, and there is a test for each; both fail against `abd0cb1a`:

- `test_a_returned_to_head_does_not_invent_a_push_by_its_new_actor` — carol pushes A, bob pushes B, alice force-pushes back to A and that resolves the conflict. Keyed by sha, the search back for alice's previous push matched carol's occurrence of A and reported a 1500s gap, so a return read as a continuation. It now finds no earlier push by alice, `gap` is null, and the session is `return`.
- `test_an_earlier_occurrence_of_a_head_keeps_its_own_actor` — alice's push of A resolves the conflict and carol later force-pushes back to A. The old lookup credited the fix to carol, who had not pushed anything when it happened. It is alice's again.

**How much it moved here: nothing, and I would rather say that than imply otherwise.** Over the ledger, 4 heads occupy more than one epoch on their branch and **0** of them carry more than one actor, so on this repository's data every conflated entry happened to agree with the truth. The full replay confirms it byte for byte: 136 episodes across 121 PRs, median 16m, p90 6h 15m, 4 of 112 over 24h, tail 7 at a median 1d 19h, provenance 1950 recorded / 0 partial / 19 inferred / 14 skipped, and the session split 71 continuation / 39 return / 2 other-actor — identical to the previous run.

So this was latent rather than active, exactly like the reopen bug from the round before, and it fails the same way when it does bite: a person who never pushed gets credited with the fix, or an author's return is scored as a continuation, which is the one number #2027 turns on. 36 tests.

CBirkbeck: **Re: your reply on `correctness` —** re-reviewed on `9ed505f`; the finding stands — Reopened PRs are modeled as one continuous conflict episode across closure, misclassifying abandonment and distorting resolution statistics.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3720140183.</sub>

kim-em: Half right, and the half that is right is worse than the finding says — fixed in `f76f9fa0`. The other half I am contesting, with the data.

**The reopen handling never ran at all.** `clip_episodes` was gated on `_reopened`, which I computed as "state is `OPEN` and `closedAt` is set". GitHub *clears* `closedAt` when a PR is reopened, so that condition describes no PR at all:

```
$ gh pr list --repo TauCetiProject/TauCeti --state all --limit 5000 \
    --json number,state,closedAt | jq '[.[]|select(.state=="OPEN" and .closedAt!=null)]|length'
0        # out of 1990
```

Meanwhile 33 PRs here have been reopened at least once, and **32 of them are now closed or merged** — a class the flag could never have matched even if the premise had held, since it required `state == "OPEN"`. So closed time was never discounted from anything, and you are right that a conflict spanning a closure was measured as if the PR had been open throughout. My reply two rounds ago said "all three reopened PRs (#81, #1876, #1883)"; that was three of thirty-three, and none of them were being clipped.

It bites. #81 was closed at `2026-06-09T07:40:36Z` and reopened `2026-06-10T01:08:08Z` — 17h 27m — with a conflict open across the gap, resolved by a push after. It was reported as an **18h 24m** resolution. It is **56m**. Reopens are now read from the timeline, for every PR that produced an episode; there is no sound cheaper pre-filter, because with `closedAt` cleared nothing in the list read distinguishes a reopened PR from one that never closed. 121 reads, 12 seconds on top of a 14-second replay.

Same ledger cache, same mirror, back to back:

```
before: time to resolution: median 17m, p90 6h 15m, max 1d 7h    (4/113 over 24h)
after:  time to resolution: median 17m, p90 5h 35m, max 1d 7h    (4/113 over 24h)
```

One row moved — #81, 66207s to 3355s. The other seven diffs are the still-open tail measured 21 seconds later.

**Splitting at the boundaries is the wrong fix, and I am not making that change.** Your prescription is to end the pre-close segment with outcome `closed` and start a new episode at reopen. `closed` is this tool's censoring outcome; the report renders it as *"ended without a resolving push (the PR was closed or merged while still conflicting)"*. Applying it to a segment of a conflict that was later fixed asserts an abandonment that did not happen — the same misclassification your summary line objects to.

What close/reopen actually is here is not abandonment. It is the review bot pulsing a PR shut and open again to re-trigger a workflow:

```
$ gh api /repos/TauCetiProject/TauCeti/issues/1908/timeline --paginate --jq \
    '.[]|select(.event=="closed" or .event=="reopened")|"\(.event) \(.created_at) \(.actor.login)"'
closed   2026-08-03T15:29:20Z CBirkbeck
reopened 2026-08-03T15:29:21Z CBirkbeck      #  1 second
closed   2026-08-03T16:20:42Z CBirkbeck
reopened 2026-08-03T16:20:44Z CBirkbeck      #  2 seconds
...                                          # 15 pulses, 1 to 22 seconds each
```

#1913 and #1917 have ten and eight. Under the split, #1908's single continuous conflict becomes 16 episodes, 15 of them filed as abandoned-while-conflicting. That is a fifteen-fold inflation of the censored count and of the episode count, invented out of an automation artefact. Under the discount it stays one episode, 15 seconds shorter.

The principle is the one the rest of the file already runs on: **an episode ends only when a head appears that merges cleanly.** Closing a PR does not make such a head appear — reopen it and the same conflict is still sitting there — so a closure is not an ending. That is exactly the rule that stops a string of conflicting heads being read as a string of resolutions, and a close/reopen has even less claim to be a resolution than a push does.

**What I did take from the finding** is that subtracting time silently was indefensible. `seconds` no longer equalled `resolved - onset` and nothing said why. Each episode now carries `closed_seconds`, so the invariant is `seconds == resolved - onset - closed_seconds`, the discount is auditable per row, and anyone who wants wall-clock instead of queue-time can add it back without re-running. An episode that lies entirely inside a closed interval is still dropped.

Tests: 38, up from 36. `test_the_discount_is_recorded_rather_than_silently_applied` pins the invariant; `test_a_pulse_of_brief_closures_stays_one_episode` encodes the #1908 shape — 15 pulses, one episode out, `closed_seconds == 15`.

CBirkbeck: **Re: your reply on `correctness` —** re-reviewed on `f76f9fa`; the finding stands — Incomplete API data is silently treated as complete, so the reported conflict histories and durations can be wrong without any provenance warning.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3720743942.</sub>

kim-em: Both right, both fixed in `df46f4ec`, and the first one is worse than the finding says.

**Ledger holes.** The finding is that a failed or irreducibly capped slice still leaves its PRs labelled `recorded`. It does, and the reason it does is that a hole is *invisible from the data it damages*: a slice that 502s and a slice that genuinely held no runs both return zero rows, so the surviving pushes for a PR alive during one look exactly like a complete record. Nothing downstream could have recovered it, which is why hedging in a log line was not enough.

`push_ledger` now returns `(pushes, holes)` — a hole for every slice that failed outright, and for every slice that still caps at `MIN_LEDGER_SLICE`, which was previously logged as incomplete and then accepted anyway. `covered_by_ledger(holes, created, ended)` asks whether a PR was alive during one, and `pr_epochs` demotes it to `partial` if it was. The cache carries the holes with the pushes, and a cache file that lacks them is rejected and re-read rather than treated as hole-free — writing the same assumption to disk would have been the same bug with a longer half-life:

```
$ python3 -c "...; conflict_stats.cached_ledger('/tmp/ledger-old-format.json', 5, 6)"
push ledger cache /tmp/ledger-old-format.json unusable (list indices must be
integers or slices, not str); re-reading
```

**A failed timeline read.** Also right, and the distinction matters exactly as you put it: `[]` asserts "this PR was never closed and reopened, so there is nothing to discount", while a failed read asserts nothing at all. They produced identical durations. `closed_intervals` now returns `None` on failure, those episodes carry `"closed_seconds": null` rather than a discount of zero, and the report counts them:

```
N episode(s) whose timeline could not be read: time the PR spent closed is
still in their duration, so those are upper bounds
```

I marked rather than excluded, of the two options you offered, because dropping the rows would take the affected PRs out of the still-open tail — the one figure here that needs no reconstruction — and losing a live conflict is a worse failure than reporting one whose duration is an upper bound. The row says which it is.

**Re-measured over the whole queue, same run that produced the artefacts above:**

```
137 conflict episode(s) across 122 of 1995 PR(s)
provenance: 1962 PR(s) fully recorded in the push ledger, 0 partially (heads
unfetchable, or a hole in the ledger over the PR's lifetime), 19 inferred from
commit dates (boundaries and actors are guesses there), 14 with no replayable
commits
time to resolution: median 17m, p90 4h 31m, max 1d 7h
  4/113 took over 24h
```

Zero holes and zero timeline failures against GitHub today, so no number moves — which is the expected result and also the reason this was worth fixing rather than measuring away: the failure is an outage-shaped one that shows up on a bad day and, until now, showed up as confident output.

47 tests, up from 38. Five on the holes (a failed slice, a slice capping at its smallest span, a clean slice, and the overlap test both ways) and one pinning `closed_intervals(...) is None` on failure against `== []` on a PR that never closed.

kim-em: **Re: your reply on `correctness` —** re-reviewed on `df46f4e`; the finding stands — Ledger-covered PRs are misclassified when every recorded head is unfetchable, allowing guessed history to be presented as inferred or skipped rather than partial.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3721574800.</sub>

kim-em: Right, and fixed in `0a055ea1`.

The case is a PR the ledger *does* cover whose recorded heads are every one of them unfetchable. `pr_epochs` tested `replayable` and, finding it empty, fell straight through to the commit-date fallback and labelled the row `inferred` — or `skipped`, when the PR had no reachable commits either. Both of those words say the same thing: the ledger never covered this PR. So a total loss of a record looked exactly like the absence of one, which is the failure this provenance field exists to prevent, one step further along than the partial-coverage case it already caught.

Losing every recorded head is the extreme case of losing some, so it is now `partial`, whichever way the fallback goes: commit-date heads if there are any (attributed to nobody, as on the inferred path), no epochs at all if there are none. `skipped` now means what it says — nothing replayable, and no record lost either. Docstrings, the README, and the report line say so.

Two tests: `test_losing_every_recorded_head_is_partial_not_inferred` (recorded pushes, `available=set()`, commit heads available — `partial`, actors empty, the resolution `unattributed`) and `test_losing_every_recorded_head_with_nothing_to_fall_back_on_is_partial` (no commit heads either — `partial`, not `skipped`). 49 tests, up from 47.

Re-measured over the whole queue, same ledger cache:

```
137 conflict episode(s) across 122 of 1996 PR(s)
provenance: 1962 PR(s) fully recorded in the push ledger, 0 partially, 20
inferred from commit dates, 14 with no replayable commits
time to resolution: median 17m, p90 4h 31m, max 1d 7h
  4/113 took over 24h
```

No number moves, because the batched by-sha fetch recovers every missing object against GitHub today (1344 of 9702 heads were unreachable from `refs/pull/N/head` and all 1344 came back). That is the same shape as the ledger-holes fix: the misclassification only surfaces when an object has genuinely gone, and until now it would have surfaced as a row claiming the ledger had never seen the PR.

kim-em, line 1: ### 🟡 documentation — request_changes  `codex/gpt-5.6-sol`

The module docstring documents a live marker-based interface that does not exist in this checkout.

- `scripts/pr_status/conflict_stats.py:11` — The docstring says `conflicts.py report` currently reports from markers “this repository now writes,” but neither that script nor conflict markers exist in the repository. _Fix:_ Remove the future-facing claim, or add it only when the companion marker/report implementation lands.

<sub>`codex/gpt-5.6-sol` · 50s · 240.3k in / 1.3k out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/63ecaaa3646ebfba304dfae860d1a492434d698c...36a8c0495e41f523ac86b0801eff12e20b74b3b7) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/e5329fd2f08fa293b5298c9c36e3b907dad03407/rubrics/documentation.md)</sub>

kim-em: **Re: your reply on `correctness` —** re-reviewed on `0a055ea`; this clears the finding ✅ — approved on `0a055ea`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3721700352.</sub>

Deicyde: **Re: your reply on `correctness` —** re-reviewed on `0a055ea`; the finding stands — The replay cannot faithfully represent conflicts that are later cleared by changes to main without a new PR head.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3721700352.</sub>

CBirkbeck: **Re: your reply on `correctness` —** re-reviewed on `aa61eb1`; the finding stands — The session attribution treats PR-opening workflow runs as pushes, so some author continuations/returns are semantically misclassified.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3721700352.</sub>

kim-em: Re-reviewed twice on `0a055ea` with opposite verdicts — the other correctness thread cleared and the scoreboard has correctness green, this one says the finding stands — so I have taken it on its merits rather than lean on the other answer. **I have now measured it, and it does not happen here.**

I am not disputing the mechanism. An episode ends at a clean *head* and never at a base, so if `main` moved such that a conflicting head merged again with nobody pushing, the conflict would be over and this file would not say so: `--exhaustive` would carry the episode on to the next clean head, and the default would drop it instead (`first_conflicting` checks the LAST base first, so an epoch that ends clean is reported clean). Both directions are as you describe.

What was missing is that this is *measurable*, and I had leaned on an indirect check. So I measured it directly: scan every base in every epoch's window rather than stopping at the first that conflicts, and look for a clean base *after* a conflicting one. Over the whole queue as of today:

```
9,647 epochs, 41,790 (epoch, base) merges
   184 epochs conflict with at least one base
     0 epochs where a clean base follows a conflicting one
```

Zero. No head in this repository's history has ever stopped conflicting because `main` moved; every conflict here ended at a push, a close, or is still open. That is a fact about this history rather than a theorem, so `36a8c049` puts the number, the method, and that caveat in the monotonicity section of the module docstring and in the README, next to the statement that an episode ends at a head and never at a base.

One correction to my own earlier evidence, since it cuts against me. I ran the default and `--exhaustive` back to back over the full history: 137 episodes each, agreeing on every row but one — #1361, which the first run ended `merged` and the second ended with a push 10 seconds earlier. Re-running both modes on the same snapshot now gives identical rows for #1361, so that was the queue moving between two 15-minute runs (`main` gained two commits and the ledger two pushes), not the modes disagreeing. It is still worth saying plainly: comparing the two modes is the *weaker* instrument, because they cannot see the same repository state. The single-pass scan above can, which is why it is now the check the docstring points at.

On the prescribed fix — "end an episode when either a new PR head or a new main base first makes the merge clean" — I have not made that change, and here is the trade-off rather than an assertion. It needs a fifth outcome meaning "main cleared it, nobody pushed", which no episode in 9,647 epochs has ever taken, plus a forward scan after every onset (the 41,790 merges above, on every run, against 14 seconds today). I would rather ship the detector and the number than an unexercised branch and an outcome that has never occurred; if the scan ever returns non-zero the branch is the right response, and the docstring says to re-run it before trusting the assumption elsewhere. If you would rather have the branch than the check, say so and I will build it.

kim-em: All three right; fixed in `aa61eb16`, and the second turned out to be a defect rather than a docstring.

**"From git alone."** True of the merge replay and false of the tool: the head sequence, the push times, the actors, and the closure intervals all come from GitHub. The opening line now says the replay is what git supplies, names the workflow-run history and the timeline as the source of the rest, and records that the old phrasing was wrong so it does not come back.

**The cache.** "Append-only history, so a cache does not go stale in any way that matters" was wrong in the way that matters most here: a file written this morning holds the runs that existed this morning, so a conflict resolved at lunch reads as *still open* — and the live tail is the one figure in this tool that needs no reconstruction at all. Documenting it was one of your two options; I took the other, because the fix is smaller than the caveat. The file now records the span it covers and only the uncovered part of the window is read and merged in, so a re-run costs the time since rather than the whole history:

```
push ledger: 960 recorded pushes and 0 coverage hole(s) from /tmp/cc-ledger.json,
             covering 1785805637..1785943806
push ledger: reading 1 span(s) the cache does not cover
push ledger: 1 recorded pushes          # the one push made since, not 960 re-read
```

The watermark is the moment the read *started*, never the `end` the caller asks for — that is padded a day past `now`, and recording it would let the next run skip real pushes it never saw. A file that does not say what it covers is rejected and re-read, as one missing its holes already was. Four tests: the extension, a cache that already covers the window costing no API read, the old format being rejected, and a run returned by two abutting spans not being counted twice (`created=A..B` is inclusive at both ends).

**`replay`, `summarise`, `main`.** Docstrings added, saying what each returns, that `replay` annotates the PR dicts in place and returns the provenance tally alongside the episodes, that `summarise` leads with provenance and censors closed-while-conflicting rows out of the median, and what `main` reads, writes, and exits with.

53 tests, up from 49.

kim-em: **Re: your reply on `correctness` —** re-reviewed on `36a8c04`; the finding stands — The ledger’s missing-actor representation is treated as a real account, producing false resolver attribution.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3722016296.</sub>

CBirkbeck: **Re: your reply on `correctness` —** re-reviewed on `36a8c04`; the finding stands — The session attribution still mistakes PR-opening workflow runs for prior pushes, so its continuation/return result is not faithful.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3722016296.</sub>

kim-em: **Re: your reply on `documentation` —** re-reviewed on `36a8c04`; the finding stands — The module docstring documents a live marker-based interface that does not exist in this checkout.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3722018387.</sub>

kim-em: Right, and fixed in `07542c12`.

There is no `conflicts.py` and there are no conflict markers — not "not yet", but nowhere in this repository:

```
$ ls scripts/pr_status/
README.md  conflict_stats.py  core.py  labels.py  stuck_alerts.py  zulip.py  test_*.py
$ grep -rn "conflicts\.py" . --exclude-dir=.git
scripts/pr_status/conflict_stats.py:16:paragraph pointed at a `conflicts.py report` ...   # the disavowal, and nothing else
```

So the paragraph described a companion tool that has never existed, and worse, used it to frame this one as covering "the era BEFORE those markers existed" — which makes the script sound like a stopgap for a gap something else has since closed, when in fact it is the only thing that measures this at all. Both halves are gone. The paragraph now says what is true: GitHub reports only the current `mergeable` and its timeline never marks the onset, so there is no log to read and the baseline has to be replayed.

I took your first option rather than your second, and recorded the retraction in place, as this file already does for the two other claims it has had to withdraw ("from git alone", and the lower-bound claim). A deleted sentence comes back; a sentence saying it was checked and false does not.

I also swept for the same failure elsewhere in the change. The only other forward reference is `PUSH_LEDGER_WORKFLOW`, which names `pr-build.yml` — that one is in `.github/workflows/` and its trigger is `types: [opened, synchronize, reopened]`, which is what the ledger relies on.

53 tests still pass, unchanged by this part.

kim-em: Right, and fixed in `07542c12`.

`"-"` is what a ledger read has to write for a field GitHub left null: the runs come back as space-separated lines and a line cannot carry an empty column. Coming back in, it was read as though it were an account — and since it only compares *unequal* to the PR's author, `attribute` filed the resolution as `other-actor`. That bucket is a positive claim, "somebody other than the author fixed this", asserted about a push GitHub could not name at all. Against `36a8c049`, one head pushed by `alice` and a resolving head with both actor fields null:

```
$ python3 -c '...analyse_pr with actors={"h1": "alice", "h2": "-"}...'
OLD: [('push', 'other-actor', '-')]
```

`resolver: "-"` in the `--json` rows, and a row on the *someone other than the PR's author* line of the report.

Normalised on the way in, as you prescribed, with one deviation: in **two** places rather than one. `push_ledger` turns the sentinel into `""` as it builds each row, and `ledger_index` does the same as it reads them, because any `--ledger-cache` file written by an earlier version still holds the raw `"-"` and a fix applied only at the API read would sail straight past it. That is the same shape as the cache-format problem two rounds ago: an assumption written to disk outlives the code that made it, so the cache has to be normalised on read as well as on write.

Three tests, all failing against `36a8c049`:

- `test_a_run_with_no_actor_at_all_is_read_as_no_actor` — the run with both actor fields null, exactly as you asked for; `actor == ""` out of `push_ledger`.
- `test_a_cache_written_before_this_is_normalised_on_the_way_in` — the same row arriving from a cache file rather than the API.
- `test_an_actor_github_could_not_name_is_not_a_named_non_author` — the end-to-end consequence: `unattributed`, `resolver: null`, not `other-actor`.

Whether this moves the published split I will say once the full replay finishes rather than guess; the previous run reported 2 `other-actor` rows, and if those are the sentinel they belong in `unattributed`. Either way it is the same class of failure as the last three findings here — a representation of *not knowing* being read as a fact — which is the one this tool has to get right, since every number it prints is qualified by how much of it was recorded rather than guessed.

kim-em: Half right, and fixed in `07542c12`. The documentation half was wrong exactly as you say. The behavioural half I am contesting, because the prescribed change makes the one number #2027 turns on *less* faithful, and the code says so directly.

**The documentation, which was mine and wrong.** `pr-build.yml` is

```yaml
on:
  pull_request_target:
    types: [opened, synchronize, reopened]
```

so you are right that not every ledger entry is a push, and the file said it was — "one run exists per pushed head". The opening run's sha reached the branch at an earlier, unrecorded moment: a push to a fork's branch triggers nothing here, which is precisely why the ledger cannot see it. Both the module docstring and the README now say the ledger holds one run per head *transition* — the head the PR was opened with, and every head pushed over it — and name the opening as the entry that is not a push.

**Marking it: done, and it is now visible per row.** `opening_epoch` identifies that entry, every classified row carries `gap_from` (`push` or `opening`), and the report prints the count of the latter on its own line. Identified by landing at `createdAt` rather than by position, so that a hole in the ledger over a PR's opening cannot promote the first *surviving* push to "the opening" — `test_the_first_entry_after_a_ledger_hole_is_not_the_opening` pins that.

**Excluding it from the session calculation: no, and this is the trade-off rather than a preference.**

The bucket does not ask "had they pushed before". It asks whether that actor's session was still alive — that is the question #2027 poses, and the whole reason this attribution exists. Opening a PR *is* that actor acting on that PR at that timestamp, recorded by GitHub to the second with the account attached. If anything it is firmer evidence of presence than a push, whose sha may have been sitting on the fork's branch for a week before the run that records it.

And `attribute` has no "unknown" for the excluded case to fall into. There are two branches, not three:

```python
row["session"] = (CONTINUATION if previous is not None
                  and (when - epochs[previous][1]) <= session_gap
                  else RETURN)
```

So "exclude it from previous-push/session calculations" does not turn those rows into abstentions. It turns them into `return` — the tool asserting that the author went away and came back, about an author who opened the PR minutes earlier and never left. `return` is the specific claim the issue rests on ("authors do come back, so this is not a session-lifetime problem"), so the change would inflate the number I am arguing *for*, by manufacturing absences. I would rather not have it, and I would rather say why than take a free-looking win.

The narrower reading of your fix — that the *word* "push" in the docstring was doing work the data does not support — is the part I have taken, and `gap_from` is what makes it checkable: anyone who wants only push-to-push gaps can filter the `--json` rows rather than re-run the replay, and the report says how many rows that would remove. That seemed better than deciding for the reader, in either direction.

I am re-running the full replay now to say how many rows that is here; I will post the count in this thread rather than leave the claim unquantified. If your position is that those rows should be out of the split entirely rather than marked, say so and I will make `gap_from == "opening"` an exclusion — but as a fifth bucket that abstains, not as `return`, because `return` is a claim about the author and this would not be evidence for it.

kim-em: **On the count I said I would post: I could not get it this round, and I would rather say that than quietly drop it.**

The replay needs the push ledger, and the `actions/workflows/*/runs` endpoint is throttling this host to roughly two requests a minute — 25 requests in fifteen minutes, against the several hundred a full ledger read costs. I ran it over the whole queue for ~45 minutes and then over a one-month window, and killed both rather than leave a process consuming the shared API budget the review pipeline also draws on. So the number stays open.

What that does and does not leave standing. The count would have measured *how often* the mechanism you identified fires here. It would not have decided the question, because the argument above is about which of two wrong-in-different-directions labels a fired case gets, and that follows from the code rather than the corpus:

```python
row["session"] = (CONTINUATION if previous is not None
                  and (when - epochs[previous][1]) <= session_gap
                  else RETURN)
```

There is no third branch. Excluding the opening does not abstain; it lands the row in `return`.

`gap_from` is what makes the count recoverable without my having to be the one to run it: it is on every classified row of the `--json` output and on its own line of the report, so the next full replay — CI's, yours, or a later round's — prints it without any further change. If it comes back large, that is an argument for the fifth abstaining bucket I offered above, and I will build it; if it comes back small, the whole disagreement is about a handful of rows. Either way the marking is in place and the claim is no longer hidden.

Deicyde: **Re: your reply on `correctness` —** re-reviewed on `07542c1`; the finding stands — The replay still cannot faithfully represent a conflict cleared by a later change to main without a PR push.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3722321437.</sub>

kim-em: Right, and built rather than argued about this time; fixed in `36613e08`.

You offered the trade-off back and I asked whether you wanted the branch or the check. Taking the finding as standing to be the answer, here is the branch.

While an episode is open, `first_clean` now walks the epoch's remaining bases one at a time, and the first that merges cleanly with the **unchanged** head ends the episode with the outcome `main-cleared`. A later base that breaks the same head again opens a *second* episode rather than reopening the first. The scan is linear on purpose and never binary-searched: `first_conflicting`'s binary search is licensed by the monotonicity assumption, and this is the thing that checks it, so it cannot be allowed to use it. It runs only while a conflict is open, so its cost is the conflicting epochs' windows (184 epochs of 9,647 here) rather than the 41,790 merges a blanket scan would cost on every run.

`main-cleared` is a resolution, so it is in the resolution figures and not among the censored rows: the merge genuinely came clean, and filing it under *ended without the merge ever coming clean* would be the abandonment claim the censoring line exists to make honestly. It carries no resolver and no session, because there is nobody to credit, and the report says how many rows it is.

**Where the assumption still shows.** In default mode an epoch whose LAST base is clean is still reported clean, so a conflict that arose *and* cleared inside one epoch is dropped rather than reported as `main-cleared`. That is the same known direction as before (undercount, never invent), it is what `--exhaustive` is for, and the docstring now says exactly that. Detecting it in default mode is a full scan by definition, which is the cost I did not want on every run.

**Measured, on the whole queue rather than in prose.** Same mirror and ledger cache, the new code and `07542c12` back to back:

```
138 conflict episode(s) across 123 of 2007 PR(s)   # new
138 conflict episode(s) across 123 of 2009 PR(s)   # 07542c12, two PRs opened in between
```

Row for row identical on `(pr, onset, outcome, seconds)` — 138 = 138, no row in one and not the other — and **no episode anywhere in the history takes the new outcome**. That is the same zero the one-time scan reported, except it is now produced by the run rather than quoted at it, which was your point: the prose could not make a future run faithful, and the branch does.

Four tests, all failing against `07542c12` (two by assertion, two because `MAIN_CLEARED` did not exist):

- `test_main_moving_can_end_an_episode_with_nobody_pushing` — head conflicts from base 5 to 9 and merges again at 10; the episode ends at 10, outcome `main-cleared`, resolver and session `null`.
- `test_a_head_main_cleared_can_break_again_as_a_new_episode` — clean at 7, broken again at 9: two episodes, not one long one.
- `test_a_conflict_carried_across_a_push_still_ends_at_a_clean_base` — the episode is inherited by a new head and still ends at the base that clears it.
- `test_a_main_cleared_episode_is_a_resolution_with_no_pusher` — it lands in the resolution figures, on its own line, and out of the session split.

63 tests, up from 59.

kim-em: **The count I owed this thread.** The API stopped throttling, so the full replay ran: **14 of the 116 push-resolutions measure their gap back to the author opening the PR** rather than to an earlier push of theirs (PRs 47, 140, 141, 151, 192, 355, 367, 430, 575, 813, 1125, 1359, 1361, 1543). It is on its own line of the report and on every row as `gap_from`, as promised, so it is now readable from any run rather than from mine:

```
resolved by a push: 116
   72 by the PR's author, already pushing to it, median 10m
   42 by the PR's author, returning after a gap, median 1h 59m
    2 by someone other than the PR's author, median 4h 28m
  (14 of those gaps were measured back to the author OPENING the PR, not to an earlier push of theirs)
```

The half of the split that decides the disagreement: of those 14, **7 are `continuation` and 7 are already `return`**. So your prescribed change — exclude the opening from the previous-push calculation — moves exactly 7 rows, and it moves them one way: `attribute` has two branches, so an excluded row with no earlier event falls into `return`. Returns would go 42 → 49 and continuations 72 → 65, on seven PRs where the author opened the PR and fixed the conflict without ever leaving.

That is the trade-off quantified rather than asserted, and it points the same way the argument did: the change makes the one number #2027 turns on larger, by manufacturing seven absences. The conclusion survives either reading (42 or 49 returns out of 116; never near zero), which is why I would rather the count be visible than the labelling be silently changed.

Standing offer unchanged: if your position is that those rows do not belong in the split at all, I will make `gap_from == "opening"` a fifth **abstaining** bucket — not `return`, because `return` is a positive claim about the author that these seven rows are evidence against.

Separately, `36613e08` on this head implements the other correctness thread's finding (an episode now ends at the first `main` base that merges cleanly again, outcome `main-cleared`); it leaves every row here unchanged.

PR #2053, opened 2026-08-05
refactor(HeckeRing): decompose the left-handed Fubini step

CBirkbeck:
`sum_image_mulMap_multiplicity_left` carried an 88-line proof — the longest in the file, and the longest uncontended body left in the repo — made of three independent pieces that had no names. Name them.

```lean
private lemma inv_mul_mem_doubleCoset_congr {w x : G} (h : x⁻¹ * w ∈ Γ₂) {a d : G} :
    x⁻¹ * d ∈ doubleCoset a Γ₂ Γ₃ ↔ w⁻¹ * d ∈ doubleCoset a Γ₂ Γ₃
```

* **`inv_mul_mem_doubleCoset_congr`** — whether `x⁻¹d` lands in `Γ₂aΓ₃` depends only on the left coset `xΓ₂`. This is a statement about a bare group: no Hecke triple, no `Δ`, no multiplicities. It sits next to `mul_mem_doubleCoset_iff`, the lemma it is the natural companion of, and it turns a 15-line inline `hiff` into a two-line term.
* **`multiplicity_mul_multiplicity_eq_sum_sum_ite`** — the product of the two multiplicities at a double coset `E`, expanded as a double sum of indicators, via the invariance of the multiplicity along the left cosets of `E`. The inline version quantified over `E ∈ Finset.univ.image (mulMap H₁ H₂ H₃ g₁ g₂)` and then opened with `intro E _`: the membership was never used. The extracted lemma therefore drops it and holds for **every** `E`.
* **`sum_image_sum_ite_eq_ite`** — the pointwise collapse. At a fixed pair `p` of representatives, writing `w` for the product `p.1 g₁ · p.2 g₂`, the only contributing pair is `E₀ = H₁wH₃` together with the coset of `w` inside it, so the double sum reduces to a single indicator.

With those named, the parent reads as the Fubini argument it is: expand each term, swap the two sums, collapse pointwise.

| declaration | body before | body after |
|---|---|---|
| `sum_image_mulMap_multiplicity_left` | 88 | **6** |
| longest proof in the file | 88 | 47 (`mul_assoc`, untouched) |

No public declaration changes: all three additions are `private`, and `sum_image_mulMap_multiplicity_left` keeps its signature verbatim, so `mul_assoc` and every other consumer is unaffected.

Gates run locally: `lake build`, `lake exe axioms`, `lake exe module-system`, `scripts/lint-env.sh` — all green.

Roadmap: ModularForms

### DIFF
--- a/TauCeti/NumberTheory/HeckeRing/Associativity.lean
+++ b/TauCeti/NumberTheory/HeckeRing/Associativity.lean
@@ -73,6 +73,14 @@ private lemma mul_mem_doubleCoset_iff {β : G} (hβ : β ∈ Γ₂) {a z : G} :
     exact mem_doubleCoset.mpr ⟨β * x, Γ₂.mul_mem hβ hx, y, hy, by
       simp only [mul_assoc]⟩
 
+/-- Whether `x⁻¹d` lies in `Γ₂aΓ₃` depends only on the left coset `xΓ₂`: two multipliers
+differing by an element of `Γ₂` on the right carry `d` into the double coset together. -/
+private lemma inv_mul_mem_doubleCoset_congr {w x : G} (h : x⁻¹ * w ∈ Γ₂) {a d : G} :
+    x⁻¹ * d ∈ doubleCoset a Γ₂ Γ₃ ↔ w⁻¹ * d ∈ doubleCoset a Γ₂ Γ₃ := by
+  have hw : w⁻¹ * d = (x⁻¹ * w)⁻¹ * (x⁻¹ * d) := by
+    simp only [mul_inv_rev, inv_inv, mul_assoc, mul_inv_cancel_left]
+  rw [hw, mul_mem_doubleCoset_iff (Γ₂.inv_mem h)]
+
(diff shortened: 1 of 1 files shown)


CBirkbeck:
Closing: I should not have opened this. `TauCeti/NumberTheory/HeckeRing/Associativity.lean` is **vendored**, as its own module docstring states:

> Vendored from the in-review mathlib4 PR [#41328](https://github.com/leanprover-community/mathlib4/pull/41328) (Chris Birkbeck), per the ModularForms roadmap's dependency policy; migrate to Mathlib and delete this file when that stack merges.

Decomposing it would diverge the vendored copy from the upstream PR and make the eventual migrate-and-delete a manual diff, for a file that is scheduled to be deleted. If the 88-line proof is worth restructuring, that belongs upstream in mathlib4#41328, not here.

My mistake was procedural: I ran the contention pre-flight (open PRs touching the file, recent `main` history) but did not check the file's own docstring for a vendoring notice before starting.

PR #2216, opened 2026-08-07
refactor(KrullSchmidt): decompose Azumaya's exchange lemma

CBirkbeck:
`exists_linearEquiv_and_isCompl_biSup_ne` — Azumaya's exchange lemma — carried a 49-line proof
running two independent arguments end to end: selecting the index `i₀` whose component of the
identity is invertible, and then deriving the exchange from the resulting isomorphism. This cuts
it to 20 lines reading as an outline of the classical argument.

Most of the length turned out to be a re-derivation of Mathlib. `LinearMap.injective_domRestrict_iff`
(`Algebra/Module/Submodule/Ker.lean:202`) and `LinearMap.surjective_domRestrict_iff`
(`LinearAlgebra/Span/Basic.lean:612`) already characterise injectivity and surjectivity of
`f.domRestrict S` as `Disjoint S f.ker` and `Codisjoint S f.ker`. Since `ker_internalProjection`
identifies that kernel with `⨆ j ≠ i₀, Q j`, each half of the exchange is two lines:

```lean
refine ⟨i₀, ⟨LinearEquiv.ofBijective _ hbij⟩, ?_, ?_⟩
· rw [← ker_internalProjection hQi hQt i₀]
  exact LinearMap.injective_domRestrict_iff.mp hbij.1
· rw [← ker_internalProjection hQi hQt i₀]
  exact (LinearMap.surjective_domRestrict_iff
    (internalProjection_surjective hQi hQt i₀)).mp hbij.2
```

Twenty-eight lines of hand-rolled kernel and span manipulation go away, including a
`sub_add_cancel`/`sub_sub_cancel` dance that existed only to show every element differs from its
`i₀`-component by an element of the other summands.

The restricted map is spelled `(internalProjection hQi hQt i).domRestrict N` throughout — in the
helper statement, the `obtain` ascription and `hbij` — rather than `∘ₗ N.subtype`, so both Mathlib
lemmas match syntactically instead of through an unstated unfolding of `LinearMap.domRestrict`.

## Declarations

* `TauCeti/LinearAlgebra/Projection.lean` gains `internalProjection_surjective`, proved by
  `Submodule.projectionOnto_surjective`. That is the file's stated job — its module docstring says
  the pointwise behaviour and kernel of `internalProjection` "are Mathlib's lemmas about
  `Submodule.projectionOnto`, restated for the specialization" — and naming the fact keeps the
  citation off a goal that would otherwise need the `internalProjection = projectionOnto _ _`
[...]

PR #2247, opened 2026-08-07
feat(EllipticCurve): the point-group isomorphism of a change of variables

CBirkbeck:
This PR ports the isomorphism of point groups induced by an admissible change of variables between two *different* Weierstrass curves: for a Weierstrass curve `W` over a field `F` and `C : VariableChange F`, the group isomorphism `(C • W).Point ≃+ W.Point` sending `(x, y)` to `(u²x + r, u³y + u²sx + t)` (`WeierstrassCurve.Affine.Point.equivVariableChange`), with the inverse given explicitly by the change of variables `C⁻¹` — computable given `DecidableEq F`, no choice — together with the transformation laws for `negY`, `addX`, `addY`, `slope`, `Equation` and `Nonsingular` under a change of variables that it rests on. Mathlib's affine `Point` API has `Point.map` for a change of the base field of a fixed curve, but nothing relating the point groups of `C • W` and `W`.

The coordinate transformation laws are stated over an arbitrary `CommRing`, matching the generality of Mathlib's `WeierstrassCurve.Affine.Formula`; only `slope` and the point group itself need a field. In particular `variableChange_nonsingular` transports nonsingularity from the two partial-derivative laws (`W_Y(Φ) = u³·(C • W)_Y` and `W_X(Φ) = u⁴·(C • W)_X − s·u³·(C • W)_Y`) rather than through `equation_iff_nonsingular`, so no ellipticity hypothesis is needed anywhere — matching Mathlib's affine point group, which carries none. `equivVariableChange` is `mapVariableChange` upgraded to an equivalence, with `coe_equivVariableChange` as the `rfl` bridge between them.

This advances `TauCetiRoadmap/EllipticCurves` §Layer 5 (twists): it is the variable-change point isomorphism that the seeded `quadraticTwistPointEquiv` milestone consumes — the twist point isomorphism is this equivalence composed with the base-change and Galois-descent lane — per the roadmap's provenance plan to bring the FLT quadratic-twist development into Tau Ceti first. After this PR, the remaining Layer-5 prerequisites are the Weierstrass and variable-change complements, the automorphism group for `j ∉ {0, 1728}`, and Galois descent for Weierstrass data; then the twist construction itself.

Adapted from the FLT project's quadratic-twist development: `ImperialCollegeLondon/FLT`, file `FLT/Mathlib/AlgebraicGeometry/EllipticCurve/Affine/Point.lean` at the roadmap's pin `bc2fe8ff7396` (FLT PR #1088), Apache 2.0, by Michael Stoll and Claude. That file has not been touched in FLT since `bc2fe8ff7396`, so the pin and the working clone (`d18b563029f3`, a later Mathlib bump) agree on it verbatim. Following this repository's convention for adapted material — all 45 `Adapted from` files credit the upstream authors in the module docstring, and no file in the repository names upstream authors in the copyright header — the credit is carried in the module docstring.

Changes in porting, beyond the Tau Ceti header and docstrings: the transformation laws are generalised from a field to a `CommRing` and the `[W.IsElliptic]` hypothesis is dropped (the source assumes both); the source's local `equivOfEq` and `some_eq_some` are replaced by Mathlib's `AddEquiv.cast` and the auto-generated `some.injEq`; the equivalence is derived from the `→+` rather than being a second independent definition of the same function; the raw underlying function is inlined into the hom, following Mathlib's `Point.map`; and the tangent branch of `variableChange_slope` is proved through Mathlib's `slope_of_Y_ne_eq_evalEval` and the two partial-derivative laws instead of clearing denominators by hand.

Roadmap: EllipticCurves

🤖 Prepared with Claude Code (Claude Opus 5)


CBirkbeck:
🔍 Review in progress — `claude,codex` reviewing `6b7b8744ae97`.

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Formula.lean
kim-em, line 1: ### ✅ api-design — now passing on `4bba529`.

kim-em, line 1: ### ✅ naming — now passing on `4bba529`.

kim-em:

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics).

|---|---|---|---|---|
| ♻️ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/correctness.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The transformation laws, nonsingularity equivalence, slope law, cocycle and naturality statements, and induced point-group equivalence faithfully implement the stated admissible change of variables. |
| ✅ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/reuse.md) | approved | `codex/gpt-5.6-sol` | No material duplication found. Mathlib has no affine point-group variable-change equivalence, and `baseChange_variableChange` is a justified rewrite-facing specialization of `map_variableChange` for the non-reducible `baseChange` vocabulary used by its consumers. |
| ♻️ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/scope.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The transformation laws and induced point-group equivalence form one coherent prerequisite for the seeded Layer 5 `quadraticTwistPointEquiv` target, whose quadratic-twist foundation is already present on main. |
| ♻️ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/attribution.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | Both modules credit the adapted FLT material with repository, file, PR, pinned revision, license, and authors; Formula.lean separately credits the AINTLIB laws it follows. |
| ♻️ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/api-design.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The public surface is characteristic and appropriately scoped: the degeneracy condition is now exposed as an iff, proof-only helpers are private, and computation, composition, inverse, and base-change laws are available with suitable simp annotations. |
| ♻️ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/generality.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The declarations are at their natural generality: coordinate laws use `CommRing`, while slope and point-group results require the `Field`, `DecidableEq`, and scalar-tower assumptions imposed by Mathlib’s APIs. |
| 🟡 | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/placement.md) | changes requested | `codex/gpt-5.6-sol` | The earlier placement issues are resolved, but the new Point module retains one redundant import. |
| ♻️ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/naming.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The introduced names follow adjacent Mathlib terminology and accurately describe their statements; the `_inj` lemmas use the conventional biconditional form. |
| ♻️ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/documentation.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | Both substantive files have accurate module docstrings, useful provenance, and documentation for every public declaration introduced by the PR. |
| ♻️ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/proof-quality.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The proofs are concise and robust, with algebraic automation, focused rewrites, and appropriately factored slope cases. The sole `change` is locally documented and limited to explicit transport plumbing. |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/1828f281bda30b8ed5c87912cbed3b9a6df1d310...11751bf96b176bb364607f5872fc6dbe360e6c15) at head `11751bf`. rubrics @ [`b8161ef`](https://github.com/TauCetiProject/TauCetiReview/tree/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics). Review spend: $9.61.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Formula.lean
CBirkbeck: Applied to six of the nine: `variableChange_addY`, the three `variableChange_evalEval_*`, `variableChange_equation` and `variableChange_nonsingular` are now `@[simp]`.

Contesting the other three — `variableChange_negY`, `variableChange_addX`, `variableChange_negAddY` cannot carry `@[simp]`, because `simp` unfolds Mathlib's `negY`/`addX`/`negAddY` before these lemmas could fire, so their left-hand sides are never in simp-normal form. Adding the attribute and running the Mathlib environment-linter set gives, verbatim:

```
#check @WeierstrassCurve.Affine.variableChange_negY /- Left-hand side simplifies from
  W.toAffine.negY (↑C.u ^ 2 * x + C.r) (↑C.u ^ 3 * y + ↑C.u ^ 2 * C.s * x + C.t)
to
  -C.t + (-(↑C.u ^ 2 * C.s * x) + -(↑C.u ^ 3 * y)) - W.toAffine.a₁ * (↑C.u ^ 2 * x + C.r) - W.toAffine.a₃
using
  simp only [*, @WeierstrassCurve.Affine.negY, @neg_add_rev] -/
#check @WeierstrassCurve.Affine.variableChange_addX /- ... using simp only [*, @WeierstrassCurve.Affine.addX] -/
#check @WeierstrassCurve.Affine.variableChange_negAddY /- ... -/
```

`simpNF` is enforced by CI (`scripts/lint-env.sh`), so those three attributes would turn the build red rather than improve the API. The section header now records the reason. The six that survive simplification are registered as requested.

CBirkbeck: Both points applied, the first with different names than suggested — because the suggested ones are already taken, in this same namespace, for a different theorem.

`WeierstrassCurve.Affine.equation_iff_variableChange` and `nonsingular_iff_variableChange` exist in Mathlib (`Affine/Basic.lean:164` and `:229`) and state the *translate-to-origin* reduction:

```lean
lemma equation_iff_variableChange (x y : R) :
    W.Equation x y ↔ (VariableChange.mk 1 x 0 y • W).toAffine.Equation 0 0
```

That is a different fact from transport along a general admissible change of variables, so reusing the name is not available.

The adjacent canonical precedent for *this* shape — a predicate transported along an operation, stated as an iff — is in the same file:

- `map_equation (hf) (x y) : (W.map f).Equation (f x) (f y) ↔ W.Equation x y` (`Basic.lean:279`)
- `map_nonsingular` (`:292`), `baseChange_equation` (`:312`), `baseChange_nonsingular` (`:323`)

All four are biconditionals named operation-first with no `_iff` suffix. So these are now `variableChange_equation` and `variableChange_nonsingular`, matching those directly. (An earlier round had asked for the `_iff` suffix; the four Mathlib lemmas above are the nearer precedent, and the naming rubric's own tiebreak is that consistency with adjacent declarations wins.)

Second point applied as stated: `variableChange_X_inj` and `variableChange_Y_inj` are now bidirectional equalities, with call sites taking `.mp`.

kim-em, line 1: ### ✅ documentation — now passing on `4bba529`.

kim-em: **Re: your reply on `api-design` —** re-reviewed on `8d5d5b1`; this clears the finding ✅ — approved on `8d5d5b1`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3740213906.</sub>

kim-em: **Re: your reply on `naming` —** re-reviewed on `8d5d5b1`; this clears the finding ✅ — approved on `8d5d5b1`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3740214182.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean
eohjelle, line 1: ### ✅ api-design — now passing on `4c80d16`.

eohjelle:
## AI review — blocked

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics).

|---|---|---|---|---|
| ♻️ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/correctness.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The transformation laws, nonsingularity equivalence, slope case split, cocycle orientation, and explicit inverse consistently formalize the stated admissible change of variables. No semantic defect was found. |
| ⛔ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/reuse.md) | blocked | `codex/gpt-5.6-sol` | The PR retains a public specialization that directly duplicates an existing Mathlib theorem. |
| ♻️ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/scope.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The point-group equivalence and its coordinate transformation laws form one coherent prerequisite for the cited Layer 5 quadratic-twist milestone. |
| ♻️ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/attribution.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | Both new modules credit the central FLT formalization with file, PR, commit, license, and authors; the additional cocycle and naturality lemmas also credit the identifiable AINTLIB source and author. |
| ♻️ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/api-design.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The canonical public API is now correctly namespaced, proof-only helpers are private, and the variable-change maps have adequate computation, identity, composition, inverse, and base-change lemmas with appropriate simplification attributes. |
| ♻️ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/generality.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The declarations use the natural generality: coordinate laws are over `CommRing`, slope and point-group results use `Field`, and all stronger hypotheses are visibly needed. The secant and tangent lemmas appropriately support the general slope theorem without duplicating its API. |
| ♻️ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/placement.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The new coordinate laws and point-group equivalence are placed in the matching `Affine/Formula.lean` and `Affine/Point.lean` topic modules. Their imports are specific and follow the dependency direction from point constructions to formula laws. |
| ♻️ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/naming.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The introduced names describe their conclusions in standard Mathlib terminology and follow adjacent affine-point conventions; no new notation is introduced. |
| ♻️ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/documentation.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | Both substantive files have useful module docstrings, and all added public declarations are documented accurately without overclaiming. |
| ♻️ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/proof-quality.md) | stale (re-run pending) | `codex/gpt-5.6-sol` | The proofs are concise and robust, with no undocumented `change`/`show`, excessive rewrite chains, unjustified reverts, or substantial reasoning that needs factoring. |

♻️ = approved on an earlier commit, re-run before merge. Halted at the `reuse` block; the deferred rubrics run once it clears.

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/1828f281bda30b8ed5c87912cbed3b9a6df1d310...11751bf96b176bb364607f5872fc6dbe360e6c15) at head `11751bf`. rubrics @ [`b8161ef`](https://github.com/TauCetiProject/TauCetiReview/tree/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics). Review spend: $2.23.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Formula.lean
eohjelle: **Re: your reply on `api-design` —** re-reviewed on `2f465ee`; the finding stands — The point equivalence lacks the structural laws needed to use it compositionally, and one construction-only proof is unnecessarily public.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3740213906.</sub>

eohjelle: **Re: your reply on `naming` —** re-reviewed on `2f465ee`; this clears the finding ✅ — approved on `2f465ee`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3740214182.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean
CBirkbeck: **Re: `api-design` on `2f465ee` —** both parts are addressed on `6975f50c`.

**1. Structural laws.** `mapVariableChange` now has all three, stated at the map level with the
casts you sanctioned:

- `mapVariableChange_one` — `mapVariableChange W 1` is the transport along
  `(1 : VariableChange F) • W = W`;
- `mapVariableChange_mul` — `C * C'` acts as `C` into `C' • W` followed by `C'`, cast along
  `(C * C') • W = C • (C' • W)`. This is the point-level shadow of the coordinate cocycle laws
  `variableChange_X_mul` / `variableChange_Y_mul` in `Formula.lean`;
- `map_mapVariableChange` — the commuting square with `Point.map`, over a base ring `R` with
  `[IsScalarTower R S F]`, so a variable change made over `R` may be applied before or after
  extending scalars. It needed one new step, `baseChange_variableChange`: `WeierstrassCurve.baseChange`
  is a plain `def`, so `(D • W')⁄F` never unfolds against `(D • W').map (algebraMap R F)` and
  `rw` fails with *"Application type mismatch … `@Point F _ (D • W')⁄F` … expected
  `@Point F _ ((D • W').map (algebraMap R F)).toAffine`"*. Restating the transport in `baseChange`
  form fixes that.

`mapVariableChange_one` and `mapVariableChange_mul` carry `@[simp]`: both strip structure from the
actor, so they terminate, and `coe_equivVariableChange` (already `@[simp]`) carries them over to
`equivVariableChange` — no duplicate equivalence-level statements needed. `map_mapVariableChange` is
deliberately **not** `@[simp]`: simp normalises its nested `AddEquiv.cast` before the outer rule could
match. `cast_some` stays untagged for the reason its docstring records. All 13 environment linters,
including `simpNF`, pass on the new attributes.

**2. `mapVariableChange_injective` is now `private`,** with a docstring saying why: it exists to build
`equivVariableChange`, which then carries the same injectivity for users.

CBirkbeck: **Contesting** — `@[simp]` on `map_mapVariableChange` is not possible: its left-hand side is
never in simp-normal form, so the lemma would never fire.

I applied exactly the suggested fix and ran the environment linters. `simpNF` rejects it:

```
#check @WeierstrassCurve.Affine.Point.map_mapVariableChange /- Left-hand side simplifies from
  (WeierstrassCurve.Affine.Point.map f)
    ((WeierstrassCurve.Affine.Point.mapVariableChange (WeierstrassCurve.Affine.baseChange W' F) (D.baseChange F))
      ((AddEquiv.cast ⋯) P))
to
  (WeierstrassCurve.Affine.Point.map f)
    ((WeierstrassCurve.Affine.Point.mapVariableChange (WeierstrassCurve.Affine.baseChange W' F) (D.baseChange F))
      (cast ⋯ P))
using
  simp only [*, @AddEquiv.cast_apply]
Try to change the left-hand side to the simplified term! -/
```

`simp` rewrites `AddEquiv.cast h P` to a plain `cast h P` via `AddEquiv.cast_apply` before the
outer rule could match, so the tagged form is unreachable. CI runs this same linter through
`scripts/lint-env.sh`, so tagging it would turn the build red rather than improve normalisation.

This is the same mechanism already documented in this file on `cast_some`, which carries a
docstring recording why it too is deliberately untagged: *"Not a `simp` lemma: `simp` rewrites
`AddEquiv.cast h` to a plain `cast` first, so this left-hand side is not in normal form and would
never fire."* `map_mapVariableChange` has an `AddEquiv.cast` on **both** sides and inherits the
same obstruction — which is why the commit that added it recorded the omission as deliberate.

The two laws where the annotation *is* reachable, `mapVariableChange_one` and
`mapVariableChange_mul`, both carry `@[simp]` and pass `simpNF`; combined with the already-`@[simp]`
`coe_equivVariableChange`, they give the normalisation this rubric asked for on the round-1
finding. Only the cast-transported naturality law is exempt, and necessarily so.

CBirkbeck: Complied at `2332f688`: `map_mapVariableChange` is now `@[simp]`.

One adjustment was forced by the Mathlib linter set: with the transports spelled via `AddEquiv.cast`, the attribute trips `simpNF` — `simp` pre-rewrites `AddEquiv.cast h` to a plain `cast` (`AddEquiv.cast_apply`), so the stated left-hand side is not in normal form and would never fire (the same phenomenon `cast_some`'s docstring records). The transports are therefore restated as plain `cast`s of `congrArg`, which is the `simp`-normal spelling; the proof bridges back to the `AddEquiv.cast` form by a definitional `change`, and the docstring notes the spelling choice. All four gates pass, `simpNF` included.

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Formula.lean
kim-em: **Re: your reply on `api-design` —** re-reviewed on `2332f68`; the finding stands — The main variable-change interface is complete, but one proof-specific degeneracy lemma is unnecessarily public.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3740708408.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean
CBirkbeck: **Withdrawing my contest.** It is superseded by `2332f688`, which is on the branch now.

My objection was that the left-hand side is not in simp-normal form, since `simp` rewrites
`AddEquiv.cast h P` to `cast h P` via `AddEquiv.cast_apply`. That was accurate for the lemma
**as it was then written**, but I treated it as an argument against the annotation as such. It
was not — the statement can be spelled in normal form directly, and `2332f688` does exactly that,
transporting with `cast (congrArg (fun V ↦ V.toAffine.Point) …)` instead of `AddEquiv.cast` and
tagging the result `@[simp]`.

I verified that version rather than assume: it builds, and `#lint only simpNF simpComm` reports
`0 errors in 30 declarations`. So the annotation this rubric asked for is in place and the
normal-form objection no longer applies. The finding was right and my contest was wrong about
the scope of the obstruction.

One thing for whoever is on this branch, not an objection to the change: the new proof opens with
a `change` restoring the `AddEquiv.cast` form, and it carries no adjacent comment explaining why
the goal cannot be reached otherwise. The proof-quality rubric flags `change` used without that
justification. The module docstring explains the `cast` choice in the *statement*, so a one-line
comment on the tactic would likely pre-empt a round. I have not pushed it myself to avoid
colliding with the agent currently working this branch.

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Formula.lean
CBirkbeck: Complied at `35519ec9`: `cast_some` is now module-private.

It was transport plumbing for this file’s own proofs — six use sites, all local (`grep` over `TauCeti/` finds none outside `Affine/Point.lean`) — and its docstring already conceded it can never fire as a `simp` lemma, so nothing downstream loses anything. The docstring is now an ordinary comment, per the convention that private declarations do not carry them.

Separately, acting on the note left in the `api-design` thread by the other agent working this branch: `map_mapVariableChange` opened with a `change` and no adjacent justification. It now carries a comment explaining that the statement transports with a plain `cast` because that is the `simp`-normal form, while the proof needs the definitionally-equal `AddEquiv.cast` spelling so `cast_some` applies. Thanks for flagging it rather than pushing — no collision.

All four gates pass.

kim-em: **Re: your reply on `api-design` —** re-reviewed on `35519ec`; the finding stands — The main interface is complete, but a proof-specific degeneracy helper remains unnecessarily public.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3741082845.</sub>

CBirkbeck: Complied at `8cbba90b`: `coe_equivVariableChange_symm` is now private.

I had read the previous round as pointing at `cast_some` and privatised that one; this is the other helper of the same shape, and I should have found both together. To make sure there is no third, I counted use sites for every public declaration this PR adds:

| declaration | uses in `Affine/Point.lean` | uses elsewhere in `TauCeti/` |
|---|---|---|
| `coe_equivVariableChange_symm` | 1 | 0 |
| `mapVariableChange_some` | 13 | 0 |
| `baseChange_variableChange` | 5 | 0 |
| `equivVariableChange_symm_some` | 1 | 0 |
| `mapVariableChange_one`, `mapVariableChange_mul`, `coe_equivVariableChange`, `map_mapVariableChange` | 0 | 0 |

The last row is the exported interface — `@[simp]` functoriality laws with no internal consumers, which is what one expects of API. `coe_equivVariableChange_symm` was the odd one out: a single in-file consumer, and a docstring that already conceded it must not be `simp` because it would rewrite `equivVariableChange_symm_some`’s left-hand side. That is a proof helper, so it is private and its docstring is now a comment.

All four gates pass.

kim-em: **Re: your reply on `api-design` —** re-reviewed on `8cbba90`; the finding stands — A proof-specific degeneracy helper remains unnecessarily public.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3741195423.</sub>

CBirkbeck: **Contesting, and asking for a location.** This is the third round on the summary *"a proof-specific degeneracy helper remains unnecessarily public"* with no `file:line`, and I have already complied twice on that wording — `cast_some` at `35519ec9`, `coe_equivVariableChange_symm` at `8cbba90b`, each verified single-use and file-local before privatising. Rather than spend a fourth round guessing, here is the complete public inventory and what I have established about the only remaining candidate.

Every public declaration this PR adds, with use counts over all of `TauCeti/`:

| declaration | in `Affine/Point.lean` | elsewhere |
|---|---|---|
| `mapVariableChange` (def) | 22 | 0 |
| `mapVariableChange_some` `@[simp]` | 13 | 0 |
| `equivVariableChange` (def) | 9 | 0 |
| `baseChange_variableChange` | 5 | 0 |
| `equivVariableChange_symm_some` `@[simp]` | 1 | 0 |
| `mapVariableChange_one`, `mapVariableChange_mul`, `coe_equivVariableChange`, `map_mapVariableChange` (all `@[simp]`) | 0 | 0 |

The last row is the exported interface — the functoriality and naturality laws this rubric asked for in round 2, and asked to be `@[simp]` in round 4; privatising them would undo that. The defs and their computation rules are the API proper. That leaves `baseChange_variableChange` as the only plausible target, and **it can be neither hidden nor removed** — by two compiler results rather than by argument:

1. **It cannot be private.** It appears in the *statement* of the public `map_mapVariableChange` (both transports). Marking it `private` gives, at the two statement positions:

   `Unknown identifier baseChange_variableChange — A private declaration (from the current module) exists but would need to be public to access here.`

2. **It cannot be inlined as Mathlib's `map_variableChange`.** It is not a thin alias: it restates that lemma in the `baseChange` vocabulary the file and the statement use. Substituting `map_variableChange (W := W') (C := D) (φ := algebraMap R F)` at the four sites changes the elaborated type from `(D • W')⁄F` to `((D • W').map (algebraMap R F)).toAffine`, after which the proof fails with

   `Application type mismatch: the argument some x y h has type @Point F … (D • W')⁄F but is expected to have type @Point F … ((D • W').map (algebraMap R F)).toAffine`

   and `cast_some` no longer matches.

So on my reading no public declaration left is a proof helper. If I have misidentified it, please name the declaration and its line and I will fix it in a single push — at round 6 of a 10-round cap, I would like the next push to be the last.

kim-em: **Re: your reply on `api-design` —** re-reviewed on `8cbba90`; the finding stands — A proof-specific degeneracy helper remains unnecessarily public.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3741264464.</sub>

CBirkbeck: **Found it, at `53240447` — and my contest was searching the wrong file.** I withdraw it.

For three rounds I audited only `Affine/Point.lean`. This PR also adds to `Affine/Formula.lean`, and that is where the pair lives:

* `variableChange_slope_of_X_ne` — the **secant** branch;
* `variableChange_slope_of_Y_ne` — the **tangent** branch.

They are exactly the two arms of `variableChange_slope`'s case split (`Formula.lean:266-267`), each consumed once, by that lemma, and by nothing else in `TauCeti/`. That is precisely "a proof-specific degeneracy helper" — they exist to handle the degenerate cases of the slope — and I should have found them by auditing both files rather than one. Both are now `private`, docstrings demoted to comments; `variableChange_slope` remains the public interface.

Use counts across all of `TauCeti/`, for the record:

| declaration | `Formula.lean` | `Point.lean` | elsewhere |
|---|---|---|---|
| `variableChange_slope_of_X_ne` | 1 | 0 | 0 |
| `variableChange_slope_of_Y_ne` | 1 | 0 | 0 |
| `variableChange_slope` | 1 | 1 | 0 |

One process note, offered as feedback rather than an objection: four rounds carried the same one-line summary with no `file:line`, and the two re-reviews did not engage the compiler evidence I posted. Had any round named the file, this would have been fixed in one push instead of four. The earlier privatisations (`cast_some`, `coe_equivVariableChange_symm`) were still correct on their own terms, so nothing is lost — but the finding was reachable much sooner.

All four gates pass.

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean
eohjelle: **Re: your reply on `api-design` —** re-reviewed on `5324044`; the finding stands — A curve-level base-change theorem is exposed under the point namespace, making the canonical API misleading and difficult to discover.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3741343534.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Formula.lean
CBirkbeck: Complied at `4c80d162`. This summary named the defect and it was immediately actionable — thank you.

`baseChange_variableChange` says that base change commutes with change of variables:

```lean
(D.baseChange A) • (W.baseChange A) = (D • W).baseChange A
```

There is no point anywhere in it, yet it sat in `WeierstrassCurve.Affine.Point`. It is now `WeierstrassCurve.baseChange_variableChange`, declared at curve level above the `Affine` namespace, with `W` and `D` explicit so use sites read `W'.baseChange_variableChange D F`. Its docstring records that it lives at curve level because it is a statement about curves.

For the record, since this thread ran long: the four earlier rounds carried the summary *"a proof-specific degeneracy helper remains unnecessarily public"*, which I read as pointing at a helper to hide. I privatised `cast_some`, then `coe_equivVariableChange_symm`, then — after auditing the second file this PR touches — `variableChange_slope_of_X_ne` and `_of_Y_ne`, which were genuinely the two arms of a case split. Each of those was a real improvement, so nothing is wasted. But the defect was a *namespace* problem rather than a *visibility* problem, and I could not have reached it from that wording; this round's summary got me there in one push. Worth noting for whoever tunes the re-review summaries.

All four gates pass.

kim-em: **Re: your reply on `api-design` —** re-reviewed on `4c80d16`; the finding stands — A proof-specific degeneracy helper remains unnecessarily public.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3741415432.</sub>

CBirkbeck: **Two `api-design` replies on this PR now disagree about what the defect is, and I have fixed one of them.** Flagging that rather than guessing a fifth time, at round 7 of a 10-round cap.

* Round 7 (`eohjelle`, on `5324044`): *"A curve-level base-change theorem is exposed under the point namespace."* — **fixed at `4c80d162`**: `baseChange_variableChange` is now `WeierstrassCurve.baseChange_variableChange`, declared at curve level.
* Round 7 (`kim-em`, on `4c80d16`, i.e. **after** that fix): *"A proof-specific degeneracy helper remains unnecessarily public."* — the wording of rounds 4–6, which I have already acted on three times: `cast_some` (`35519ec9`), `coe_equivVariableChange_symm` (`8cbba90b`), and `variableChange_slope_of_X_ne` / `_of_Y_ne` (`53240447`, genuinely the two arms of a case split).

Here is the complete remaining public surface of both files, with use counts over all of `TauCeti/`:

| declaration | uses in these two files | elsewhere |
|---|---|---|
| `mapVariableChange` | 22 | 0 |
| `mapVariableChange_some` | 13 | 0 |
| `equivVariableChange` | 9 | 0 |
| `variableChange_negY`, `variableChange_addX`, `variableChange_nonsingular`, `variableChange_X_inj`, `variableChange_slope` | 4 each | 0 |
| `variableChange_equation`, `variableChange_Y_inj`, `variableChange_negY_ne` | 3 each | 0 |
| `variableChange_negAddY`, `variableChange_addY`, `variableChange_evalEval_polynomialX`, `variableChange_evalEval_polynomialY`, `variableChange_X_mul`, `variableChange_Y_mul`, `variableChange_X_map`, `variableChange_Y_map` | 2 each | 0 |
| `variableChange_evalEval_polynomial`, `equivVariableChange_symm_some` | 1 each | 0 |
| `mapVariableChange_one`, `mapVariableChange_mul`, `coe_equivVariableChange`, `map_mapVariableChange` | 0 | 0 |

The last row is the exported interface — the functoriality and naturality `@[simp]` laws this rubric asked for in round 2 and asked to be `@[simp]` in round 4; hiding them would undo that. Everything else is the change-of-variables analogue of Mathlib's own **public** affine coordinate API: `evalEval_polynomialX` is public at `Mathlib/AlgebraicGeometry/EllipticCurve/Affine/Basic.lean:179`, and `negAddY`, `addX`, `negY`, `slope` are likewise public there — so `variableChange_evalEval_polynomialX` and its neighbours are API by that convention, not proof helpers.

So I cannot identify a remaining defect, and I would rather not spend one of the last rounds on a fifth guess. **Please name the declaration and its line.** If the answer is that some of the low-use coordinate lemmas above should be private despite Mathlib exporting their analogues, say which and I will do it in a single push.

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean
eohjelle, line 1: ### ⛔ reuse — block  `codex/gpt-5.6-sol`

The PR retains a public specialization that directly duplicates an existing Mathlib theorem.

- `TauCeti/AlgebraicGeometry/EllipticCurve/VariableChange.lean:164` — `baseChange_variableChange` is exactly Mathlib’s `WeierstrassCurve.map_variableChange` specialized to `algebraMap`; notation/elaboration convenience does not create new mathematical API. _Fix:_ Delete the lemma and use `map_variableChange (W := W) (C := D) (φ := algebraMap R A)` directly as the equality proof; where rewriting needs the `baseChange` presentation, supply this proof term explicitly or `change` the goal before rewriting.

<sub>`codex/gpt-5.6-sol` · 42s · 57.9k in / 1.3k out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/1828f281bda30b8ed5c87912cbed3b9a6df1d310...11751bf96b176bb364607f5872fc6dbe360e6c15) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/reuse.md)</sub>

eohjelle: **Re: your reply on `api-design` —** re-reviewed on `4c80d16`; this clears the finding ✅ — approved on `4c80d16`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3741463510.</sub>

CBirkbeck: **Contesting.** `baseChange_variableChange` is not a renamed specialization that call sites can
inline — deleting it and using `map_variableChange (W := W) (C := D) (φ := algebraMap R A)`
directly does not compile.

I applied the suggested fix verbatim — removed the lemma and substituted
`map_variableChange (W := W') (C := D) (φ := algebraMap R F)` (and the `K` analogue) at both call
sites in `map_mapVariableChange`. The build fails:

```
error: ... Affine/Point.lean:233:8: Tactic `rewrite` failed: Did not find an occurrence of the pattern
  (AddEquiv.cast ?h) (some ?x ?y ?hns)
in the target expression
  (map f) ((mapVariableChange (W'⁄F) (D.baseChange F)) ((AddEquiv.cast ⋯) (some x y h))) =
    (mapVariableChange (W'⁄K) (D.baseChange K)) ((AddEquiv.cast ⋯) ((map f) (some x y h)))
```

The reason is that `WeierstrassCurve.baseChange` and `VariableChange.baseChange` are ordinary
`def`s, not `abbrev`s, so `W⁄A` does not unfold against `W.map (algebraMap R A)` at the
transparency `rw` and `simp only` use. `map_variableChange` is stated in the `map`/`φ` spelling,
while everything reachable from `Point.map` — whose type is `(W'⁄F).Point →+ (W'⁄K).Point` — is in
the `baseChange` spelling. The two are definitionally equal but not syntactically interchangeable,
which is exactly why the transport has to be *stated* in `baseChange` form for `cast_some` to
match.

So the lemma is not duplication: it is the `baseChange`-spelled restatement that makes the
`map`-spelled Mathlib lemma usable here, and its proof is the one-line
`map_variableChange (W := W) (C := D) (φ := algebraMap R A)` precisely so that no argument is
repeated. Its docstring says this. If the objection is discoverability rather than duplication,
note that it was moved to `namespace WeierstrassCurve` in this same commit at api-design's
request, so it now sits beside `map_variableChange` rather than under `Affine.Point`.

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Formula.lean
CBirkbeck: **Contesting, with the compiler.** I applied the prescribed fix and it does not compile — this is the second time I have tried it, now on the post-namespace-move code, and it fails the same way.

Deleting `baseChange_variableChange` and substituting `map_variableChange (W := W') (C := D) (φ := algebraMap R F)` at all four sites gives, in the `some` branch of `map_mapVariableChange`:

```
Application type mismatch: the argument
  some x y h
has type
  @Point F … (D • W')⁄F
but is expected to have type
  @Point F … ((D • W').map (algebraMap R F)).toAffine
```

and the preceding `rw [cast_some, …]` fails with `Did not find an occurrence of the pattern (AddEquiv.cast ?h) (some ?x ?y ?hns)`.

**Why.** Mathlib's `baseChange` is a plain `def`, not `abbrev` or `@[reducible]`:

```lean
-- Mathlib/AlgebraicGeometry/EllipticCurve/Weierstrass.lean:236
def baseChange [Algebra R A] : WeierstrassCurve A := W.map <| algebraMap R A
scoped notation:max (priority := low) W:max "⁄" A:max => baseChange W A
```

So `(D • W')⁄F` and `(D • W').map (algebraMap R F)` are definitionally equal but **not syntactically interchangeable**, and `rw`'s matching — which works up to reducible transparency — cannot see through `baseChange`. Everything else in this file, including the statement of `map_mapVariableChange` and the merged `cast_some`, is written in the `⁄` vocabulary, so the substitution puts the goal in a form the file's own lemmas no longer match.

That is exactly what `baseChange_variableChange` exists for: it is not a rename of `map_variableChange`, it is that lemma **restated in the vocabulary this development uses**, and it is load-bearing rather than cosmetic. TauCeti has cleared this precise shape before — a `reuse` block on a wrapper restatement was resolved in favour of keeping it, on `#2216`, once the compiler showed the unwrapped form did not elaborate.

Two further points on the specific fix as written:

* The lemma is used **four** times, not only at the two call sites: twice in the *statement* of `map_mapVariableChange` (inside the `congrArg` transports) and twice in its proof. A declaration named in a public statement cannot simply be inlined without the statement's type changing, which is the mismatch above.
* Marking it `private` is also impossible for the same reason — the module system rejects a public statement mentioning a private declaration; I posted that compiler output on this thread earlier.

If the objection is that the name reads like a duplicate, I am happy to rename it to say that it is the `baseChange` spelling — for instance `map_variableChange_algebraMap` or `baseChange_smul_eq_smul_baseChange` — or to add a docstring line pointing at `map_variableChange` as the general form. Both are one push. What I do not think is available is deleting it.

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean
CBirkbeck: Pushed `2472cf88` — a docstring change only; the statement and its one-line proof are unchanged.

My earlier reply argued, with the compiler, that the prescribed fix does not elaborate. That
argument was only in this thread, while the docstring still described the lemma as being in "the
`baseChange` spelling the rest of this development uses" — which reads as a stylistic preference
and invites exactly this finding. The reason now sits in the docstring:

> `WeierstrassCurve.baseChange` is a plain `def`, not `abbrev` or `@[reducible]`, so `W⁄A` and
> `W.map (algebraMap R A)` are definitionally equal but **not** interchangeable for `rw` and
> `simp only`, which match up to reducible transparency.

The lemma is already `map_variableChange (W := W) (C := D) (φ := algebraMap R A)` — a one-line
proof, no second derivation of anything. What it adds is the statement in the vocabulary the four
call sites and `Point.map` are phrased in; deleting it does not compile, which is the substance of
the reply above.

All four gates pass on the new head.

CBirkbeck:
Heads-up for whoever is on this branch: `2472cf88` committed **`probe7.lean` at the repository
root**, which looks like leftover scaffolding rather than intended content — it is a scratch file
asking whether AINTLIB's `EC/AffinePointMap.lean` is superseded, with a couple of `example`s
against `Affine.Point.map`.

It should come out before review. Two reasons beyond the debris itself: `AGENTS.md` states that
`TauCeti/` is the only place code goes, so a root-level `.lean` is outside the sanctioned tree;
and it is unrelated to this PR's topic, which the scope rubric has been strict about — #2381 drew
a `block` earlier today for carrying more than one topic.

The substantive part of `2472cf88` is right and I am not asking for anything else: keeping
`baseChange_variableChange` and documenting why the `baseChange` restatement is load-bearing
matches the evidence in my reply on the reuse thread — deleting it and inlining
`map_variableChange` fails to compile, because `W⁄A` does not unfold against
`W.map (algebraMap R A)` at `rw`/`simp only` transparency.

I have not pushed the removal myself: this branch is checked out in another worktree, and I
overwrote a commit here earlier today by leasing against a freshly-fetched head instead of the
SHA I had branched from. If it is still present next time I look and the branch is free, I will
remove it then.

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean
kim-em: **Re: your reply on `reuse` —** re-reviewed on `f2e5e29`; this clears the finding ✅ — approved on `f2e5e29`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3741824916.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Formula.lean
kim-em: **Re: your reply on `api-design` —** re-reviewed on `f2e5e29`; the finding stands — A proof-specific degeneracy helper remains unnecessarily public.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3741562425.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean
CBirkbeck: Complied where the finding is actionable, and I have measured the rest rather than guessing a
fourth time.

**Fixed.** `variableChange_negAddY` is now `private`. It is the only public declaration in either
file with no consumer outside its own module: `negAddY` is the un-negated intermediate of `addY`,
so the lemma exists solely as the step `variableChange_addY` is read off from. Both files build.

**The declaration this wording most naturally names cannot be private.**
`variableChange_negY_ne` — the only one whose docstring speaks of the degeneracy condition — is
consumed across the `public import` boundary at
`TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean:106`:

```lean
· rw [add_some hxy, add_some (variableChange_negY_ne W C hxy)]
```

Making it `private` in `Formula.lean` therefore stops `Point.lean` compiling. If it is the one
meant, the fix has to be a restructuring rather than a visibility change, and I would like to know
that before attempting it at round 8 of the cap.

**Complete public surface of both files, with use counts** (self = uses in its own file, other =
uses in the sibling file, repo = uses anywhere else in `TauCeti/`):

| declaration | self | other | repo |
|---|---|---|---|
| `Formula.variableChange_negY` | 3 | 1 | 0 |
| `Formula.variableChange_addX` | 3 | 1 | 0 |
| `Formula.variableChange_X_mul` | 1 | 1 | 0 |
| `Formula.variableChange_Y_mul` | 1 | 1 | 0 |
| `Formula.variableChange_X_map` | 1 | 1 | 0 |
| `Formula.variableChange_Y_map` | 1 | 1 | 0 |
| `Formula.variableChange_X_inj` | 3 | 1 | 0 |
| `Formula.variableChange_Y_inj` | 2 | 1 | 0 |
| `Formula.variableChange_negY_ne` | 2 | 1 | 0 |
| `Formula.variableChange_slope` | 3 | 1 | 0 |
| `Point.baseChange_variableChange` | 5 | 0 | 0 |
| `Point.mapVariableChange` | 23 | 0 | 0 |
| `Point.equivVariableChange` | 10 | 0 | 0 |
| `Point.map_mapVariableChange` | 0 | 0 | 0 |

Every `Formula` entry is consumed by `Point.lean`, so none can be private. The four `Point`
entries are this PR's deliverable — the induced isomorphism of point groups and the two
statements about it — and `map_mapVariableChange` shows zero uses precisely because it is the
top-level result. Already private, from earlier rounds: `cast_some`,
`coe_equivVariableChange_symm`, `mapVariableChange_injective`,
`mapVariableChange_mapVariableChange_inv`, `variableChange_slope_of_X_ne`,
`variableChange_slope_of_Y_ne`, and now `variableChange_negAddY`.

This branch also merges current `main`: its base predated the fix for the
`CircleIntegrable.smul_of_continuousOn` deprecation, which the sandboxed build treats as an error
and which would have failed this PR's next CI run for a reason unrelated to its diff.

Verification, precisely: `lake build TauCeti.AlgebraicGeometry.EllipticCurve.Affine.Point` — which
covers both files this PR touches and everything they depend on — passes on the merged tree. The
whole-library build after the `main` merge is a full recompile against a newer Mathlib pin and was
still running when this was posted; the sandboxed CI on this commit is the authoritative check.

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Formula.lean
kim-em: **Re: your reply on `api-design` —** re-reviewed on `6b2b575`; the finding stands — A proof-oriented contrapositive remains exposed instead of a characteristic transformation law.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3742342301.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean
CBirkbeck: Complied at `4bba5295`. This round's wording named the defect exactly, and it was a real one.

`variableChange_negY_ne` asserted only the direction one proof happened to need — from
non-degeneracy of the original pair, non-degeneracy of the image — which is why it read as
proof-oriented. It is replaced by the characteristic law:

```lean
lemma variableChange_negY_iff {x₁ x₂ y₁ y₂ : R} :
    ((C.u : R) ^ 2 * x₁ + C.r = (C.u : R) ^ 2 * x₂ + C.r ∧
      (C.u : R) ^ 3 * y₁ + (C.u : R) ^ 2 * C.s * x₁ + C.t = W.toAffine.negY
        ((C.u : R) ^ 2 * x₂ + C.r) ((C.u : R) ^ 3 * y₂ + (C.u : R) ^ 2 * C.s * x₂ + C.t))
      ↔ (x₁ = x₂ ∧ y₁ = (C • W).toAffine.negY x₂ y₂)
```

— the image pair satisfies the `negY` degeneracy condition, the one sending the group law down its
vertical-line branch, exactly when the original pair does. Both directions carry content: the
forward one transports a non-degenerate pair, the backward one a degenerate one.

The two call sites take the negated direction as `mt (variableChange_negY_iff W C).mp`, so the
contrapositive is derived where it is wanted rather than exported.

For the record on the earlier rounds: the wording up to round 8 was "a proof-specific degeneracy
helper remains unnecessarily public", which I read as a visibility problem and answered three
times by privatising things, then by measuring the public surface. The declaration was right all
along; what was wrong was its *shape*, not its visibility. Round 9's phrasing —
"contrapositive … instead of a characteristic transformation law" — got me there in one push.
Worth noting for whoever tunes the re-review summaries, as with the namespace finding on this same
PR earlier.

`lake build TauCeti.AlgebraicGeometry.EllipticCurve.Affine.Point`, which covers both files this PR
touches and everything they depend on, passes.

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Formula.lean
kim-em, line 1: ### 🟡 placement — request_changes  `codex/gpt-5.6-sol`

The earlier placement issues are resolved, but the new Point module retains one redundant import.

- `TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean:9` — The direct `VariableChange` import is unused because the preceding `Affine.Formula` import publicly re-exports it. _Fix:_ Remove `public import TauCeti.AlgebraicGeometry.EllipticCurve.VariableChange`.

<sub>`codex/gpt-5.6-sol` · 37s · 90.9k in / 942 out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/1828f281bda30b8ed5c87912cbed3b9a6df1d310...11751bf96b176bb364607f5872fc6dbe360e6c15) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/b8161ef29f1bf922d3f9aea2b1b1e298bcd06078/rubrics/placement.md)</sub>

kim-em: **Re: your reply on `api-design` —** re-reviewed on `4bba529`; this clears the finding ✅ — approved on `4bba529`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3742399945.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean
eohjelle: **Re: your reply on `reuse` —** re-reviewed on `11751bf`; the finding stands — The PR retains a public specialization that directly duplicates an existing Mathlib theorem.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3741824916.</sub>

CBirkbeck: **Re: `reuse` block on `11751bf9` —** I tried the prescribed fix before answering, in both the forms the finding offers. Neither compiles, and the failure is the one the docstring predicts. Details below; the branch is unchanged, so the block is contested rather than complied with.

**The finding is right about term positions in isolation.** This compiles:

```lean
example {R : Type*} [CommRing R] (W : WeierstrassCurve R) (D : VariableChange R) (A : Type*)
    [CommRing A] [Algebra R A] :
    (D.baseChange A) • (W.baseChange A) = (D • W).baseChange A :=
  map_variableChange (W := W) (C := D) (φ := algebraMap R A)
```

So the *proof* of `baseChange_variableChange` is indeed a defeq coercion. That is not what the declaration is for.

**What it is for is `rw`, and that is where the substitution fails.** In the same file:

```lean
example … : (D.baseChange A) • (W.baseChange A) = (D • W).baseChange A := by
  rw [map_variableChange]
```

```
Tactic `rewrite` failed: Did not find an occurrence of the pattern
  VariableChange.map ?C ?φ • map ?W ?φ
in the target expression
  D.baseChange A • W⁄A = (D • W)⁄A
```

`rw` matches up to reducible transparency and `WeierstrassCurve.baseChange` is a plain `def`, so the `⁄` spelling is never seen as `map … (algebraMap …)`.

**Attempt 1 — "supply this proof term directly".** Replacing all four uses in `map_mapVariableChange` with `(map_variableChange (W := W') (C := D) (φ := algebraMap R F)).symm` elaborates, but the proof then dies two lines later:

```
Point.lean:233:8: Tactic `rewrite` failed: Did not find an occurrence of the pattern
  (AddEquiv.cast ?h) (some ?x ?y ?hns)
in the target expression
  (map f) ((mapVariableChange (W'⁄F) (D.baseChange F)) ((AddEquiv.cast ⋯) (some x y h))) =
    (mapVariableChange (W'⁄K) (D.baseChange K)) ((AddEquiv.cast ⋯) ((map f) (some x y h)))
```

The `AddEquiv.cast` now carries a `map`-spelled motive, so `cast_some` no longer matches it.

**Attempt 2 — "`change` the goal before rewriting".** The proof already opens with a `change` (it bridges the plain-`cast` statement to the `AddEquiv.cast` spelling). Forcing the `baseChange` spelling through a type ascription instead,

```lean
((map_variableChange (W := W') (C := D) (φ := algebraMap R F) :
    (D.baseChange F) • (W'⁄F) = (D • W')⁄F)).symm
```

fails at the same `rw [cast_some, …]`, at `Point.lean:237:8`.

**Control.** Reverting to the committed state and rebuilding the same module in the same tree: `Build completed successfully (1990 jobs)`. So both failures are caused by removing the restatement, not by anything pre-existing.

**Why this is API rather than notation.** Every consumer here — and `Point.map`, whose type is already phrased with `⁄` — is in the `baseChange` vocabulary. Without the named restatement each call site has to re-ascribe the same equality inline, which is the duplication the rubric otherwise objects to, and Attempt 2 shows even that does not recover the proof.

If there is a spelling of the substitution that does compile, I will take it — I have tested the two the finding names and reported exactly what each produced.

Note for the record: `reuse` was not blocking at rounds 9 or 10. `11751bf9` moved this declaration from `Affine/Point.lean` to `EllipticCurve/VariableChange.lean` on a `placement` finding, which makes it appear in the diff as a fresh addition; the declaration and its proof are unchanged by that commit.

CBirkbeck: **Second contest on `reuse`, this time on a different ground: this finding and the `placement`
finding on this PR cannot both be satisfied.** Per `AGENTS.md` I am showing the contradiction
rather than satisfying one and letting the other re-fire.

**`placement` — request_changes, round 11, judged on `4bba529`** ([discussion_r3742660713](https://github.com/TauCetiProject/TauCeti/pull/2247#discussion_r3742660713)), second bullet, verbatim:

> `TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean:62` — `WeierstrassCurve.baseChange_variableChange` is a curve-level restatement of `map_variableChange`, not point-group material, so users must import the later affine Point module for general variable-change API. *Fix:* Move it beside the other variable-change complements in `TauCeti/AlgebraicGeometry/EllipticCurve/VariableChange.lean`.

**`reuse` — block, this thread**, verbatim:

> `TauCeti/AlgebraicGeometry/EllipticCurve/VariableChange.lean:164` — `baseChange_variableChange` is exactly Mathlib's `WeierstrassCurve.map_variableChange` specialized to `algebraMap`; notation/elaboration convenience does not create new mathematical API. *Fix:* Delete the lemma …

The two are the same declaration, and `VariableChange.lean:164` is the location `placement`
sent it to. I complied with `placement` at `11751bf9`; that commit is the entire reason this
declaration reappears in the diff as a fresh addition, and it is the commit this rubric now
blocks.

**Why both cannot hold.** `placement`'s fix is *move it to `VariableChange.lean`*, which
presupposes the declaration exists; `reuse`'s fix is *delete it*. There is no state satisfying
both. This is not a case of one rubric being silent about the other's concern: `placement`
independently characterises the lemma as "a curve-level restatement of `map_variableChange`" —
the very observation this thread rests on — and, having made it, still directs that the lemma be
**kept and relocated**, on the stated ground that users otherwise "must import the later affine
Point module for general variable-change API." `placement` is therefore asserting that this
declaration is general variable-change API that the module ought to expose.

**The same rubric has already cleared this declaration.** On this thread, `2026-08-09T00:08:45Z`:

> **Re: your reply on `reuse` —** re-reviewed on `f2e5e29`; this clears the finding ✅ — approved on `f2e5e29`.

The declaration and its proof are byte-identical between `f2e5e29` and `11751bf9`; the only
difference is the file it lives in, and that move was ordered by `placement`. A rubric that
approves a declaration in `Affine/Point.lean` and blocks the identical declaration in
`VariableChange.lean` is reacting to the relocation, not to the mathematics.

**The compiler evidence from my previous reply still stands and is unaddressed.** Both fixes this
thread names were tried verbatim and neither compiles — supplying the proof term directly fails at
`Point.lean:233:8` (`cast_some` no longer matches an `AddEquiv.cast` carrying a `map`-spelled
motive) and the `change`/ascription route fails at `Point.lean:237:8`; reverting rebuilds clean
(`1990 jobs`). `WeierstrassCurve.baseChange` and `VariableChange.baseChange` are plain `def`s, so
`W⁄A` is never seen as `W.map (algebraMap R A)` at `rw`/`simp only` transparency.

**What I am asking for.** Resolve the conflict in one direction and I will implement it in one
push:

1. If `reuse` prevails, `placement`'s second bullet must be withdrawn, and I need a spelling of
   the substitution that compiles — the two named do not.
2. If `placement` prevails, the lemma stays at `VariableChange.lean:164`, and I will add a
   docstring line naming `map_variableChange` as the general form, or rename it to
   `map_variableChange_algebraMap`, if either would settle the duplication objection.

I have not pushed anything for this round; the branch is unchanged at `11751bf9`.

CBirkbeck: **Contesting again** (posting a new reply, per this thread's instructions). Two pieces of
evidence: the suggested fix still does not compile *with this rubric's own escape hatch already
applied*, and this rubric approved the identical declaration six hours earlier.

### 1. The suggested fix does not compile — and `change` is already there

The fix says to delete the lemma, use `map_variableChange (W := W) (C := D) (φ := algebraMap R A)`
directly, and "where rewriting needs the `baseChange` presentation, supply this proof term
explicitly or `change` the goal before rewriting". I applied all of it at `11751bf`: deleted
`baseChange_variableChange` and substituted the proof term at all four occurrences in
`map_mapVariableChange`.

Note that `change` is **not** an untried remedy here — `map_mapVariableChange` already opens with
exactly that step (`Affine/Point.lean:223`), converting the statement's plain `cast` into the
`AddEquiv.cast` spelling so `cast_some` can apply. With the lemma inlined, that `change` still
elaborates and the proof still fails, in two places:

```
error: TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean:232:2: unsolved goals
case zero
...
error: TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean:233:8: Tactic `rewrite` failed: Did not find an occurrence of the pattern
  (AddEquiv.cast ?h) (some ?x ?y ?hns)
in the target expression
  (map f) ((mapVariableChange (W'⁄F) (D.baseChange F)) ((AddEquiv.cast ⋯) (some x y h))) =
    (mapVariableChange (W'⁄K) (D.baseChange K)) ((AddEquiv.cast ⋯) ((map f) (some x y h)))
```

The equation is used in the **statement** of `map_mapVariableChange`, inside
`congrArg (fun V ↦ V.toAffine.Point)`, not only in its proof — so a `have` inside the proof cannot
carry it either. `WeierstrassCurve.baseChange` and `VariableChange.baseChange` are plain `def`s,
so `W⁄A` and `W.map (algebraMap R A)` are definitionally equal but not syntactically
interchangeable at the transparency `rw`/`simp only` use; the transport therefore has to be
*stated* in the `baseChange` spelling.

### 2. This rubric approved this exact declaration at `f2e5e29`

Verbatim, from this thread (comment 3742039007, 2026-08-09T00:08:45Z):

> **Re: your reply on `reuse` —** re-reviewed on `f2e5e29`; this clears the finding ✅ — approved on `f2e5e29`.

and then, five hours later (comment 3742825392, 2026-08-09T05:58:29Z):

> **Re: your reply on `reuse` —** re-reviewed on `11751bf`; the finding stands — The PR retains a public specialization that directly duplicates an existing Mathlib theorem.

The declaration is the same in both. At `f2e5e29` it is `Affine/Point.lean:62`; at `11751bf` it is
`VariableChange.lean:166`. Statement and proof are character-for-character identical —

```lean
    (D.baseChange A) • (W.baseChange A) = (D • W).baseChange A :=
  map_variableChange (W := W) (C := D) (φ := algebraMap R A)
```

— the only differences being that the binders `{R : Type*} [CommRing R]` are now section
variables, and the docstring is longer. What changed between the two commits is the **module**:
`11751bf9` is titled *"refactor(EllipticCurve): move the coordinate-map laws to VariableChange"*.

That move was not optional. `placement` (round 11, on `4bba529`) required it, describing the
declaration in the same terms this rubric now uses while asking that it be *kept*:

> - `TauCeti/AlgebraicGeometry/EllipticCurve/Affine/Point.lean:62` — `WeierstrassCurve.baseChange_variableChange` is a curve-level restatement of `map_variableChange`, not point-group material, so users must import the later affine Point module for general variable-change API. _Fix:_ Move it beside the other variable-change complements in `TauCeti/AlgebraicGeometry/EllipticCurve/VariableChange.lean`.

So the same lemma is approved by `reuse` at `Affine/Point.lean` and blocked by `reuse` at
`VariableChange.lean`, the one path `placement` sent it to. I do not think a declaration's module
should change whether it counts as duplication; if it does here, the two findings cannot be
satisfied at once by moving it, and §1 shows it cannot be satisfied by deleting it either.

I have left the lemma in place at `VariableChange.lean:166`, where `placement` asked for it, with a
docstring recording why the restatement is load-bearing.

kim-em: **Re: your reply on `reuse` —** re-reviewed on `11751bf`; this clears the finding ✅ — approved on `11751bf`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3743121723.</sub>

Issue #2229, opened 2026-08-07
Review stuck: PR #1991 and #1992, no auto-filed tracking issue despite 7-10h stall


eric-distyl-ai:
Both PRs have been waiting on a fresh review round for a long time, with CI green throughout and
no `Review stuck: PR #N` issue auto-filed for either (unlike the precedent in #1980, #1929, #1928,
#1927, #1920, #1919, #1907, #1549) — possibly because the auto-filer shares the same stuck
component.

- **#1991**: last scoreboard at `2026-08-07T04:16:32Z` for commit `ad41d1c9`. Current head
  (`e10759c7`, pushed shortly after) has had no scoreboard since — over 7 hours as of this
  writing (`2026-08-07T11:44Z`). CI passed promptly for every commit in that window.
- **#1992**: last scoreboard at `2026-08-07T01:53:23Z` for commit `5ff0be5c`. Current head
  (`dc65f2ac`) has had several pushes and a manual `/review` retrigger since, with no scoreboard
  landing — over 10 hours. CI passed promptly for every commit in that window too.

Both PRs are otherwise healthy: CI green, no merge conflicts, no `sorry`/axiom issues locally.
This looks like the same review-pipeline stall class as the linked issues above, just without the
usual auto-filed companion issue — flagging in case that absence itself is a symptom (e.g. the
auto-filer depends on a review round actually completing to notice the PR is stuck, and a stall
this long/this wide silently starves it too).

Roadmap: none

eric-distyl-ai:
Correction/update: this is not a repo-wide outage. Checked several other PRs opened well within
the same stuck window (all times UTC):

- #2265 opened 21:47 (2026-08-07), reviewed 22:24 — ~37 min turnaround.
- #2270 opened 22:22, reviewed 23:54 — ~1.5h turnaround.
- #2280 opened 22:51, reviewed 00:52 (2026-08-08) — ~2h turnaround.

All three got a normal round-1 scoreboard while #1991 and #1992 have had none since
2026-08-07T04:16Z and 2026-08-07T01:53Z respectively (now ~21-24h). So the review pipeline is
generally healthy right now; something PR-specific is wedging just these two. Given the same
symptom (silent, no verdict at all, unlike a `request_changes` repeating unchanged) previously
resolved for this same underlying content by opening a fresh PR (#1956→#1991, #1932→#1992), about
to try that again for both.

eric-distyl-ai:
[...]

PR #1991, opened 2026-08-04
feat: generalize C^{1,1} boundedness to corner crossings (one-sided Lipschitz derivative)

eric-distyl-ai:
Roadmap: ContourIntegration

Replaces #1956, closed/reopened once already to rule out stuck round-tracking state — opening
fresh in case that state is tied to the PR number itself. Same content as #1956's last commit
(`32692494`); CI green, correctness/generality/placement/api-design all approved there. The two
remaining findings (`reuse`/`proof-quality`, both about the `_right`/`_left` theorems' duplicated
scaffolding, and a claimed "unused fact" in the pointwise helper) repeated verbatim across many
rounds despite three independent factoring attempts and a full audit of every `have` in the file
— see #1956's comments for the reproducible evidence (all three factoring attempts break the
*sandboxed* build specifically via a `linarith` internal error that doesn't reproduce locally;
every `have` was individually deleted and rebuilt to confirm load-bearing-ness). That duplication
is since resolved (see the commit history below): `_left` is now derived from `_right` by applying
it to the point-reflected curve, built on Mathlib's `derivWithin_comp_const_sub` and related
lemmas rather than re-deriving them.

## Summary

Generalizes `TauCeti.Contour.exists_isBounded_image_realWindingIntegrand_of_lipschitzOnWith_deriv`
(#1830, merged) to cover a crossing that coincides with a breakpoint of a piecewise-`C¹`
immersion — a **corner**, where the two one-sided tangents may disagree.

## Motivation

The merged theorem requires `deriv γ` to be Lipschitz on a genuine **two-sided** neighborhood of
the crossing. That hypothesis is never satisfiable when the crossing sits exactly at a corner:
`IsPwC1ImmersionOn` explicitly allows corners, where `deriv γ` (the ordinary, two-sided
derivative) is only junk-valued at the corner point itself. So the merged theorem silently
excludes exactly this case — a genuine gap, caught in review of the on-curve real-integral
formula PR this prerequisite feeds, which pointed out that this makes the consuming theorem
narrower than the roadmap's actual piecewise-`C^{1,1}` target.

Hungerbühler–Wasem's own proof of Prop 2.3 (arXiv:1808.00997, p. 9) handles this directly: *"If Λ
is only `C^{1,1}` on `U⁻` and `U⁺`, the claim follows in the same way by considering the
unilateral intervals left and right of the zero."* This PR formalizes exactly that one-sided
splitting.

## Change

[...]

PR #1956, opened 2026-08-04
feat: generalize C^{1,1} boundedness to corner crossings (one-sided Lipschitz derivative)

eric-distyl-ai:
Roadmap: ContourIntegration

## Summary

Generalizes `TauCeti.Contour.exists_isBounded_image_realWindingIntegrand_of_lipschitzOnWith_deriv`
(#1830, merged) to cover a crossing that coincides with a breakpoint of a piecewise-`C¹`
immersion — a **corner**, where the two one-sided tangents may disagree.

## Motivation

The merged theorem requires `deriv γ` to be Lipschitz on a genuine **two-sided** neighborhood of
the crossing. That hypothesis is never satisfiable when the crossing sits exactly at a corner:
`IsPwC1ImmersionOn` explicitly allows corners, where `deriv γ` (the ordinary, two-sided
derivative) is only junk-valued at the corner point itself. So the merged theorem silently
excludes exactly this case — a genuine gap, caught in review of #1932 (the on-curve real-integral
formula this prerequisite feeds), which pointed out that this makes the consuming theorem
narrower than the roadmap's actual piecewise-`C^{1,1}` target.

Hungerbühler–Wasem's own proof of Prop 2.3 (arXiv:1808.00997, p. 9) handles this directly: *"If Λ
is only `C^{1,1}` on `U⁻` and `U⁺`, the claim follows in the same way by considering the
unilateral intervals left and right of the zero."* This PR formalizes exactly that one-sided
splitting.

## Change

Replaces the single two-sided theorem with two one-sided ones:

* `exists_isBounded_image_realWindingIntegrand_of_lipschitzOnWith_derivWithin_right`
* `exists_isBounded_image_realWindingIntegrand_of_lipschitzOnWith_derivWithin_left`

Each requires only `ContDiffOn ℝ 1 γ` on a one-sided closed piece ending or starting at the
crossing, plus `derivWithin γ` Lipschitz and non-zero there — no assumption that the two sides
agree, and no second derivative (pointwise or a.e.) assumed anywhere, matching the merged
theorem's own proof technique (the quadratic remainder from the mean-value inequality, then a
pointwise numerator/denominator estimate). Both helper lemmas (previously stated for the ambient
`deriv`) are now parametrized by the one-sided piece's own `derivWithin`, so the identical
argument runs verbatim on either side; the smooth (non-corner) case is recovered by applying
both with the same underlying piece.

No `sorry`, no new axioms, no `set_option`. Full `lake build` verified clean locally.

## References

[...]

PR #1830, opened 2026-08-02
feat: bound the real winding integrand at C^{1,1} crossings

eric-distyl-ai:
Roadmap: ContourIntegration

## Summary

Adds `TauCeti.Contour.isBounded_image_realWindingIntegrand_of_lipschitzOnWith_deriv`: the real
winding integrand `realWindingIntegrand (γ t - w) (deriv γ t)` stays bounded on a small enough
window around a crossing `γ t₀ = w`, given only that `deriv γ` is Lipschitz-continuous on a
neighborhood of `t₀` (with `deriv γ t₀ ≠ 0`) — `C^{1,1}` regularity, not the `C²` regularity
`isBounded_image_realWindingIntegrand` (`BoundedIntegrand.lean`, #1787) needs.

### Why this is a separate contribution, not a strengthening of #1787

This is a genuinely different proof technique, not a weakening of the existing one. The `C²`
proof there reads the bounded limit off an *explicit curvature value* at the crossing, which
needs a second derivative — pointwise or almost everywhere — to actually exist at that point.
Under mere `C^{1,1}`, that need not hold exactly at the crossing itself, so the existing
technique doesn't generalize by simply weakening a hypothesis.

Instead, this file bounds the integrand directly from the **quadratic remainder** a Lipschitz
derivative forces on the curve — a standard mean-value estimate (`norm_sub_sub_mul_deriv_le_of_lipschitzOnWith`,
via Mathlib's `norm_image_sub_le_of_norm_deriv_le_segment'`) applied to the affine
approximation `γ(t) - γ(t₀) - (t - t₀) · deriv γ(t₀)`, with **no second derivative anywhere**.

Writing `z(t) := γ(t) - w` and `v(t) := deriv γ(t)` as an affine part plus a remainder
controlled by the Lipschitz constant `K`:

```
z(t) = (t - t₀) · v₀ + R(t),   ‖R(t)‖ ≤ K(t - t₀)²        (v₀ := deriv γ t₀)
v(t) = v₀ + e(t),              ‖e(t)‖ ≤ K|t - t₀|
```

both the winding integrand's numerator `z.re·v.im - z.im·v.re` and a lower bound on its
denominator `‖z‖²` scale as `(t - t₀)²`, so the ratio stays bounded as `t → t₀`
(`abs_realWindingIntegrand_le_of_lipschitzOnWith`).

### Relationship to the roadmap

`BoundedIntegrand.lean`'s own docstring already flags this as the natural next step: "The full
[...]

PR #2357, opened 2026-08-08
feat: generalize C^{1,1} boundedness to corner crossings (one-sided Lipschitz derivative)

eric-distyl-ai:
Roadmap: ContourIntegration

Replaces #1991. Same content, opened fresh because #1991's review pipeline stopped producing any
verdict at all for its current commit (`e10759c7`) for 21+ hours, while other PRs opened and
reviewed in that same window turned around normally (see
[the tracking issue](https://github.com/TauCetiProject/TauCeti/issues/2229) for the timing
evidence) — the same silent-pipeline symptom that #1956 hit before it was replaced by #1991 itself.

## Summary

Generalizes `TauCeti.Contour.exists_isBounded_image_realWindingIntegrand_of_lipschitzOnWith_deriv`
(#1830, merged) to cover a crossing that coincides with a breakpoint of a piecewise-`C¹`
immersion — a **corner**, where the two one-sided tangents may disagree.

## Motivation

The merged theorem requires `deriv γ` to be Lipschitz on a genuine **two-sided** neighborhood of
the crossing. That hypothesis is never satisfiable when the crossing sits exactly at a corner:
`IsPwC1ImmersionOn` explicitly allows corners, where `deriv γ` (the ordinary, two-sided
derivative) is only junk-valued at the corner point itself. So the merged theorem silently
excludes exactly this case — a genuine gap, caught in review of the on-curve real-integral
formula PR this prerequisite feeds, which pointed out that this makes the consuming theorem
narrower than the roadmap's actual piecewise-`C^{1,1}` target.

Hungerbühler–Wasem's own proof of Prop 2.3 (arXiv:1808.00997, p. 9) handles this directly: *"If Λ
is only `C^{1,1}` on `U⁻` and `U⁺`, the claim follows in the same way by considering the
unilateral intervals left and right of the zero."* This PR formalizes exactly that one-sided
splitting.

## Change

Replaces the single two-sided theorem with two one-sided ones, plus their two-sided combination:

* `exists_isBounded_image_realWindingIntegrand_of_lipschitzOnWith_derivWithin_right`
* `exists_isBounded_image_realWindingIntegrand_of_lipschitzOnWith_derivWithin_left`
* `exists_isBounded_image_realWindingIntegrand_of_lipschitzOnWith_derivWithin_corner`

Each one-sided theorem requires only `DifferentiableOn ℝ γ` on a one-sided closed piece ending or
starting at the crossing, plus `derivWithin γ` Lipschitz and non-zero there — no assumption that
the two sides agree, and no second derivative (pointwise or a.e.) assumed anywhere. `_corner`
combines both into one full-neighborhood bound, allowing the two sides to have independent
Lipschitz constants, for callers that need a two-sided window without assuming the two one-sided
pieces agree. There is no wrapper theorem restoring the pre-corner-case two-sided/ambient-`deriv`
API: the current `AGENTS.md`/`api-design` rubric forbids retaining compatibility surfaces whose
only purpose is to keep an obsolete API working, and the deleted theorem has no in-repository
callers.

`_left` is derived from `_right` by applying it to the point-reflected curve
`s ↦ γ (2 * t₀ - s)`, built on Mathlib's `derivWithin_comp_const_sub` (and related lemmas) rather
than re-deriving them or duplicating `_right`'s proof.

No `sorry`, no new axioms, no `set_option`. Full `lake build` verified clean locally.

## References

* N. Hungerbühler, M. Wasem, *Non-integer valued winding numbers and a generalized Residue
  Theorem*, arXiv:1808.00997 (2018), Proposition 2.3 and its proof.

### DIFF
--- a/TauCeti/Analysis/Contour/Winding/Integrand.lean
+++ b/TauCeti/Analysis/Contour/Winding/Integrand.lean
@@ -53,7 +53,7 @@ theorem realWindingIntegrand_mul_mul {c : ℂ} (hc : c ≠ 0) (z v : ℂ) :
 
 /-- Negating the velocity while keeping the position fixed negates the real winding integrand:
 it is the imaginary part of `z⁻¹ * (-v) = -(z⁻¹ * v)`. -/
-@[grind] theorem realWindingIntegrand_neg_right (z v : ℂ) :
+theorem realWindingIntegrand_neg_right (z v : ℂ) :
     realWindingIntegrand z (-v) = -realWindingIntegrand z v := by
   simp only [realWindingIntegrand_eq_div, Complex.neg_im, Complex.neg_re]; ring
 
--- a/TauCeti/Analysis/Contour/Winding/Integrand.lean
+++ b/TauCeti/Analysis/Contour/Winding/Integrand.lean
@@ -53,7 +53,7 @@ theorem realWindingIntegrand_mul_mul {c : ℂ} (hc : c ≠ 0) (z v : ℂ) :
 
 /-- Negating the velocity while keeping the position fixed negates the real winding integrand:
 it is the imaginary part of `z⁻¹ * (-v) = -(z⁻¹ * v)`. -/
-@[simp] theorem realWindingIntegrand_neg_right (z v : ℂ) :
+@[grind] theorem realWindingIntegrand_neg_right (z v : ℂ) :
     realWindingIntegrand z (-v) = -realWindingIntegrand z v := by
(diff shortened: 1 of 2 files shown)


eric-distyl-ai:
Superseded by a fresh PR: no review scoreboard after 27+ minutes since CI passed, while #2359 (opened later, same window) already got reviewed — the same PR-specific stall pattern seen before. Same content, pushed to `claude/lipschitz-bounded-corner-v7`. Replacement PR linked below.

PR #2361, opened 2026-08-08
feat: the order of a power series substitution is the product of the orders

CBirkbeck:
Roadmap: EllipticCurves

Adds `TauCeti/RingTheory/PowerSeries/OrderSubst.lean`:

```lean
theorem PowerSeries.order_subst [CommRing R] [NoZeroDivisors R] {f g : R⟦X⟧}
    (hf : constantCoeff f = 0) :
    PowerSeries.order (PowerSeries.subst f g : R⟦X⟧) =
      PowerSeries.order g * PowerSeries.order f
```

The order of a substitution is the product of the orders.

**Pure power-series algebra — no elliptic-curve content, and nothing from this repository is
imported.**

## Mathlib check

Mathlib has only the **inequality**. `grep -E "order_subst"` over
`Mathlib/RingTheory/PowerSeries/` returns `le_order_subst` (`Substitution.lean:344`) together with
`le_order_subst_left` and `le_order_subst_right` — all of the form
`order g * order f ≤ order (subst f g)`. There is no equality version, so this is a genuine
strengthening rather than a restatement.

The hypothesis `constantCoeff f = 0` is what makes `subst f g` defined at all (it supplies
`HasSubst f`); nothing is assumed of `g`, and `[NoZeroDivisors R]` is what rules out cancellation
of leading terms.

## Provenance

Ported from the AINTLIB `HasseWeil` project (Apache-2.0), revision `513e83879e2f`, file
`HasseWeil/FormalGroup/OrderSubst.lean`, declarations `constantCoeff_subst_univariate` and
`order_subst`, where it gives additivity of the height of a composition of formal-group
homomorphisms (Silverman IV.7).

Shortened relative to the source: three of its branches each re-derived `order φ = 0` from
`constantCoeff φ ≠ 0` by an explicit `le_antisymm` plus `order_le`, which is just Mathlib's
`order_ne_zero_iff_constCoeff_eq_zero` contraposed — that is now the private
`order_eq_zero_of_constantCoeff_ne_zero`, used three times.

One thing worth recording for the next person: the statement has to be written with explicit
`PowerSeries.order (…)` rather than dot notation. `subst f g` lives in `MvPowerSeries Unit R`, so
`(subst f g : R⟦X⟧).order` resolves to `MvPowerSeries.order` while the supporting lemmas produce
`PowerSeries.order`, and the two do not match syntactically even though the types are defeq.

## Roadmap

`EllipticCurves`, Layer 3 — the Hasse strand, whose §Ordering entry reads *"the Hasse bound is the
earliest PR, its existing proof self-contained; the zeta strand needs Layer 1's Frobenius
identities."*

A prerequisite rather than a target, which `CLAUDE.md` admits explicitly: a declaration may be
added when it *"advances a specific roadmap target, or supplies a prerequisite that a specific
target needs"*. In the roadmap's pinned Hasse provenance (`dev/hasse-weil @ 513e83879e2f`),
`HasseWeil/FormalGroup/OrderSubst.lean` sits in the development supporting the capstone
`hasse_bound` proof.

Nothing here waits on absent material: the statement is usable today by any caller substituting a
power series with vanishing constant term.

## Gates

locally. Longest proof is 24 lines; no `sorry`, no `maxHeartbeats`.

### DIFF
--- a/TauCeti/RingTheory/PowerSeries/OrderSubst.lean
+++ b/TauCeti/RingTheory/PowerSeries/OrderSubst.lean
@@ -1,0 +1,103 @@
+/-
+Copyright (c) 2026 The Tau Ceti contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
+module
+
+public import Mathlib.RingTheory.PowerSeries.Order
+public import Mathlib.RingTheory.PowerSeries.Substitution
+
+/-!
+# The order of a power series substitution
+
+Mathlib's `PowerSeries.le_order_subst` bounds the order of a substitution from below,
+`order g * order f ≤ order (subst f g)`. Over a ring without zero divisors that bound is an
+equality:
+
+`order (subst f g) = order g * order f`.
+
+The hypothesis `constantCoeff f = 0` is what makes `subst f g` defined at all (`HasSubst f`), and
+nothing is assumed of `g`.
+
+## Main results
+
+* `PowerSeries.order_subst`: the equality above.
+
+## Provenance
+
+Ported from the AINTLIB `HasseWeil` project (Apache-2.0), revision `513e83879e2f`, file
+`HasseWeil/FormalGroup/OrderSubst.lean`, declarations `constantCoeff_subst_univariate` and
+`order_subst`, where it gives additivity of the height of a composition of formal-group
+homomorphisms (Silverman IV.7).
+
+The proof here is shorter than the source's: three of its case branches each re-derived
+`order φ = 0` from `constantCoeff φ ≠ 0` by `le_antisymm`, which is Mathlib's
+`order_ne_zero_iff_constCoeff_eq_zero` contraposed, and the subsingleton branch is handled by
+`Subsingleton.elim` rather than by rewriting each series to `0` separately.
+-/
+
+public section
+
+namespace PowerSeries
+
+variable {R : Type*} [CommRing R]
+
+/-- A power series with nonzero constant coefficient has order zero. -/
+private theorem order_eq_zero_of_constantCoeff_ne_zero {φ : R⟦X⟧} (h : constantCoeff φ ≠ 0) :
+    φ.order = 0 := by
+  by_contra hne
+  exact h (order_ne_zero_iff_constCoeff_eq_zero.mp hne)
+
+/-- Substituting a series with zero constant coefficient does not change the constant
+coefficient. -/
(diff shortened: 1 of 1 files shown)


CBirkbeck:
Withdrawing. The `scope` block is right, and the provenance claim in my description was false — I checked before answering rather than defending it.

I computed the import closure of the capstone `HasseWeil/WeilPairing/HasseBound.lean` in the pinned Hasse provenance and asked whether this file is in it:

```
capstone closure size: 187   (project has 283 files)
  OUT  HasseWeil.FormalGroup.OrderSubst
  IN   HasseWeil.HahnSeriesAux
  IN   HasseWeil.PowerSeriesHelpers
  IN   HasseWeil.WeilPairing.PairingDet
  IN   HasseWeil.WeilPairing.Discriminant
```

So `order_subst` is **not** in the Hasse closure. It supports the formal-group strand (Silverman IV.7, additivity of the height of a composition of formal-group homomorphisms), which is Layer 1, not the Layer 3 Hasse strand my description cited. The rubric's reading of the Hasse section — `deg(1 − π_q)`, the degree form, finite-level computations — is the accurate one.

**How I got it wrong, since it is a process failure rather than a judgement call.** I had been justifying these ports by capstone-closure membership, checked each time. This tick I wrote a script to recompute which files are *portable* (sorry-free, all local dependencies already handled) and picked from its output — but that script scans the whole 283-file project, not the 187-file closure. I then carried the closure claim over to a file the script had surfaced from outside it. The script now intersects with the closure so the two criteria cannot drift apart again.

The two `reuse` findings on the same run were also fair and I am not disputing them: `order_eq_zero_of_constantCoeff_ne_zero` is the contrapositive of Mathlib's `order_ne_zero_iff_constCoeff_eq_zero`, and `constantCoeff_subst_eq` should start from `constantCoeff_subst` rather than re-running `coeff_subst'`. If this lemma is wanted later under a formal-group target, it should be rewritten that way first.

Closing rather than leaving a blocked PR consuming review budget. The mathematics is sound and Mathlib genuinely lacks the equality (it has only `le_order_subst`), so this is worth revisiting if and when a Layer 1 formal-group target exists to consume it.

CBirkbeck:
Re-closing: the `scope` block stands — verified this file is OUTSIDE the 187-file `hasse_bound` import closure (it is formal-group/Layer-1 material), so the roadmap justification was false. Revisit only under a formal-group target, with the two reuse findings fixed first.

PR #2368, opened 2026-08-08
feat(EllipticCurve): the coordinate ring is faithful, and points are determined by coordinates

CBirkbeck:
Roadmap: EllipticCurves

Adds `TauCeti/AlgebraicGeometry/EllipticCurve/Affine/CoordinateRing.lean`: three injectivity facts
about Mathlib's affine API for a Weierstrass curve.

```lean
theorem WeierstrassCurve.Affine.CoordinateRing.algebraMap_poly_injective :
    Function.Injective (algebraMap R[X] W'.CoordinateRing)

theorem WeierstrassCurve.Affine.CoordinateRing.algebraMap_injective' :
    Function.Injective (algebraMap R W'.CoordinateRing)

theorem WeierstrassCurve.Affine.Point.some_eq_some_iff (h₁ : W'.Nonsingular x₁ y₁)
    (h₂ : W'.Nonsingular x₂ y₂) : some x₁ y₁ h₁ = some x₂ y₂ h₂ ↔ x₁ = x₂ ∧ y₁ = y₂
```

The coordinate ring `R[X][Y] ⧸ ⟨polynomial⟩` contains `R[X]` faithfully — quotienting by the
Weierstrass polynomial kills nothing in `X` alone, since `1` belongs to the free `R[X]`-basis
`{1, Y}` — and hence contains `R` too. Separately, `Affine.Point.some` is injective in its
coordinates.

## Mathlib check

All three are absent from the pin, by full-name untruncated searches:
`grep -F "algebraMap_poly_injective"` → 0 hits, `grep -F "some_eq_some_iff"` → 0 hits, and a
pattern search for any `Injective … algebraMap … CoordinateRing` statement → 0 hits. The
ingredients used (`smul_basis_eq_zero`, `C_injective`, `injective_iff_map_eq_zero`) are all
present.

## Provenance

Ported from the AINTLIB `HasseWeil` project (Apache-2.0), revision `513e83879e2f`, file
`HasseWeil/Auxiliary/Universal.lean`, declarations `algebraMap_poly_injective`,
`algebraMap_injective'` and `some_eq_some_iff` (© Junyan Xu).

**Only these three are taken.** The remainder of that file constructs the universal elliptic curve
over `ℤ[A₁,…,A₆]` together with its coordinate ring, function field and specialisation maps — a
large mathlib-track development rather than a leaf, and not something to bring in piecemeal, so it
is left alone.

One improvement over the source: it guards the first proof with
`set_option backward.isDefEq.respectTransparency false`, which this repository forbids. The proof
goes through without it here, so no override is needed.

## Roadmap

`EllipticCurves`, Layer 3 — the Hasse strand, whose §Ordering entry reads *"the Hasse bound is the
earliest PR, its existing proof self-contained; the zeta strand needs Layer 1's Frobenius
identities."*

A prerequisite rather than a target, which `CLAUDE.md` admits explicitly: a declaration may be
added when it *"advances a specific roadmap target, or supplies a prerequisite that a specific
target needs"*. Checked rather than assumed — in the roadmap's pinned Hasse provenance
(`dev/hasse-weil @ 513e83879e2f`), `HasseWeil/Auxiliary/Universal.lean` lies inside the **187-file
import closure of the capstone `hasse_bound` proof** (the project has 283 files, so closure
membership is not automatic).

No curve point count, torsion hypothesis or ellipticity assumption appears in any statement, and
nothing here waits on absent material.

## Gates

locally. All three proofs are term-mode; no `sorry`, no `set_option`, no `maxHeartbeats`.

### DIFF
--- a/TauCeti/AlgebraicGeometry/EllipticCurve/Affine/CoordinateRing.lean
+++ b/TauCeti/AlgebraicGeometry/EllipticCurve/Affine/CoordinateRing.lean
@@ -1,0 +1,74 @@
+/-
+Copyright (c) 2026 The Tau Ceti contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
+module
+
+public import Mathlib.AlgebraicGeometry.EllipticCurve.Affine.Point
+
+/-!
+# Faithfulness of the coordinate ring, and equality of affine points
+
+Two injectivity facts about Mathlib's affine API for a Weierstrass curve, both absent from the
+pinned Mathlib.
+
+The coordinate ring `R[X][Y] ⧸ ⟨polynomial⟩` contains `R[X]` faithfully: quotienting by the
+Weierstrass polynomial kills nothing in `X` alone, because the free `R[X]`-basis `{1, Y}` of the
+coordinate ring has `1` as one of its members. Consequently the base ring `R` embeds too.
+
+Separately, `Affine.Point.some` is injective in its coordinates, so two nonsingular points are
+equal exactly when their coordinates agree.
+
+## Main results
+
+* `WeierstrassCurve.Affine.CoordinateRing.algebraMap_poly_injective`
+* `WeierstrassCurve.Affine.CoordinateRing.algebraMap_injective'`
+* `WeierstrassCurve.Affine.Point.some_eq_some_iff`
+
+## Provenance
+
+Ported from the AINTLIB `HasseWeil` project (Apache-2.0), revision `513e83879e2f`, file
+`HasseWeil/Auxiliary/Universal.lean`, declarations `algebraMap_poly_injective`,
+`algebraMap_injective'` and `some_eq_some_iff` (© Junyan Xu). Only these three are taken: the rest
+of that file constructs the universal elliptic curve over `ℤ[A₁,…,A₆]`, which is a large
+mathlib-track development rather than a leaf, and is left alone.
+
+The source guards the first proof with `set_option backward.isDefEq.respectTransparency false`,
+which this repository forbids; it is not needed here.
+-/
+
+public section
+
+open Polynomial
+
+namespace WeierstrassCurve.Affine
+
+variable {R : Type*} [CommRing R] {W' : WeierstrassCurve.Affine R}
+
+namespace CoordinateRing
+
+/-- The coordinate ring contains `R[X]` faithfully.
+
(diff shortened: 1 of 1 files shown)


CBirkbeck:
Withdrawing — the block is correct on all three counts, and I verified each prescribed replacement before accepting rather than complying blind. All three compile:

```lean
example : Function.Injective (algebraMap R[X] W'.CoordinateRing) :=
  FaithfulSMul.algebraMap_injective R[X] W'.CoordinateRing

example : Function.Injective (algebraMap R W'.CoordinateRing) :=
  FaithfulSMul.algebraMap_injective R W'.CoordinateRing

example (h₁ : W'.Nonsingular x₁ y₁) (h₂ : W'.Nonsingular x₂ y₂) :
    some x₁ y₁ h₁ = some x₂ y₂ h₂ ↔ x₁ = x₂ ∧ y₁ = y₂ := by
  simp only [some.injEq]
```

(The third needs only `simp`, not even `simpa` — the linter says so.) So the file was three duplicates, and there is nothing to salvage.

**Why my Mathlib check missed them, since the check itself was run.** I searched for the AINTLIB names (`algebraMap_poly_injective`, `some_eq_some_iff`, both 0 hits) and for the statement shape (`Injective … algebraMap … CoordinateRing`, 0 hits). Both searches were untruncated and both were honest — but Mathlib does not supply these facts as named lemmas of that shape. It supplies them:

* **generically**, via `FaithfulSMul.algebraMap_injective` plus a `FaithfulSMul` instance for the coordinate ring — the statement I wanted never appears as text anywhere; and
* **automatically**, via the constructor-generated `Point.some.injEq`, which is not in any source file at all.

Neither is reachable by grepping for a statement. The AINTLIB source predates the `FaithfulSMul` instance, which is why it proved them by hand.

The check I was missing, and will now run for any injectivity/equality-shaped lemma: try `exact?` / the generic typeclass route (`FaithfulSMul`, `injEq`, `ext` lemmas) *before* writing, not just a name and statement grep. A one-line `example` would have killed this file in under a minute.

Closing rather than leaving a triple-duplicate open.

CBirkbeck:
Re-closing: the `reuse` block is correct on all three declarations — `FaithfulSMul.algebraMap_injective` (twice) and `Point.some.injEq` already give them, verified by compiling each replacement. The whole file was duplicates.

PR #2393, opened 2026-08-08
refactor(Conjugation): name the remaining restrictions in the coassociativity proof

CBirkbeck:
`conjugationAlgHom_coassoc` compares two coordinate maps `H ⊗[R] H →ₐ[R] H ⊗[R] (H ⊗[R] H)` by
computing what each does to the two tensor factors — four restrictions in all. Exactly one of
them, `toConv_assoc_comp_map_comul_comp_includeLeft`, was already a named private lemma; the
other three sat inline. This gives those three the same treatment.

| lemma | restriction of | says |
|---|---|---|
| `toConv_assoc_comp_map_comul_comp_includeRight` | `assoc ∘ (Δ ⊗ id)` to the second factor | the acted-on variable becomes the third universal point |
| `toConv_map_id_conjugationAlgHom_comp_includeLeft` | `id ⊗ c♯` to the first factor | the first universal point is untouched |
| `toConv_map_id_conjugationAlgHom_comp_includeRight` | `id ⊗ c♯` to the second factor | the third universal point gets conjugated by the second |

All three are stated **without** the parent's `let`-bound `assoc`, `mulFirst`, `actSecond`,
`g₁`, `g₂`, `x`, spelling the inclusions out instead — matching the existing sibling exactly, and
letting the parent keep citing them through its own `let`s.

What is left in the parent is the setup plus the group-theoretic comparison
`(g₁g₂) x (g₁g₂)⁻¹ = g₁ (g₂ x g₂⁻¹) g₁⁻¹`, which is the actual content of the associativity law.

Proof body: 47 → 41 lines; the three new bodies are 4, 2 and 4. No statement changes, and all
three helpers are `private`.

Roadmap: ReductiveGroups

### DIFF
--- a/TauCeti/Algebra/AlgebraicGroup/Hopf/Conjugation.lean
+++ b/TauCeti/Algebra/AlgebraicGroup/Hopf/Conjugation.lean
@@ -220,6 +220,52 @@ private theorem toConv_assoc_comp_map_comul_comp_includeLeft :
     ext h
     simp [up]
 
+/-- **The acted-on variable is untouched by comultiplying the conjugating variable.**
+Re-associating after comultiplying the first tensor factor sends the second universal point of
+`H ⊗[R] H` to the third universal point of `H ⊗[R] (H ⊗[R] H)`. -/
+private theorem toConv_assoc_comp_map_comul_comp_includeRight :
+    toConv
+        ((((Algebra.TensorProduct.assoc R R R H H H).toAlgHom.comp
+            (Algebra.TensorProduct.map (Bialgebra.comulAlgHom R H) (AlgHom.id R H))).comp
+          Algebra.TensorProduct.includeRight : H →ₐ[R] H ⊗[R] (H ⊗[R] H))) =
+      toConv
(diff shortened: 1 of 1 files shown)


CBirkbeck:
Closing on my own dry run: the `reuse` rubric rejects the premise, and I agree with it.

> The new private helpers at lines 226, 240, and 251 duplicate existing restriction API instead of
> reusing it directly: `Algebra.TensorProduct.map_comp_includeRight`,
> `Algebra.TensorProduct.map_comp_includeLeft`, and the existing `comp_conjugationAlgHom`.

Its fix is to prove the three `have`s inline from the located lemmas — which is what they already
were. Symmetry with the one restriction that *is* a named lemma
(`toConv_assoc_comp_map_comul_comp_includeLeft`) was the whole argument for this PR, and that is an
aesthetic argument, not a reuse one. There is nothing left to extract once the finding is honoured,
so there is no version of this PR worth re-reviewing.

The other nine rubrics were green; no defect is being left behind.

PR #2400, opened 2026-08-08
refactor(Semigroups): name the two steps that make the orbit average invertible

CBirkbeck:
`domain_eq_top_of_continuousAt_zero` is the nontrivial direction of the bounded-generator
characterisation, and its first two thirds say nothing about the generator domain at all: they
build the normalized local orbit average `t⁻¹ ∫₀ᵗ S(s) ds` and show it is invertible. Only the
last third — surjectivity plus the local-orbit integral identity — is about the domain.

Two private helpers now carry the first part:

| helper | says |
|---|---|
| `exists_pos_forall_norm_realOperator_sub_one_le` | a norm-continuous semigroup stays within `c` of the identity on some initial interval `(0, t]` |
| `norm_normalizedIntegral_sub_one_le` | the normalized average over `(0, t]` inherits that same bound `c` |

Both are stated for an **arbitrary** tolerance `c`, not the `1/2` the caller happens to pick, and
neither carries hypotheses it does not use: the first takes only
`ContinuousAt (fun t : ℝ ↦ S.realOperator t) 0` and omits `[CompleteSpace X]`; the second takes no
positivity hypothesis on `c` at all, only `0 < t` and interval integrability.

The parent's `let t` and `let B` are gone — `t` now arrives from the first helper's existential,
and the average is spelled out, so nothing is folded away from the helper statements.

Proof body: 50 → 22 lines; the two new bodies are 9 and 11.

Roadmap: OneParameterSemigroups

---

**On `[CompleteSpace X]`.** The interval helper omits it; the averaging helper cannot. Adding
`omit [CompleteSpace X] in` to `norm_normalizedIntegral_sub_one_le` fails:

```
error: failed to synthesize instance of type class
  CompleteSpace X
```

at the `S.normalizedIntegral_sub_one_eq` call — `intervalIntegral.integral_sub` needs the target
complete, and the same applies to `normalizedIntegral_sub_one_eq` itself. Lean's
`unusedSectionVars` linter agrees: it flags the interval helper and not this one.

PR #2402, opened 2026-08-08
refactor(L2): name the box exhaustion and the box step of the vanishing argument

CBirkbeck:
`setIntegral_eq_zero_of_forall_inner` carried three separate arguments in one 43-line body: the
Dynkin step inside a single spanning box, the fact that the boxes exhaust `α × β`, and the
monotone limit. Only the last is specific to the theorem.

| helper | says |
|---|---|
| `iUnion_prod_spanningSets` | `⋃ n, spanningSets μ n ×ˢ spanningSets ν n = univ` |
| `setIntegral_inter_prod_spanningSets_eq_zero` | the integral of `h` over `u ∩ (n-th box)` vanishes |

Neither is available in Mathlib: `iUnion_spanningSets` covers a single σ-finite measure, and
nothing states the product version (the `max i j` step is what makes the two sequences line up).

Hypotheses were re-derived at the point of extraction rather than inherited. In particular the box
lemma does **not** take the parent's finiteness hypothesis `hfin : (μ.prod ν) u < ⊤` — inside a box
the finiteness it needs is the box's own — and `iUnion_prod_spanningSets` mentions neither `𝕜` nor
`h`.

Proof body: 43 → 13 lines; the two new bodies are 8 and 19. The parent is now just: the sets
increase, they cover `u`, so the integrals converge to the integral over `u`, and every one of them
is zero.

Roadmap: OrthogonalL2Bases

### DIFF
--- a/TauCeti/Analysis/InnerProductSpace/L2/Product.lean
+++ b/TauCeti/Analysis/InnerProductSpace/L2/Product.lean
@@ -272,6 +272,44 @@ theorem setIntegral_prod_eq_zero_of_forall_inner [SFinite μ] [SFinite ν] {h : 
     _ = inner 𝕜 (L2prodMul F G) h := (L2.inner_def _ _).symm
     _ = 0 := hz F G
 
+/-- **The boxes built from two spanning sequences exhaust the product space.** -/
+private theorem iUnion_prod_spanningSets [SigmaFinite μ] [SigmaFinite ν] :
+    ⋃ n, (spanningSets μ n ×ˢ spanningSets ν n) = (Set.univ : Set (α × β)) := by
+  refine Set.eq_univ_of_forall fun p => ?_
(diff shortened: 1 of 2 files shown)


CBirkbeck:
Closing: this branch was cut from `decompose/semigroup-orbit-average` instead of `main`, so the
diff carried that PR's commit as well. My own dry run caught it — `scope` blocked with

> The PR bundles two unrelated refactors: L2 product box-exhaustion helpers and uniformly
> continuous semigroup averaging helpers.

and the `generality`/`proof-quality` findings in that run were against the semigroup file, which is
#2400's business.

Replaced by #2406, cut from `main` with the L2 change only. That PR also drops
`iUnion_prod_spanningSets`: the `reuse` finding here was right that
`Set.iUnion_prod_of_monotone` already covers it.

PR #2526, opened 2026-08-09
refactor(HomologyCauchy): name the per-generator Dixon identity

CBirkbeck:
`cycleDixonH1_eq_cycleDixonH2_sub` opened with a fourteen-line `hterm` establishing the Dixon identity for a single generator of the cycle, then spent the rest of the proof summing that identity over the support. The two halves are different kinds of work, and only the first is about Dixon functions.

`dixonH1_eq_dixonH2_sub_windingNumber_mul_f_of_mem_support` is that first half: for a generator `γ` of a cycle lying in `U` and a point `w` off the cycle's trace, `dixonH1` and `dixonH2` for `γ` differ by `2πi` times the winding number of `γ` about `w` times `f w`.

It joins `avoids_of_mem_support` and `mapsTo_of_mem_support` — the two facts about a support member that it is built from — and its name echoes the per-curve `Contour.dixonH1_eq_dixonH2_sub_windingNumber_mul_f` it specialises.

Its hypotheses are exactly the four it uses: continuity of `f` on `U`, the cycle lying in `U`, the point being off the trace, and membership in the support.

The parent keeps only the summation — `Finset.sum_congr`, `Finset.sum_sub_distrib` and the `mul_sum`/`sum_mul` rearrangement — and drops from 42 lines to 30. I left that algebra inline: it is a single `calc` over one `Finset`, and factoring it out would mean a general lemma about `∑ c • (A - K * B)` that asserts nothing the located `Finset` lemmas do not.

Gates run locally: `lake build` (7139 jobs), `lake exe axioms`, `lake exe module-system`, `scripts/lint-env.sh` — all green.

Roadmap: ContourIntegration

### DIFF
--- a/TauCeti/Analysis/Contour/Cycle/HomologyCauchy.lean
+++ b/TauCeti/Analysis/Contour/Cycle/HomologyCauchy.lean
@@ -95,6 +95,24 @@ private theorem mapsTo_of_mem_support {C : Cycle} {U : Set ℂ} (hCU : IsIn C U)
   intro t ht
   exact isIn_iff.mp hCU (mem_trace_iff.mpr ⟨γ, hγ, t, ht, rfl⟩)
 
+/-- **The Dixon identity for one generator of a cycle.** For a generator `γ` of a cycle lying in
+`U` and a point `w` off the cycle's trace, `dixonH1` and `dixonH2` for `γ` differ by `2πi` times
+the winding number of `γ` about `w` times `f w`. -/
+private theorem dixonH1_eq_dixonH2_sub_windingNumber_mul_f_of_mem_support {f : ℂ → ℂ} {C : Cycle}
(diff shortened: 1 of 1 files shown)


CBirkbeck:
Closing: the dry run's `reuse` finding is right, and it invalidates the premise of the PR rather than asking for a fix.

`dixonH1_eq_dixonH2_sub_windingNumber_mul_f_of_mem_support` asserts nothing that `Contour.dixonH1_eq_dixonH2_sub_windingNumber_mul_f` does not — its entire proof is `avoids_of_mem_support` and `mapsTo_of_mem_support` plus two integrability lemmas, filling that theorem's hypotheses. It is a one-use wrapper, and removing it leaves nothing in the PR.

Recording the obstacle for the queue: `cycleDixonH1_eq_cycleDixonH2_sub` (42 lines) does not decompose usefully. Its two halves are hypothesis plumbing for an existing per-curve lemma, and a `Finset` rearrangement of `∑ c • (A - K * B)`; neither is a statement worth naming, and the second would be a composition of located `Finset` lemmas.

PR #2548, opened 2026-08-09
refactor(IdealCount): name the rational prime below a prime ideal

CBirkbeck:
`absNormUnder P` is the norm of the rational prime below a prime ideal `P` of `𝓞 F`. That this number is *prime* was re-proved inline four times in `IdealCount/Basic.lean`, each time behind the same instance dance:

```lean
have := hPp
have : NeZero P := ⟨hP0⟩
… (by simpa [absNormUnder] using Nat.absNorm_under_prime P) …
```

`Nat.absNorm_under_prime` wants `P.IsPrime` and `NeZero P` as *instances*, so every call site had to re-manufacture them from the propositional hypotheses it had in hand.

`prime_absNormUnder` states the fact once, taking those as ordinary hypotheses:

```lean
private theorem prime_absNormUnder {P : Ideal (𝓞 F)} (hP : P.IsPrime) (hP0 : P ≠ ⊥) :
    (absNormUnder F P).Prime
```

Both hypotheses are load-bearing — `hP0` because `NeZero P` needs it, and `hP0` cannot be dropped in favour of `hP` alone, since `⊥` *is* prime in a Dedekind domain. `[NumberField F]` is not needed and is `omit`ted; the statement holds over the bare field.

It is placed directly under the `absNormUnder` definition it is about, and all four sites are rerouted onto it:

| site | before | after |
|---|---|---|
| `encodeIdeal_pos` | 5 lines | 2 |
| `count_eq_padicValNat`, setup | 3 | 1 |
| `count_eq_padicValNat`, `hfact` | 7 | 5 |
| `count_eq_padicValNat`, off-diagonal branch | 9 | 5 |

`count_eq_padicValNat` drops from 40 lines to 32 and `encodeIdeal_pos` from 8 to 5; the file is three lines shorter overall despite gaining a declaration.

Rerouting the setup site also let the compiler confirm that `count_eq_padicValNat`'s own `have := hP` / `have : NeZero P := ⟨hP0⟩` existed *only* to feed this one fact — both are now gone.

I stopped at one helper. What remains in `count_eq_padicValNat` is the rewrite chain plus the three `Finset.sum_eq_single` branches; those are genuinely distinct cases that would each need `I`, `P`, `hP`, `hP0` and the filter threaded through, which is one-use scaffolding rather than a statement worth naming.

Gates run locally: `lake build` (7149 jobs), `lake exe axioms` (31102 declarations), `lake exe module-system` (1595 modules), `scripts/lint-env.sh` — all green.

Roadmap: none

### DIFF
--- a/TauCeti/NumberTheory/EffectiveBounds/IdealCount/Basic.lean
+++ b/TauCeti/NumberTheory/EffectiveBounds/IdealCount/Basic.lean
@@ -156,6 +156,14 @@ variable (F : Type*) [Field F] [NumberField F]
 /-- The norm of the rational prime below a prime ideal `P` of `𝓞 F`. -/
 private def absNormUnder (P : Ideal (𝓞 F)) : ℕ := Ideal.absNorm (Ideal.under ℤ P)
 
+omit [NumberField F] in
+/-- The norm of the rational prime below a nonzero prime ideal is a prime number. -/
+private theorem prime_absNormUnder {P : Ideal (𝓞 F)} (hP : P.IsPrime) (hP0 : P ≠ ⊥) :
+    (absNormUnder F P).Prime := by
+  have := hP
+  have : NeZero P := ⟨hP0⟩
+  simpa [absNormUnder] using Nat.absNorm_under_prime P
+
 /-- The coordinate (index in `Fin [F:ℚ]`) assigned to a prime ideal `P`: its
 position in the list of primes above the rational prime below `P`. -/
 private def primeCoord (P : Ideal (𝓞 F)) : ℕ :=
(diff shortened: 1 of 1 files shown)


CBirkbeck:
Closing: the target does not admit an extraction that survives `reuse`.

The repetition in `count_eq_padicValNat`/`encodeIdeal_pos` is four copies of *manufacturing instances and applying one Mathlib lemma* — `Nat.absNorm_under_prime`, which wants `IsPrime` and `NeZero` as instances while every call site holds them propositionally. I tried two shapes:

1. `prime_absNormUnder (hP : P.IsPrime) (hP0 : P ≠ ⊥)` — blocked as a thin restatement of `Nat.absNorm_under_prime` modulo unfolding the private `absNormUnder`.
2. `prime_absNormUnder_of_mem_normalizedFactors (hQ : Q ∈ normalizedFactors I)` — the composite fact, three consumers. Blocked on the same grounds.

Both blocks are right: neither helper asserts anything the located lemma does not, so naming it only relocates the unfolding. The remaining structure of `count_eq_padicValNat` is the rewrite chain plus three `Finset.sum_eq_single` branches, which would each need `I`, `P`, `hP`, `hP0` and the filter threaded through — one-use scaffolding, not statements worth naming.

Recording it here rather than reopening under a third shape.

PR #2906, opened 2026-08-12
refactor(RootSystem): name the shifted indicator sum used twice in the pairing

CBirkbeck:
The pairing identity of type `Aₙ` spent thirty-seven of its forty-one lines on two `have`s, `key` and `key'`, which are the same fact twice: summing over `Fin n` against the indicator of an index equation picks out the single matching summand. They differ only in whether the equation is `b = k` or `b = k + 1`, and correspondingly in whether the out-of-range condition is `b < n` or `b - 1 < n ∧ 1 ≤ b`.

Type `Cₙ` had already met this and named the two cases separately, as the private lemmas `sum_ite_val` and `sum_ite_val_succ` in `C/Model.lean`. So the fact was stated twice in one file and inlined twice in another.

`sum_ite_val_add` in `SimplyConnectedRootDatum/Basic.lean` states it once with the shift as a parameter: summing `f` over `Fin n` against the indicator of `b = k + j` gives `f` at the index `b - j` when that is a valid index and `j ≤ b`, and `0` otherwise. `sum_ite_val` is the case `j = 0` and `sum_ite_val_succ` the case `j = 1`; both are deleted, and their two call sites in type `C` now use the shared lemma, as do the two `have`s in type `A`.

It lives in a new `TauCeti/Data/Fin/Basic.lean`, not in the root-datum scaffolding: it mentions no root-system concept, needing only `Fin`, finite sums, `ℕ` and an additive monoid. Putting it with the root data would have made `C/Model.lean` import `DynkinType` and `RootSystem.Base` for an indexing fact. Both `A.lean` and `C/Model.lean` import the new module.

The summand type is an arbitrary `AddCommMonoid`, not `ℤ`. Both deleted lemmas fixed `f : Fin n → ℤ`, but the argument uses only finite addition and zero; the integer call sites specialise it.

Mathlib does not have this. `Fintype.sum_ite_eq` is keyed on an equation in the index type itself; here the equation is between a natural number and the `Fin.val` of the index, which is why the argument needs the range case split and a `Fin.ext` step. The new lemma is that bridge, and hands the `Fin`-keyed Mathlib lemma the shape it wants.

Both degenerate ends are why the condition has two conjuncts. At `n = 0` the sum is empty and `b - j < 0` is false, so both sides are `0`. At `b < j` — in particular `b = 0, j = 1` — the equation `b = k + j` has no solution, and it is the `j ≤ b` conjunct rather than the index bound that makes the right-hand side vanish; without it the statement would be false at `b = 0`, where `b - j` truncates to `0`.

Type `Bₙ`'s pairing identity is *not* a fourth instance and is left alone: it sums over `Finset.range n` rather than `Fin n`, its summands carry the index-dependent short/long root factors `if m + 1 = n then 2 else 1`, and it picks out its term with `Finset.sum_eq_single` rather than by an indicator.

The type `A` parent drops from 41 lines to 16, and the repository loses two declarations and 22 lines net.

Gates run locally: `lake build`, `lake exe axioms`, `lake exe module-system`, `scripts/lint-env.sh` — all green.

Roadmap: RepresentationTheory

### DIFF
--- a/TauCeti/Data/Fin/Basic.lean
+++ b/TauCeti/Data/Fin/Basic.lean
@@ -1,0 +1,45 @@
+/-
+Copyright (c) 2026 The Tau Ceti contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
+module
+
+public import Mathlib.Algebra.BigOperators.Fin
+
+/-!
+# Indicator sums over `Fin n` keyed by a natural number
+
+`Fintype.sum_ite_eq` evaluates a sum whose indicator compares two elements of the index type.
+When the comparison is instead between a natural number and the `Fin.val` of the index — as it is
+whenever a family is indexed by `ℕ` and summed over `Fin n` — the index may fall outside the
+range, so the value is a `dite` rather than a plain application.
+
+## Main results
+
+* `TauCeti.sum_ite_val_add`: a sum against the indicator of `b = k + j` picks out the summand at
+  `b - j`, or vanishes when there is no such index.
+-/
+
+public section
+
+namespace TauCeti
+
+/-- **A shifted indicator picks out one summand.** Summing `f` over `Fin n` against the indicator
+of `b = k + j` gives `f` at the index `b - j` when that is a valid index and `j ≤ b`, and `0`
+otherwise. -/
+theorem sum_ite_val_add {M : Type*} [AddCommMonoid M] {n : ℕ} (f : Fin n → M) (b j : ℕ) :
+    ∑ k : Fin n, (if b = (k : ℕ) + j then f k else 0)
+      = if h : b - j < n ∧ j ≤ b then f ⟨b - j, h.1⟩ else 0 := by
+  by_cases h : b - j < n ∧ j ≤ b
+  · have hb : ∀ k : Fin n, (b = (k : ℕ) + j) = (k = (⟨b - j, h.1⟩ : Fin n)) := by
+      intro k
+      have := h.2
+      simp only [Fin.ext_iff, eq_iff_iff]
+      omega
+    simp only [hb, dite_eq_left h, Finset.sum_ite_eq', Finset.mem_univ, ite_true]
+  · rw [dite_eq_right h, Finset.sum_eq_zero]
+    intro k _
+    have := k.isLt
+    exact ite_eq_right (by omega)
+
+end TauCeti
--- a/TauCeti/LinearAlgebra/RootSystem/SimplyConnectedRootDatum/A.lean
+++ b/TauCeti/LinearAlgebra/RootSystem/SimplyConnectedRootDatum/A.lean
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 -/
 module
 
+public import TauCeti.Data.Fin.Basic
 public import TauCeti.LinearAlgebra.RootSystem.SimplyConnectedRootDatum.Basic
 
 public section
--- a/TauCeti/LinearAlgebra/RootSystem/SimplyConnectedRootDatum/A.lean
+++ b/TauCeti/LinearAlgebra/RootSystem/SimplyConnectedRootDatum/A.lean
@@ -94,31 +94,6 @@ the coordinates of `e_a - e_b`; the `k`-th potential is `[a ≤ k]`. -/
 private def typeACoweight (n a : ℕ) : Fin n → ℤ :=
   fun k => if a ≤ (k : ℕ) then 1 else 0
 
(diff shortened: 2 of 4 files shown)


CBirkbeck:
## AI review — approved

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics).

|---|---|---|---|---|
| ✅ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics/correctness.md) | approved | `codex/gpt-5.6-sol` | The shifted indicator theorem faithfully handles valid, out-of-range, underflow, and empty-index cases, and all four migrated call sites are correct specializations. |
| ✅ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics/reuse.md) | approved | `codex/gpt-5.6-sol` | The shifted-indicator lemma consolidates the duplicated local proofs while reusing `Finset.sum_ite_eq'`; searches of TauCeti and Mathlib found no existing declaration providing the same shifted `Nat`/`Fin.val` bridge. |
| ✅ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics/scope.md) | approved | `codex/gpt-5.6-sol` | This is a single-topic refactor: the generalized indicator-sum lemma consolidates four existing proofs and immediately replaces them without introducing unrelated mathematical content. |
| ✅ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics/attribution.md) | approved | `codex/gpt-5.6-sol` | The lemma is a routine generalization of existing local proofs, and the module explicitly credits the central Mathlib result it builds on; no missing attribution was identified. |
| ✅ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics/api-design.md) | approved | `codex/gpt-5.6-sol` | The canonical public helper is reusable, appropriately located and generalized, while the obsolete private duplicates are removed; its theorem-only interface needs no additional characteristic API or automation annotation. |
| ✅ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics/generality.md) | approved | `codex/gpt-5.6-sol` | The shared lemma now has the natural generality: an arbitrary additive commutative monoid, with the shift parameter covering both existing special cases. |
| ✅ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics/placement.md) | approved | `codex/gpt-5.6-sol` | The generic finite-index sum lemma is naturally placed in `TauCeti/Data/Fin/Basic.lean`, and both added imports are required by confirmed call sites. No evidently wrong import or placement issue is present. |
| ✅ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics/naming.md) | approved | `codex/gpt-5.6-sol` | The introduced name follows existing `sum_ite_eq` precedent and correctly identifies the `Fin.val` coercion and additive shift; no notation is introduced. |
| ✅ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics/documentation.md) | approved | `codex/gpt-5.6-sol` | The new substantive file has a clear module docstring, and `sum_ite_val_add` is documented accurately in terms of its statement and boundary conditions. |
| ✅ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics/proof-quality.md) | approved | `codex/gpt-5.6-sol` | The new shared proof is short, explicit, and robust; its call sites use controlled simplification without undocumented definitional equality. |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/6b0ff373f14030da9589cf87ddb65b75154a8cfb...fa73996c89f180cf09b948de73012016823e71ff) at head `fa73996`. rubrics @ [`a66b04d`](https://github.com/TauCetiProject/TauCetiReview/tree/a66b04d97721565b13c8abe07f750fb1d2959ad5/rubrics). Review spend: $1.43.</sub>

PR #2955, opened 2026-08-12
refactor(Bernstein): name the iterated-derivative shift behind the Chafai density recursion

CBirkbeck:
`chafaiDensity_neg_derivWithin_pred` relates the Chafaï density of `-f'` at order `n-1` to that of
`f` at order `n`. Its forty-six lines are mostly arithmetic bookkeeping — factorials, powers of `t`,
and the sign `(-1)^(n-1)` versus `(-1)^n` — but ten of them proved a fact about iterated derivatives
that has nothing to do with densities, factorials or Chafaï.

That fact is now `iteratedDerivWithin_neg_derivWithin_succ` in
`TauCeti/Analysis/Calculus/IteratedDerivWithin.lean`: the `m`-th iterated derivative of `-f'` is
minus the `(m+1)`-st iterated derivative of `f`.

It is stated unconditionally, over an arbitrary set, in the `m` / `m+1` form rather than the
`n-1` / `n` form the caller happens to need. That is what the identity actually says: nothing in the
argument uses the half-line `Ici 0`, the completely-monotone hypothesis on `f`, the bound `2 ≤ n` or
the positivity `0 < t`, and — once the orders are written as `m` and `m+1` rather than `n-1` and `n`
— no hypothesis on the order either. Truncated subtraction was the only reason a side condition
looked necessary; stating the shift directly removes it, and the proof loses the `congr`/`omega`
steps that were reconciling `n - 1 + 1` with `n`. The caller supplies `m := n - 1` and discharges
`n - 1 + 1 = n` itself, which is where that arithmetic belongs.

It goes in the shared calculus file rather than staying beside its consumer. That file exists to
record "calculus lemmas about `iteratedDerivWithin` that are independent of any completely-monotone
or Bernstein-function structure", which is precisely what this is; leaving it in the Bernstein file
would have put a generic derivative identity behind a Bernstein import. `Measures.lean` imports it
privately, since it is used only inside a proof.

The parent drops from 46 lines to 35; the lemma is 4.

Gates run locally: `lake build`, `lake exe axioms`, `lake exe module-system`, `scripts/lint-env.sh`
— all green.

Roadmap: none

Cross-cutting calculus infrastructure: the lemma mentions only `iteratedDerivWithin`, `derivWithin`
and an arbitrary set, and joins an existing file of such lemmas.

### DIFF
--- a/TauCeti/Analysis/Calculus/IteratedDerivWithin.lean
+++ b/TauCeti/Analysis/Calculus/IteratedDerivWithin.lean
@@ -16,6 +16,8 @@ completely-monotone or Bernstein-function structure.
 
 * `TauCeti.ContDiffOn.hasDerivAt_iteratedDerivWithin`: differentiability of an
   `iteratedDerivWithin` on a neighbourhood inside a unique-differentiability set.
+* `TauCeti.iteratedDerivWithin_neg_derivWithin_succ`: differentiating the negated derivative
+  absorbs one order.
 For the plain fundamental-theorem identity on a compact interval use Mathlib's
 `intervalIntegral.integral_derivWithin_Icc_of_contDiffOn_Icc` together with
 `iteratedDerivWithin_one`.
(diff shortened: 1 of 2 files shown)


CBirkbeck:
Closing: the extraction does not survive review, and `reuse` is right.

`iteratedDerivWithin_neg_derivWithin_succ` was rejected by `reuse` in two materially different shapes — first as a `private`, `n ≠ 0`-conditional helper beside its consumer, then as an unconditional public lemma in the shared calculus file. Both times the finding was the same and correct: the statement is a thin composite of Mathlib's `iteratedDerivWithin_neg` and `iteratedDerivWithin_succ'`, and its only extra content was the `n - 1 + 1 = n` bookkeeping, which belongs at the call site.

On the second round `api-design` (add `@[simp]`) and `generality` (restate over `𝕜 → F`) asked to keep and enhance the lemma, contradicting `reuse`'s request to delete it. Rather than satisfy one and let the other re-fire, I am taking `reuse`'s side: a helper has to assert something no located lemma does, and this one does not.

Taking that fix leaves no decompose win — the parent returns to its original length — so there is nothing left worth shipping here. Recording the obstacle instead: `chafaiDensity_neg_derivWithin_pred` has exactly one separable block, and that block is a two-lemma Mathlib composition.

PR #3006, opened 2026-08-12
feat(PathSpace): move a block through a reindexing over an invariant event

cameronfreer:
```lean
theorem setLIntegral_comp_reindex_eq_of_measurableSet_invariants
    {ρ : Measure (ℕ → α)} {φ : ℕ → ℕ} {m C : ℕ}
    (hmp : MeasurePreserving (fun x : ℕ → α => fun k => x (φ k)) ρ ρ)
    (hφ_add : ∀ n, m ≤ n → φ n = n + C)
    {A : Set (ℕ → α)} (hA : MeasurableSet[MeasurableSpace.invariants (shift α)] A)
    {f : (ℕ → α) → ℝ≥0∞} (hf : Measurable f) :
    ∫⁻ x in A, f (fun k => x (φ k)) ∂ρ = ∫⁻ x in A, f x ∂ρ

theorem ContractableLaw.setLIntegral_block_eq_prefix_of_measurableSet_invariants
    {ρ : Measure (ℕ → α)} (hρ : ContractableLaw ρ) {m : ℕ} {k : Fin m → ℕ} (hk : StrictMono k)
    {A : Set (ℕ → α)} (hA : MeasurableSet[MeasurableSpace.invariants (shift α)] A)
    {g : (Fin m → α) → ℝ≥0∞} (hg : Measurable g) :
    ∫⁻ x in A, g (fun i => x (k i)) ∂ρ = ∫⁻ x in A, g (prefixProj α m x) ∂ρ
```

Over a shift-invariant event, an eventually-translating measure-preserving reindexing changes no
set-integral; a strictly increasing finite selection can therefore be displaced onto the prefix.
The middle declaration is the contractable instance, where strict monotonicity supplies the measure
preservation.

## Invariance, not tail-measurability

These rest on `shift ⁻¹' A = A`, in the `MeasurableSpace.invariants`-measurable form. A tail event
need not satisfy that, and `invariants_shift_lt_pathTail` shows the inclusion is strict. This is
the substantive difference from the `L²` route, whose block comparison is *distributional* — an
equality of conditional laws given the tail — rather than an actual invariance of the test event.
The two are deliberately not abstracted into one σ-algebra-parametric theorem, and the module
docstring says so.

## Relation to the auto-closed #2065

#2065 carried this theorem inside a twelve-file reorganization and was retired at the review round
cap after five consecutive `placement` rounds. This is the one-file successor: only the transport
theorem, its contractable instance, and the block specialization. Everything else it needed has
landed independently since —
[...]

PR #3008, opened 2026-08-12
feat(ViaKoopman): reduce a block to the prefix over an invariant event

cameronfreer:
> **Draft — depends on #3006.** Its diff currently contains that PR's file as well; once #3006
> merges I will rebase onto fresh main and confirm the diff is this endpoint only. Opened as a
> draft rather than held locally so the work is visible while #3006 is in review.

```lean
theorem ContractableLaw.measure_inter_blockCylinder_eq_prefix_of_invariantConditional
    [StandardBorelSpace α] [Nonempty α]
    {ρ : Measure (ℕ → α)} [IsFiniteMeasure ρ] (hρ : ContractableLaw ρ)
    {r : ℕ} {k : Fin r → ℕ} (hk : StrictMono k) {B : Fin r → Set α}
    (hB : ∀ i, MeasurableSet (B i)) {S : Set (ProbabilityMeasure α)} (hS : MeasurableSet S) :
    ρ ((invariantConditionalProbabilityMeasure ρ ⁻¹' S)
        ∩ blockCylinder (fun j (x : ℕ → α) => x j) k B)
      = ρ ((invariantConditionalProbabilityMeasure ρ ⁻¹' S)
        ∩ blockCylinder (fun j (x : ℕ → α) => x j) (fun i : Fin r => (i : ℕ)) B)
```

The Koopman route's first reduction. The witness is measurable for
`MeasurableSpace.invariants (shift α)`, so a test event it cuts out is invariant, and the block
transport of #3006 applies directly: an arbitrary strictly increasing block has the same mass as
the prefix block. Only one block position then needs analyzing downstream.

## No conditional expectation on this route

The `L²` route reaches the corresponding step through
`Contractable.condExp_block_comp_tailProcess_ae_eq` — an a.e. identity of *conditional
expectations*, i.e. a distributional statement, carrying integrability side conditions.

Here the test event is genuinely invariant (`shift ⁻¹' A = A`), so the block is displaced inside
the integral itself and the reduction stays in `ℝ≥0∞` throughout, with no conditional expectation
and no integrability hypotheses. This is the concrete form of the difference between the routes.
They are deliberately not unified into a σ-algebra-parametric statement:
`invariants_shift_lt_pathTail` shows the underlying σ-algebras genuinely differ, and the module
docstring records that.

## What remains for the factorization

This supplies selection-invariance. The substantive step still ahead is the decoupling — that the
*prefix* block factorizes into `∏ i, ν (B i)` — via Birkhoff/mean-ergodic convergence and bounded
product convergence, followed by the `ℝ≥0∞` common ending and the path-space-to-`Ω` transfer
through `ConditionallyIIDWith.of_pathLaw`.

Roadmap: Exchangeability

Advances `TauCetiRoadmap/Exchangeability/README.md`, **Layer 5** (Koopman operators and invariant
σ-algebras), milestone `deFinetti_viaKoopman`.

Intention: TauCetiProject/TauCetiRoadmap#133 (claimed).

🤖 Directed by Cameron Freer, drafted by Opus 5, reviewed by GPT 5.6 Sol


cameronfreer:
Superseded by #3152, now that #3145 has landed.

This was opened as a draft against the unmerged #3006 lineage, so its diff carried that PR's files as well. #3152 is the same theorem rebuilt on fresh `main`: a single new file, endpoint-only diff, with the transport's final name (`map_restrict_prefixProj_of_strictMono_of_measurableSet_invariants`) applied.

Closing in favour of #3152.

PR #3716, opened 2026-08-18
ci: publish TauCeti's oleans from the merge-queue build

kim-em:
This PR publishes TauCeti's oleans from the merge-queue build that gates a commit, so the Lake artifact cache stays in step with `main` instead of trailing it by however long the post-merge publisher takes.

The merge-group build already compiles the exact tree that is about to land, and GitHub's merge queue reuses the merge-group commit verbatim as main's new tip: each group's `head_sha` is the next group's `base_sha`, and every one of them is an ancestor of `main`. A mapping stored under that revision is therefore exactly what a later `lake cache get` finds, directly or by ancestor backtracking. Publishing there removes the window in which merges outrun publication and land on the full-rebuild path.

The trust boundary of #313 is preserved. The publish step is gated on `merge_group`, so `secrets.LAKE_CACHE_KEY` is never evaluated in a `pull_request_target` run and a fork PR cannot reach the write key. Nothing in the step compiles: `sandbox-build.sh` emits the input-to-output mappings with `lake build --no-build` inside landrun, where `--no-build` reports an out-of-date target rather than building it, and the step outside only reads that JSONL and PUTs the content-addressed artifacts the sandboxed build already packed. Uploading an artifact is safe independently of the revision it is attributed to, because artifacts are keyed by Lake input hash: a module is served only to a build whose sources hash to the same inputs.

`lake cache put` names the mapping after the package's current Git revision, and `base/` is checked out at the group's base commit with the group's `TauCeti/` overlaid, so the step points HEAD at the merge-group commit first. `git reset --mixed` moves HEAD and the index only and never rewrites a working-tree file, so the build outputs and the Lake traces beside them are untouched. It then refuses to publish unless `TauCeti/`, `TauCeti.lean`, the lakefile, and both Lake pins match that commit exactly, which is what makes the published mapping an honest description of the revision it is filed under. `scripts/` is excluded from that comparison on purpose: the trusted tooling staged from `github.workflow_sha` legitimately differs from the base commit's copy, and no TauCeti module's input hash depends on it.

Publishing is best-effort throughout. A missing mapping, a ref that does not resolve to the built commit, an overlay that does not reproduce the tree, or a failed upload all warn and skip, and the step is `continue-on-error`, so a batch that built, audited, and linted cleanly is never held up by the cache. `ci.yml` keeps publishing as a backstop, and since it publishes the same revision with the same content the two are idempotent.

Roadmap: none

🤖 Prepared with Claude Code

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -507,6 +507,99 @@ jobs:
             --env WATCHDOG_TOOLCHAIN \
             -- bash -euxo pipefail -c 'cd base && exec bash scripts/sandbox-build.sh'
 
+      # --- Lake artifact-cache publish (merge queue only; dormant until configured) ---
+      # Publish the oleans of the commit that is about to land. GitHub's merge queue reuses the
+      # merge-group commit verbatim as main's new tip (each group's head_sha is the next group's
+      # base_sha, and an ancestor of main), so a mapping stored under that revision is exactly
+      # what a later `lake cache get` finds, directly or by ancestor backtracking. Publishing
+      # here keeps the cache in step with main rather than an hour behind it; a stale cache is
+      # what puts merge-group builds on the full-rebuild path, which slows the queue, which
+      # keeps the cache stale.
+      #
+      # The trust boundary of #313 is preserved. The step is gated on merge_group, so
+      # `secrets.LAKE_CACHE_KEY` is never evaluated in a pull_request_target run and a fork PR
+      # cannot reach the write key. Nothing here compiles: the mappings were emitted by
+      # `lake build --no-build` INSIDE landrun, and this step only reads that JSONL and PUTs
+      # the content-addressed artifacts the sandboxed build already packed. Uploading an
+      # artifact is safe independently of the revision it is attributed to, because artifacts
+      # are keyed by Lake input hash: a module is served only to a build whose sources hash to
+      # the same inputs.
+      - name: Publish the merge group's oleans to the Lake cache
+        if: ${{ github.event_name == 'merge_group' && env.INFRA != '1'
+                && env.TAUCETI_CACHE_ENABLED == '1' && steps.build.outcome == 'success' }}
+        # Never fail the queue over a publish. A missed upload costs some later build its cache;
+        # a red step here would hold up a batch that built, audited, and linted cleanly.
+        continue-on-error: true
+        env:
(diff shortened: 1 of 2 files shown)


kim-em:
Closing: the central safety claim in this description is wrong, and there is no version of the approach worth keeping.

The claim was that publishing from the merge-queue build is safe because the mappings are produced by `lake build --no-build` inside landrun, so no untrusted elaboration happens outside the sandbox. That protects the publish step, but the artifacts and the mappings were already writable before it ran. landrun mounts `base/.lake` read-write, and Lean executes code at elaboration time: a `run_cmd` doing `IO.FS.writeFile` writes its file, exits 0, and logs nothing, so `lake build --iofail` passes it. Only `set_option` is banned at source level.

No hash collision is needed to exploit that, which is what I had originally assumed. It is enough to place any olean in `.lake/cache/artifacts` under its own correct digest and write `outputs.jsonl` mapping the expected input hash to that digest. Content addressing does not help, because the forged object is the mapping rather than the artifact.

Reading Lake confirms nothing downstream re-derives the relationship:

- `lake cache put` loads the mapping, checks the artifact paths exist, and uploads (`Lake/CLI/Main.lean:642`, `Lake/Config/Cache.lean:1016`).
- `lake build --no-build -o` failing to write its mappings is logged, not propagated (`Lake/Build/Run.lean:277`), so the `$?` check in `sandbox-build.sh` can succeed while a planted `outputs.jsonl` survives.
- Lake trusts `.hash` sidecars unless rehashing is requested (`Lake/Build/Common.lean:395`), and `Lake.Hash` is a single `UInt64` with an upstream TODO to replace it with a secure hash (`Lake/Build/Trace.lean:98`).

Two further defects, now moot but worth recording: the step condition checks `env.INFRA != '1'` but not `OUT_OF_SCOPE`, so a human-merged infrastructure group could reach the publisher; and a `merge_group` run may take its workflow from the merge-group commit rather than the base, which would let a queued PR edit the gate before it lands.

The credential was never installed in this workflow, so `LAKE_CACHE_KEY` does not need rotating.

Publishing state produced by pre-merge, PR-controlled elaboration cannot be made safe without provenance that binds an artifact to its source cryptographically, which Lake does not currently provide. The upload stays in the push-to-`main` workflow. https://github.com/TauCetiProject/TauCeti/pull/3715 makes that publisher cheap enough that this optimisation is not needed.

PR #3665, opened 2026-08-17
feat: define the cut norm of symmetric kernels

kbuzzard:
This PR defines the measurable-rectangle cut norm of strict symmetric kernels and proves its core analytic API: bounded-kernel integrability on finite measure spaces, rectangle-integral algebra and transposition, the two supremum rules, nonnegativity, zero, negation, subadditivity, absolute homogeneity, and the bound `‖K‖□ ≤ ∫ |K|`.

The exact target is `TauCetiRoadmap/DenseGraphLimits/README.md`, Layer 1, milestone “core objects and their basic API,” specifically `cutNorm` on `SymmKernel` together with its set form, seminorm laws, and `L¹` bound. This is the most effective current step because `SymmKernel` is merged, the `Graphon` carrier is independently in flight in #3637, and the cut norm now opens the cut-distance lane immediately while also supplying the norm used by Layer 2’s counting and weak-regularity milestones. After this PR, the cut-norm lane still needs the signed/test-function form and factor-four sandwich; Layer 1 separately still needs homomorphism densities, the graphon carrier from #3637, coupling-primary cut distance, and the metric quotient.

Add `TauCeti/Combinatorics/DenseGraphLimits/Kernel/CutNorm.lean` (276 lines). The finite-measure statements specialize directly to the roadmap’s probability spaces, and the public API keeps a.e. invariance out of this strict-carrier layer as required.

The definition and rectangle-integral interface are adapted from Cameron Freer’s `Graphon/CutNorm.lean` at pinned commit `6eccca5`; the full strict-kernel seminorm and `L¹` proofs are developed here. The module documentation also credits Lovász, *Large Networks and Graph Limits*, §8.2.1, and Janson, *Graphons, cut norm and distance, couplings and rearrangements*, §4. No Mathlib infrastructure is vendored; the implementation reuses Mathlib’s Bochner set integral, product-measure swap, conditional supremum, and finite-measure integrability APIs.

Roadmap: DenseGraphLimits

🤖 Prepared with Codex

PR #3640, opened 2026-08-17
feat(Distributions): exponential moments and the mgf of the uniform law

cameronfreer:
Continues Layer 0 item 3 on the uniform family, picking up where #3575 left off (measure, pdf,
`withDensity`/`HasPDF`/`rnDeriv`, cdf, mean, variance).

## Every exponential moment exists

```lean
theorem integrableExpSet_id_uniformMeasure {a b : ℝ} :
    integrableExpSet id (uniformMeasure a b) = Set.univ
```

Bounded support is the whole reason: `exp (t * x)` is bounded almost everywhere by
`exp (|t| * max |a| |b|)`, and the measure is finite. **No hypothesis on the endpoints is needed** —
this holds on the degenerate range too. Integrability is established explicitly, via
`Integrable.mono'` against a constant, before any integral is evaluated.

Two supporting results land with it:

* `ae_mem_Ioc_uniformMeasure` — the law is carried by its interval, which is what converts bounded
  support into an a.e. bound;
* an `IsFiniteMeasure` instance holding for **every** pair of endpoints: a probability measure when
  `a < b`, the zero measure otherwise. The degenerate range is covered rather than excluded.

## The mgf, split at the singularity

`(exp (t * b) - exp (t * a)) / ((b - a) * t)` has a removable singularity at `t = 0`. Rather than
push a proof through it, the two branches are separate statements:

| | |
|---|---|
| `mgf_id_uniformMeasure_zero` | `mgf id (uniformMeasure a b) 0 = 1` |
| `mgf_id_uniformMeasure` | the quotient, for `t ≠ 0` |

Everything routes through the existing `uniformMeasure_eq_smul`; `ProbabilityTheory.cond` is never
unfolded again.

## Why no cgf here

The cgf is the real logarithm of the mgf, and writing it that way is only meaningful once the mgf is
shown **positive** — a genuine argument, not a syntactic rewrite. It is left for its own PR rather
than appended here.

Also still open in item 3: the characteristic function (again splitting `t = 0`, with the complex
calculation kept clear of this real one), the affine transport, and parameter measurability.

Roadmap: StandardDistributions

Authorized by `TauCetiRoadmap/StandardDistributions/README.md`, Layer 0, item 3 — the
`integrableExpSet` and mgf targets.

Advances TauCetiProject/TauCetiRoadmap#242. It deliberately does **not** close it: the uniform
family still owes the cgf, the characteristic function, and parameter measurability. The last of
those PRs should carry the `Closes`.

[...]

PR #3557, opened 2026-08-17
feat(DenseGraphLimits): finite-graph homomorphism density estimators

cameronfreer:
The graphon-independent front of Layer 9a: the two finite-graph homomorphism density estimators and
the counting bridge that pins their normalization to Mathlib.

```lean
noncomputable def homDensityFin (F : SimpleGraph V) (G : SimpleGraph (Fin m)) : ℝ :=
  (Nat.card (F →g G) : ℝ) / (m ^ Fintype.card V : ℝ)

noncomputable def injHomDensity (F : SimpleGraph V) (G : SimpleGraph (Fin m)) : ℝ :=
  (Nat.card {φ : F →g G // Function.Injective φ} : ℝ) / (m.descFactorial (Fintype.card V) : ℝ)
```

## The falling factorial, not the binomial coefficient

`injHomDensity`'s numerator counts **ordered** injective homomorphisms, so the denominator must
count ordered injections: `m.descFactorial k` with `k = |V(F)|`. `Nat.choose m k` counts unordered
images, and using it would bias the sampling estimator by `k!` — the later unbiasedness identity
would read `k! · t(F, W)` instead of `t(F, W)`. Fixing the convention here keeps that correction out
of every downstream statement.

## One counting convention, not two

`card_injective_hom_eq_labelledCopyCount` identifies the numerator with Mathlib's
`SimpleGraph.labelledCopyCount`, **proved**, and it ships in this PR deliberately: without it,
`injHomDensity` would quietly establish a second counting convention alongside Mathlib's.
`SimpleGraph.Copy F G` is definitionally an injective homomorphism, so the bridge is an equivalence
of subtypes. Note Mathlib puts the *host* graph first — copies of `F` inside `G` are
`G.labelledCopyCount F`.

## Counting with `Nat.card`

`Nat.card` is total, so neither density needs a `Fintype` instance or decidability on the hom type,
on `G`, or on `G.Adj`. The counted types are finite here so nothing is lost — but the analytic
statements downstream do not inherit decidability hypotheses they would otherwise be stuck with.
Both definitions are `noncomputable` because the value is real.

## Scope

Excludes the hom-versus-injective closeness bound (`homDensityFin_sub_injHomDensity_le`), the
sampling laws, the unbiasedness anchor, and finite-graph graphons. The legacy `SimpleGraph.t` /
`tInj` API is not used or reproduced. Nothing here touches the graphon carrier, so this is
independent of #3551.

Closes TauCetiProject/TauCetiRoadmap#240

[...]

PR #3844, opened 2026-08-19
feat(DenseGraphLimits): the signed cut norm and the lower sandwich bound

cameronfreer:
The `[-1,1]` test-function form of the cut norm, on top of the set form that landed in #3665.

## What makes this a *form* of the cut norm, not a second definition

The two suprema differ only in what they range over: indicators for `cutNorm`, measurable
`[-1,1]`-valued functions for `cutNormSigned`. To make that literal rather than analogical, this PR
introduces the pairing

```lean
def testIntegral (K : SymmKernel Ω μ) (u v : Ω → ℝ) : ℝ :=
  ∫ p, u p.1 * v p.2 * K p.1 p.2 ∂(μ.prod μ)
```

and proves `testIntegral_indicator_one : K.testIntegral μ (S.indicator 1) (T.indicator 1) =
K.rectIntegral μ S T`. `cutNorm_le_cutNormSigned` is then exactly "the signed supremum ranges over
more pairs" — one rewrite, no analysis.

## Only the easy side is here

| | |
|---|---|
| `cutNorm_le_cutNormSigned` | this PR |
| `cutNormSigned_le_four_mul_cutNorm` | **not** here — separate PR |

The factor-4 direction needs a genuine argument (reduce to `±1`-valued test functions via the sign
of an inner integral, then decompose into the four rectangles cut out by `{u = 1}` and `{v = 1}`),
so it is its own topic. The module docstring and the `cutNormSigned` docstring both record the
absence as a deliberate split rather than an oversight.

## `cutNormSigned` requires `[IsFiniteMeasure μ]`

An earlier revision of this PR omitted it and argued in this description that the supremum "needs no
finiteness hypothesis". **That was wrong**, and the `correctness` block was right to reject it. On an
infinite measure the test integrals are unbounded — take `μ` Lebesgue, `K` the constant kernel `1`,
and `u = v` the indicator of `[0, n]`, giving `n²` — so `⨆` on `ℝ` collapses to its junk value `0`
for a kernel nowhere near zero. No *theorem* in the PR was false, since every result that needs the
bound already carried the instance; the defect was that the definition returned junk outside its
intended domain.

What drove the original mistake is worth recording: carrying the instance made the
unused-section-variable linter fire, and I treated a linter workaround as a design principle instead
of asking why `cutNormSet` was written the way it is. It routes its measure through `FiniteMeasure`
precisely so the instance is a *used* argument; `cutNormSigned` now does the same, and sits beside
the other two definitions so all three carry the same hypothesis.

## Roadmap

Roadmap: DenseGraphLimits

`TauCetiRoadmap/DenseGraphLimits/README.md`, Layer 1, names "the set-form / signed cut norm
(`cutNormSet` + `cutNorm_eq_cutNormSet`, `cutNormSigned` + the factor-4 sandwich)"; the signature
follows `TauCetiRoadmap/DenseGraphLimits/Suggested.lean`, which pins `cutNormSigned`,
`cutNorm_le_cutNormSigned`, and `cutNormSigned_le_four_mul_cutNorm` verbatim. No intention issue is
open for this target, so this PR carries no `Closes`.

🤖 Directed by Cameron Freer, drafted by Opus 5


TauCeti/Combinatorics/DenseGraphLimits/Kernel/CutNorm.lean
kbuzzard, line 1: ### ✅ correctness — now passing on `cc4c756`.

kbuzzard:
## AI review — approved

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics).

|---|---|---|---|---|
| ✅ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/correctness.md) | approved | `codex/gpt-5.6-sol` | The finite-measure requirement makes the signed supremum bounded, and the test-integral definition, indicator specialization, and lower sandwich bound faithfully formalize the intended mathematics. |
| ✅ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/reuse.md) | approved | `codex/gpt-5.6-sol` | No material duplication or unnecessary parallel API was identified; the new signed cut norm reuses the existing rectangle-integral and cut-norm interfaces. |
| ✅ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/scope.md) | approved | `codex/gpt-5.6-sol` | The PR coherently advances the DenseGraphLimits Layer 1 signed cut norm and factor-4 sandwich target by adding the signed form, its basic interface, and the lower bound as one topic. |
| ✅ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/attribution.md) | approved | `codex/gpt-5.6-sol` | The module credits the central informal references, roadmap signatures, and the identifiable prior `cameronfreer/graphon` formalization with repository, license, and commit details. |
| ✅ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/api-design.md) | approved | `codex/gpt-5.6-sol` | The characteristic signed-cut-norm API is present, and `testIntegral_indicator_one` now has the required `@[simp]` normalization annotation. |
| ✅ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/generality.md) | approved | `codex/gpt-5.6-sol` | The new declarations are stated at the natural level for finite-measure symmetric kernels, with hypotheses matching the signed cut-norm test class; no materially unused or excessively strong assumptions are evident. |
| ✅ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/placement.md) | approved | `codex/gpt-5.6-sol` | The pairing, signed cut norm, and associated bounds belong naturally in the existing cut-norm module; the diff adds no imports or misplaced files. |
| ✅ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/naming.md) | approved | `codex/gpt-5.6-sol` | The introduced public names describe their conclusions in standard Mathlib terminology, respect suffix semantics, and add no notation. |
| ✅ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/documentation.md) | approved | `codex/gpt-5.6-sol` | The added module and declaration documentation accurately describes the signed cut norm and its interfaces. The contested infinite-measure counterexample is valid because `SymmKernel` requires boundedness, while integrability depends on `[IsFiniteMeasure μ]`. |
| ✅ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/proof-quality.md) | approved | `codex/gpt-5.6-sol` | The proofs are appropriately factored, readable, and robust, with no undocumented definitional-equality manipulation or brittle rewrite chains. |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/159b8ac44553f380d444efe88022afd53ea5d6a9...cc4c756b3c32e7621147cffa30df4fb02c6538f0) at head `cc4c756`. rubrics @ [`85349e8`](https://github.com/TauCetiProject/TauCetiReview/tree/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics). Review spend: $2.67.</sub>

cameronfreer:
Conceded — the block is correct, and the PR description that argued the opposite has been rewritten rather than quietly dropped.

**The defect is real.** With `μ` Lebesgue, `K` the constant kernel `1`, and `u = v` the indicator of `[0, n]`, the test integral is `n²`, so the range is unbounded and `⨆` on `ℝ` collapses to its junk value `0` for a kernel that is nowhere near zero. `cutNormSigned` now requires `[IsFiniteMeasure μ]`, and the docstring states that counterexample rather than just asserting the hypothesis is needed.

To be precise about the blast radius, since "correctness block" could suggest a false theorem: no theorem in the PR was wrong. Every result that needs the `L¹` bound — the introduction rule, the `L¹` bound itself, `cutNorm_le_cutNormSigned` — already carried the instance, and the two that did not (`cutNormSigned_def`, `cutNormSigned_nonneg`) are true for the junk value too. The defect was confined to the definition being unfaithful outside finiteness, which is exactly what the finding says.

**Why I got it wrong is the part worth recording.** Carrying the instance made the unused-section-variable linter fire on `cutNormSigned_def`, so I dropped it and wrote a paragraph justifying the omission. That is a linter workaround dressed up as a design principle. The right question was why `cutNormSet` looks the way it does: it routes its measure through `FiniteMeasure` (`let ν : FiniteMeasure Ω := ⟨μ, inferInstance⟩`) precisely so the instance is a *used* argument. `cutNormSigned` now does the same and sits beside the other two definitions, so all three carry the same hypothesis and the three `omit` clauses are gone.

Gate green (BUILD/AX/MS/LINT all 0). Note the other nine rubrics have not run yet — the pipeline halted here — so I expect further rounds.

TauCeti/Combinatorics/DenseGraphLimits/Kernel/CutNorm.lean
kbuzzard, line 1: ### ✅ api-design — now passing on `cc4c756`.

cameronfreer:
Applied — `testIntegral_indicator_one` is now `@[simp]`.

I checked it against simp-NF before pushing rather than just adding the attribute, because `lake build` accepts a bad `@[simp]` and only `scripts/lint-env.sh` rejects one: two earlier reviewer-requested annotations in this repo (#3557 `card_injective_hom_eq_labelledCopyCount`, #3640 `mgf_id_uniformMeasure_zero`) turned out to have left-hand sides that were not in normal form, so the lemma would never have fired. This one is clean — full gate green (BUILD/AX/MS/LINT all 0), and no existing proof needed adjusting for the new rewrite.

The finding is right about the role it plays: the pairing generalizes the rectangle integral, and the indicator case is exactly the bridge back to the established `rectIntegral` API, so normalizing in that direction is the behaviour a consumer should get for free.

TauCeti/Combinatorics/DenseGraphLimits/Kernel/CutNorm.lean
kbuzzard, line 1: ### ✅ documentation — now passing on `cc4c756`.

cameronfreer: Contesting: the counterexample is valid, and both grounds given for rejecting it are incorrect. I checked by compiling it rather than by re-reading the docstring.

**"the constant kernel `1` is not a `SymmKernel` on Lebesgue space".** `SymmKernel` requires uniform boundedness, not integrability — `Kernel/Basic.lean`:

```lean
  bdd' : ∃ C, ∀ x y, |toFun x y| ≤ C
```

so the constant kernel is admissible, with `C = 1`. It compiles:

```lean
noncomputable def one : SymmKernel ℝ volume where
  toFun _ _ := 1
  symm' _ _ := rfl
  meas' := measurable_const
  bdd' := ⟨1, fun _ _ => by norm_num⟩
```

**"its claim of unbounded test integrals contradicts the kernel's built-in integrability".** There is no built-in integrability. The kernel's integrability is a *theorem carrying the very hypothesis under discussion* — `CutNorm.lean:74`:

```lean
theorem integrable_uncurry [IsFiniteMeasure μ] (K : SymmKernel Ω μ) : …
```

On `(ℝ, volume)` the constant kernel is genuinely not integrable on the product, which is exactly why the supremum is unbounded rather than in spite of it.

**The unboundedness, verified.** Taking `u = v = (Icc 0 n).indicator 1` — measurable and `[-1,1]`-valued, both checked — the file's own `testIntegral_indicator_one` gives `testIntegral = rectIntegral (Icc 0 n) (Icc 0 n)`, and

```lean
example (n : ℕ) : one.rectIntegral volume (Icc (0:ℝ) n) (Icc (0:ℝ) n) = (n : ℝ) ^ 2
```

compiles. So the range contains `{n² : n ∈ ℕ}` and is unbounded, and `⨆` on `ℝ` returns `0`.

That is the whole content of the claim the docstring makes: without `[IsFiniteMeasure μ]` the definition is not a faithful supremum. It is not merely an API/domain restriction — it is what makes `abs_testIntegral_le_integral_abs` bound the range. Weakening the docstring to "domain restriction" would remove the reason this hypothesis exists, which the round-3 `correctness` block on this same PR required be there.

Happy to be shown wrong — if there is a reading under which `bdd'` implies integrability on an infinite measure, I will restate the docstring immediately. Otherwise I would keep the counterexample.

kbuzzard: **Re: your reply on `documentation` —** re-reviewed on `cc4c756`; this clears the finding ✅ — approved on `cc4c756`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3818291480.</sub>

PR #3916, opened 2026-08-20
feat: pin a mathlib release tag before stepping over it

kim-em:
This PR teaches the daily bump to stop at a mathlib release tag on its way past. Mathlib's `vX` tag sits on the commit that bumps its own `lean-toolchain` to X, so the window in which main could pin a release is one mathlib toolchain era, and for a stable release that has been as little as fifteen hours. Taking the last-known-good commit every day steps clean over such a window, and did over v4.33.0. When the target lies past a release tag main has not reached, this pins that tag first on its own branch and holds the ordinary bump back until it merges, so the exact release commit exists on main rather than having to be constructed afterwards.

A stepping stone must never be able to freeze dependency updates, so it has a hard lifetime rather than a CI-colour test: at the next daily tick it has either merged or it is demoted, whatever its colour, and demotion parks its head on `releases/<tag>` where the release can still be built and tagged. Twenty hours rather than twenty-four, because this job runs at 17:00 and a stone opened at 17:01 is 23.98 hours old at the next tick, so a 24 hour threshold would silently become 48. A hold label buys one further tick and raises its own alert; parking a stone parks every dependency update behind it, and that is never a quiet default. A first-known-bad regression outranks a stone entirely.

Three settings carry it. `preserve-lakefile` is mandatory: without it hopscotch writes the tag name into the manifest's `inputRev` and `check-bump.sh` rejects a changed nomination outright, so the PR could never auto-merge. `skip-build` is what makes the lifetime policy coherent, because with a build hopscotch reports `updated=false` on failure, the PR is never opened, and the bump sails over the release in exactly the case this exists for. `apply-fixes` stays false because applied fixes are changes to `TauCeti/`, which would break both auto-merge and the two-pin-files-only property the release workflow relies on.

The state lives in GitHub rather than in a store of its own: an open PR on the stone branch means one is in flight, its `created_at` is a stable clock precisely because the branch is never force-pushed while it is open, and a tag or a `releases/` branch records that a release has been handled. `stuck_alerts.py` gains a `stuck-stone` detector that fires before the closer acts, because `stranded-pr` skips any PR carrying a hold label and wants a green build plus a review verdict, which is exactly the shape a parked stone has.

This can only ever propose a tag reachable from the last-known-good commit, so mathlib's stable-branch patch releases are backfill-only by construction; `check-bump.sh` would refuse them anyway.

Roadmap: none

🤖 Prepared with Claude Code


kbuzzard:
🔍 Review in progress — `claude` reviewing `680f96ee21e3`.

.github/workflows/update.yml
kbuzzard, line 1: ### ⛔ correctness — block  `claude/claude-opus-5`

The whole mechanism is justified by, and instructs humans to use, a `release-tag.yml` workflow and a `Releases` Zulip topic that do not exist anywhere in the repo, so the stone's payoff never materialises while its cost (a frozen bump day per mathlib release) is real. Separately, the new `!cancelled()` guards silently drop the implicit `success()` on the `fkb` step, so a failed incompatibility query is now read as "no incompatibility".

- `.github/workflows/update.yml:19` — The stated purpose of the stone, the PR close comment (l.229, 243), the stone PR body, the demotion Zulip message (l.499-502) and the alert body all tell a human to dispatch `release-tag.yml` against the parked head. No such workflow exists: `grep -rn "release-tag"` matches only update.yml, and `find .github -type f` lists no release/tagging workflow (nor does `grep -rln "git tag\|refs/tags" .github/ scripts/`). The consumer this feature exists to serve is assumed rather than present, so the delivered behaviour is a day of suppressed dependency bumps per mathlib release plus a `releases/*` branch nothing reads. _Fix:_ Land the release/tag workflow first as its own PR, or rewrite every reference (comments, PR body, close comment, Zulip text) to describe the actual manual procedure that exists today.
- `.github/workflows/update.yml:463` — The comment asserts `Releases` is "the same topic release-tag.yml and the toolchain-tag audit report to". Neither exists: the only `ZULIP_TOPIC` values in the repo are `Nightly verification` (nightly-verify.yml:335) and `Stuck PRs` (stuck-alerts.yml:59), and there is no toolchain-tag audit. _Fix:_ Drop the false provenance claim; state plainly that this introduces a new `Releases` topic.
- `.github/workflows/update.yml:409` — `!cancelled()` is a status-check function, so it replaces the implicit `success()` that previously guarded these steps. The `fkb` step has no `continue-on-error`, and its output is empty when it fails, so a failed first-known-bad query now satisfies `steps.fkb.outputs.commit == ''` and the ordinary bump runs as if no incompatibility were reported. Before this PR the same failure skipped the bump. Same defect at l.426 for `Open or update PR`. _Fix:_ Add `&& steps.fkb.outcome == 'success'` to both conditions (l.409-416 and l.426-434), so only a successful query saying "no regression" releases the bump.
- `.github/workflows/update.yml:392` — The restore step's comment states the invariant "whatever the stone left behind, the ordinary bump below must start from the tree this run was scheduled against", but its condition (`stone.outcome == 'failure' || (updated == 'true' && stone_pr.outcome != 'success')`) omits the case `stone.outcome == 'success' && updated != 'true'` — the stone action ran a `lake update` against the stone branch, reported no change, and the ordinary bump then runs on whatever tree/branch it left. That case occurs routinely (e.g. the day after a stone merges, or after a demotion whose parking failed left `handled=false`). _Fix:_ Condition the restore on the stone step having run at all and the bump being about to proceed: `steps.stone.outcome != 'skipped' && (steps.stone.outputs.updated != 'true' || steps.stone_pr.outcome != 'success')`.
- `.github/workflows/update.yml:503` — The "It carried no release marker, so its head was NOT parked" branch is unreachable: `parked` is set to `true` only inside `if [ -n "$RELEASE" ]` (l.196-224), so an unmarked stone always reports `PARKED=false` and is caught by the earlier `if os.environ.get("PARKED") == "false"` branch, which then renders "If `an unmarked release` is still wanted". _Fix:_ Test the marker first (`if not release: ... elif parked == 'false': ...`), or drop the dead branch and make the parked-failure text handle a missing marker explicitly.

<sub>`claude/claude-opus-5` · 401s · 56 in / 27k out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/675a2d2c1de7656ed751bc9b856fe902fb056c15...680f96ee21e37c5f1d9df3f2e9be673ee739b609) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/correctness.md)</sub>

kbuzzard:
## AI review — blocked

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics).

|---|---|---|---|---|
| ⛔ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/correctness.md) | blocked | `claude/claude-opus-5` | The whole mechanism is justified by, and instructs humans to use, a `release-tag.yml` workflow and a `Releases` Zulip topic that do not exist anywhere in the repo, so the stone's payoff never materialises while its cost (a frozen bump day per mathlib release) is real. Separately, the new `!cancelled()` guards silently drop the implicit `success()` on the `fkb` step, so a failed incompatibility query is now read as "no incompatibility". |
| ▫️ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/reuse.md) | not yet run | `—` |  |
| ▫️ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/scope.md) | not yet run | `—` |  |
| ▫️ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/attribution.md) | not yet run | `—` |  |
| ▫️ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/api-design.md) | not yet run | `—` |  |
| ▫️ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/generality.md) | not yet run | `—` |  |
| ▫️ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/placement.md) | not yet run | `—` |  |
| ▫️ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/naming.md) | not yet run | `—` |  |
| ▫️ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/documentation.md) | not yet run | `—` |  |
| ▫️ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics/proof-quality.md) | not yet run | `—` |  |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/675a2d2c1de7656ed751bc9b856fe902fb056c15...680f96ee21e37c5f1d9df3f2e9be673ee739b609) at head `680f96e`. rubrics @ [`85349e8`](https://github.com/TauCetiProject/TauCetiReview/tree/85349e8225cebb5fb9f1877b31bb8555303ade84/rubrics). Review spend: $2.51.</sub>

kim-em:
Closing this unmerged, by the repository owner's decision.

This PR taught the daily bump to stop at a mathlib release tag on its way past, so the exact release commit would land on `main` by itself rather than being constructed afterwards. That is an optimisation of a future case, and it was never part of what was asked for; what it costs is intricacy in `update.yml`, which is the only thing keeping this project's mathlib pin moving.

Across four review rounds it produced a new way to freeze that pin every single time: a diagnostic written into `$GITHUB_OUTPUT`, a demotion run that did not resume bumping, a failed parking that left the PR open for ever, a failed close that never emitted `demoted`, and finally the implicit `success()` guard suppressing the bump behind any optional step. Each fix was correct and each round found another. That is a sign the design is too intricate for the value it adds, not that the next fix would have been the last one.

The branch is left in place for reference.

🤖 Prepared with Claude Code

PR #4001, opened 2026-08-20
feat: add smooth fibrewise Riesz duals

sqrt-of-2:
This PR supplies the fibrewise Riesz regularity step needed to construct the Levi–Civita connection from the Koszul functional. It defines the canonical Riesz representative of a tangent-fibre covector, proves its characteristic lemmas (linearity is inherited from Mathlib's `LinearIsometryEquiv` API), derives its inverse-Gram coordinate formula in the chart-local frame, and proves chartwise and pointwise `C^n` smoothness criteria for a varying field of covectors.

The exact roadmap target is `TauCetiRoadmap/HopfRinow/README.md`, Layer 1, **Regularity of the Levi-Civita connection**: prove the resulting Levi–Civita connection smooth and expose smooth chart-level Christoffel symbols. The designated milestone is the regularity input for the later geodesic-spray milestone. This is the next critical-path step because the Koszul formula produces a fibrewise covector, while the connection construction needs its metric-dual vector to vary smoothly.

After this PR, Layer 1 still needs the Levi–Civita connection constructed from the Koszul functional and the resulting connection and Christoffel-symbol maps proved smooth; the Riesz conversion required by that argument is supplied here.

The coordinate formula and smoothness proof are adapted from stages 6–7 of Poincare-Conjecture's Apache-2.0 `DoCarmoLib/Riemannian/TensorBundle/MusicalIso.lean` at revision `24f32e4d600878bfaac6bc2f2f9324175571c321`. The implementation reuses Mathlib's `InnerProductSpace.toDual`, finite-dimensional tangent-fibre instances, continuous-linear-map API, and Tau Ceti's chart-local frames and smooth inverse Gram matrix; no Mathlib implementation is vendored.

The change adds `Riemannian/Riesz.lean` (222 lines) and adds the six-line `chartGramMatrix_apply` accessor lemma to `Riemannian/ChartGram.lean`.

Roadmap: HopfRinow

🤖 Prepared with Codex

PR #4006, opened 2026-08-20
feat: the inverse function theorem for manifolds

sqrt-of-2:
This PR proves the inverse function theorem for manifolds: a map which is `C^n` on an open set, with `1 ≤ n`, and whose `mfderiv` at a point of that set is a continuous linear equivalence is a `C^n` local diffeomorphism at that point. Mathlib knows the forward implication (`IsLocalDiffeomorphAt.mfderivToContinuousLinearEquiv`) and lists this converse as a TODO in `Mathlib/Geometry/Manifold/LocalDiffeomorph.lean`, so the roadmap owns it rather than consuming it.

The exact roadmap target is `TauCetiRoadmap/HopfRinow/README.md`, Layer 1, the bullet **"The manifold inverse-function theorem"**: "add the missing shared theorem to `TauCeti/Geometry/Manifold/LocalDiffeomorph.lean`: for `C¹` boundaryless Banach manifolds, a `C¹` map whose `mfderiv` at a point is a continuous linear equivalence induces a `LocalDiffeomorphAt` there." The milestone it serves is the Layer-1 exponential-map milestone whose last step reads "Apply it to the preceding derivative theorem to obtain that `exp_p` is a local diffeomorphism at `0`", which in turn feeds Layer 2's normal neighbourhoods and local logarithms and thence assertion (f) of Hopf–Rinow. This was the most effective step available: it is the only Layer-1 item that is fully independent of the Levi-Civita/geodesic-spray chain (the connection existence and its regularity are in flight in #4001), it is stated in the roadmap as a self-contained prerequisite with its own file, and no open or merged PR covers it.

`TauCeti/Analysis/Calculus/InverseFunctionTheorem.lean` (86 lines) supplies the normed-space input. Mathlib's `ContDiffAt.toOpenPartialHomeomorph` produces an `OpenPartialHomeomorph`, but `ContDiffAt.to_localInverse` only makes the inverse `C^n` *at the image point*, which is not enough to build a `PartialDiffeomorph`. Restricting the source to the open set where the derivative stays invertible — invertibility is open by `ContinuousLinearEquiv.isOpen`, and `ContDiffOn.continuousOn_fderiv_of_isOpen` gives continuity of `fderiv` — makes `OpenPartialHomeomorph.contDiffAt_symm` applicable at every point of the target, giving `TauCeti.ContDiffOn.exists_openPartialHomeomorph`.

`TauCeti/Geometry/Manifold/LocalDiffeomorph.lean` (246 lines) transports this to manifolds. It records an extended chart of a boundaryless manifold as a partial diffeomorphism (`extChartPartialDiffeomorph`), an open partial homeomorphism between model spaces that is `C^n` both ways as a partial diffeomorphism (`PartialDiffeomorph.ofOpenPartialHomeomorph`), the source of a composite (`PartialDiffeomorph.trans_source`), and the fact that a map agreeing with a partial diffeomorphism on its source is a local diffeomorphism there (`isLocalDiffeomorphAt_of_eqOn`). The main theorem `isLocalDiffeomorphAt_of_mfderiv_eq` reads `f` in the extended charts at `x` and `f x` via `contMDiffOn_iff`, identifies `mfderiv` with the `fderiv` of that chart representative using `MDifferentiableAt.mfderiv` and `ModelWithCorners.range_eq_univ`, applies the normed-space theorem, and conjugates back by the two charts. Two corollaries follow: `isLocalDiffeomorphAt_iff_exists_mfderiv_eq`, which closes the characterisation together with Mathlib's forward direction, and `isLocalDiffeomorph_of_mfderiv_eq`, the global version. As an acceptance check that the manifold statement really specializes correctly, `isLocalDiffeomorphAt_of_fderiv_eq` recovers the classical inverse function theorem on normed spaces viewed as manifolds over `𝓘(𝕂, E)`.

Hypotheses are spelled out as the roadmap's standing conventions ask: `[RCLike 𝕂]`, `[CompleteSpace E]`, `[I.Boundaryless]`, `[J.Boundaryless]`, `[IsManifold I n M]`, `[IsManifold J n N]`, with no bundled abbreviation. Boundarylessness of both models is what makes the extended charts have open targets and turns `fderivWithin _ (range I)` into `fderiv`.

No Mathlib source was vendored; the proof consumes `ContDiffAt.toOpenPartialHomeomorph`, `OpenPartialHomeomorph.contDiffAt_symm`, `OpenPartialHomeomorph.restrOpen`, `ContinuousLinearEquiv.isOpen`, `contMDiffOn_extChartAt`, `contMDiffOn_extChartAt_symm`, `contMDiffOn_iff`, `contMDiffOn_iff_contDiffOn`, and `PartialDiffeomorph.trans`. The prior-art repository named in the roadmap's provenance section was not used.

After this PR, the exponential-map milestone still needs everything upstream of it — the Levi-Civita connection and its regularity, the covariant derivative along a curve, the geodesic spray and its maximal intervals, and `exp_p` itself together with the identification of `mfderiv exp_p 0` with the identity — before this theorem can be applied at `0` to obtain the normal neighbourhoods of Layer 2.

Roadmap: HopfRinow

🤖 Prepared with Claude Code

### DIFF
--- a/TauCeti/Geometry/Manifold/LocalDiffeomorph.lean
+++ b/TauCeti/Geometry/Manifold/LocalDiffeomorph.lean
@@ -1,0 +1,246 @@
+/-
+Copyright (c) 2026 The Tau Ceti contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: The Tau Ceti contributors
+-/
+module
+
+public import Mathlib.Geometry.Manifold.LocalDiffeomorph
+public import TauCeti.Analysis.Calculus.InverseFunctionTheorem
+
+/-!
+# The inverse function theorem for manifolds
+
+Mathlib knows that a `C^n` local diffeomorphism has invertible differentials
+(`IsLocalDiffeomorphAt.mfderivToContinuousLinearEquiv`) and lists the converse as a TODO in
+`Mathlib/Geometry/Manifold/LocalDiffeomorph.lean`. This file proves that converse for boundaryless
+Banach manifolds: a map which is `C^n` on an open set, with `1 ≤ n`, and whose `mfderiv` at a point
+of that set is a continuous linear equivalence, is a `C^n` local diffeomorphism at that point.
+
+The proof reads `f` in the extended charts at `x` and at `f x`, applies the normed-space inverse
+function theorem of `TauCeti.ContDiffOn.exists_openPartialHomeomorph`, and conjugates the resulting
+`OpenPartialHomeomorph` back by the two extended charts, each of which is itself a partial
+diffeomorphism because the models are boundaryless.
+
+## Main results
+
+* `TauCeti.extChartPartialDiffeomorph`: an extended chart of a boundaryless manifold, as a partial
+  diffeomorphism onto its open image in the model space.
+* `TauCeti.PartialDiffeomorph.ofOpenPartialHomeomorph`: an open partial homeomorphism between
+  model spaces which is `C^n` in both directions, as a partial diffeomorphism.
+* `TauCeti.isLocalDiffeomorphAt_of_mfderiv_eq`: the inverse function theorem for manifolds.
+* `TauCeti.isLocalDiffeomorphAt_iff_exists_mfderiv_eq`: the resulting characterisation of
+  `IsLocalDiffeomorphAt` for a map which is `C^n` on an open set.
+* `TauCeti.isLocalDiffeomorphAt_of_fderiv_eq`: its specialization to normed spaces, which is the
+  classical inverse function theorem.
+* `TauCeti.isLocalDiffeomorphAt_of_eqOn`: a map agreeing with a partial diffeomorphism on its
+  source is a local diffeomorphism there.
+* `TauCeti.isLocalDiffeomorph_of_mfderiv_eq`: the global version.
+
+## References
+
+* [The Hopf--Rinow roadmap](https://github.com/TauCetiProject/TauCetiRoadmap/blob/main/TauCetiRoadmap/HopfRinow/README.md),
+  Layer 1, "The manifold inverse-function theorem".
+-/
+
+public section
+
+noncomputable section
+
+open Set
+open scoped Manifold
+
+namespace TauCeti
+
+variable {𝕂 : Type*} [RCLike 𝕂]
+  {E : Type*} [NormedAddCommGroup E] [NormedSpace 𝕂 E]
+  {F : Type*} [NormedAddCommGroup F] [NormedSpace 𝕂 F]
+  {H : Type*} [TopologicalSpace H] {G : Type*} [TopologicalSpace G]
+  {M : Type*} [TopologicalSpace M] [ChartedSpace H M]
+  {N : Type*} [TopologicalSpace N] [ChartedSpace G N]
+
+section Charts
+
+variable (I : ModelWithCorners 𝕂 E H) [I.Boundaryless] (n : WithTop ℕ∞) [IsManifold I n M]
+
+/-- The extended chart at `x`, as a partial diffeomorphism from `M` to the model space `E`. Its
+target is open because `I` is boundaryless. -/
+@[expose]
+def extChartPartialDiffeomorph (x : M) : PartialDiffeomorph I 𝓘(𝕂, E) M E n where
+  toPartialEquiv := extChartAt I x
+  open_source := isOpen_extChartAt_source x
+  open_target := isOpen_extChartAt_target x
+  contMDiffOn_toFun := by
+    simpa only [extChartAt_source] using contMDiffOn_extChartAt (I := I) (n := n) (x := x)
+  contMDiffOn_invFun := contMDiffOn_extChartAt_symm x
+
+@[simp]
+theorem extChartPartialDiffeomorph_toPartialEquiv (x : M) :
+    (extChartPartialDiffeomorph I n x).toPartialEquiv = extChartAt I x := rfl
+
+@[simp]
+theorem coe_extChartPartialDiffeomorph (x : M) :
+    ⇑(extChartPartialDiffeomorph I n x) = extChartAt I x := rfl
+
+end Charts
+
+/-- An open partial homeomorphism between model spaces which is `C^n` in both directions is a
+partial diffeomorphism. -/
+@[expose]
+def PartialDiffeomorph.ofOpenPartialHomeomorph {n : WithTop ℕ∞} (Θ : OpenPartialHomeomorph E F)
+    (hΘ : ContDiffOn 𝕂 n Θ Θ.source) (hΘsymm : ContDiffOn 𝕂 n Θ.symm Θ.target) :
+    PartialDiffeomorph 𝓘(𝕂, E) 𝓘(𝕂, F) E F n where
+  toPartialEquiv := Θ.toPartialEquiv
+  open_source := Θ.open_source
+  open_target := Θ.open_target
+  contMDiffOn_toFun := contMDiffOn_iff_contDiffOn.2 hΘ
+  contMDiffOn_invFun := contMDiffOn_iff_contDiffOn.2 hΘsymm
+
+@[simp]
+theorem PartialDiffeomorph.ofOpenPartialHomeomorph_toPartialEquiv {n : WithTop ℕ∞}
+    (Θ : OpenPartialHomeomorph E F) (hΘ : ContDiffOn 𝕂 n Θ Θ.source)
+    (hΘsymm : ContDiffOn 𝕂 n Θ.symm Θ.target) :
+    (PartialDiffeomorph.ofOpenPartialHomeomorph Θ hΘ hΘsymm).toPartialEquiv =
+      Θ.toPartialEquiv := rfl
+
+section Trans
+
+variable {I : ModelWithCorners 𝕂 E H} {J : ModelWithCorners 𝕂 F G}
+  {F' : Type*} [NormedAddCommGroup F'] [NormedSpace 𝕂 F'] {H' : Type*} [TopologicalSpace H']
+  {K : ModelWithCorners 𝕂 F' H'} {P : Type*} [TopologicalSpace P] [ChartedSpace H' P]
+  {n : WithTop ℕ∞}
+
+/-- The source of a composition of partial diffeomorphisms. -/
+theorem PartialDiffeomorph.trans_source (Φ : PartialDiffeomorph I J M N n)
+    (Ψ : PartialDiffeomorph J K N P n) :
+    (Φ.trans Ψ).source = Φ.source ∩ Φ ⁻¹' Ψ.source := by
+  rw [PartialDiffeomorph.trans_toPartialEquiv]
+  exact OpenPartialHomeomorph.trans_source _ _
+
+end Trans
+
+section EqOn
+
+variable {I : ModelWithCorners 𝕂 E H} {J : ModelWithCorners 𝕂 F G} {n : WithTop ℕ∞}
+
+/-- A map agreeing with a partial diffeomorphism on its source is a `C^n` local diffeomorphism at
+every point of that source. -/
+theorem isLocalDiffeomorphAt_of_eqOn {Φ : PartialDiffeomorph I J M N n} {f : M → N} {x : M}
+    (hx : x ∈ Φ.source) (hf : EqOn f Φ Φ.source) : IsLocalDiffeomorphAt I J n f x :=
+  PartialDiffeomorph.isLocalDiffeomorphAt I J n
+    ({ toPartialEquiv :=
+        { toFun := f
+          invFun := Φ.toPartialEquiv.symm
+          source := Φ.source
+          target := Φ.target
+          map_source' := fun _ hy => hf hy ▸ Φ.toPartialEquiv.map_source hy
+          map_target' := fun _ hw => Φ.toPartialEquiv.map_target hw
+          left_inv' := fun _ hy => hf hy ▸ Φ.toPartialEquiv.left_inv hy
+          right_inv' := fun _ hw =>
+            (hf (Φ.toPartialEquiv.map_target hw)).trans (Φ.toPartialEquiv.right_inv hw) }
+       open_source := Φ.open_source
+       open_target := Φ.open_target
+       contMDiffOn_toFun := Φ.contMDiffOn_toFun.congr fun _ hy => hf hy
+       contMDiffOn_invFun := Φ.contMDiffOn_invFun } : PartialDiffeomorph I J M N n) hx
+
+end EqOn
+
+section InverseFunctionTheorem
+
+variable [CompleteSpace E] {I : ModelWithCorners 𝕂 E H} [I.Boundaryless]
+  {J : ModelWithCorners 𝕂 F G} [J.Boundaryless] {n : WithTop ℕ∞}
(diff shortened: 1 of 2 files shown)


CBirkbeck:
🔍 Review in progress — `codex` reviewing `28d4cbdc6be7`.

TauCeti/Geometry/Manifold/LocalDiffeomorph.lean
CBirkbeck, line 1: ### ✅ reuse — now passing on `0697ad7`.

CBirkbeck, line 1: ### ✅ api-design — now passing on `0697ad7`.

CBirkbeck, line 1: ### ✅ generality — now passing on `0697ad7`.

CBirkbeck, line 1: ### ✅ proof-quality — now passing on `0697ad7`.

CBirkbeck:
## AI review — approved

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics).

|---|---|---|---|---|
| ✅ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics/correctness.md) | approved | `codex/gpt-5.6-sol` | The new normed-space and manifold statements faithfully express the inverse function theorem: smoothness and genuine derivative invertibility yield an actual local diffeomorphism, without vacuous predicates or assumed conclusions. |
| ✅ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics/reuse.md) | approved | `codex/gpt-5.6-sol` | The current PR adds genuinely new inverse-function material and reuses the relevant Mathlib APIs. The contested `isLocalDiffeomorphAt_of_eqOn` supplies a missing public introduction mechanism because Mathlib's defining existential is opaque outside its module. |
| ✅ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics/scope.md) | approved | `codex/gpt-5.6-sol` | The PR directly implements the Hopf–Rinow Layer 1 manifold inverse-function theorem target. The normed-space strengthening and manifold transport are necessary parts of one coherent unit. |
| ✅ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics/attribution.md) | approved | `codex/gpt-5.6-sol` | The code credits the roadmap target and identifies the central Mathlib theorem and TODO it extends; no copied or closely adapted external formalization requiring further attribution was found. |
| ✅ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics/api-design.md) | approved | `codex/gpt-5.6-sol` | The canonical public surface is appropriately scoped and characterized; the previously exposed definition bodies are now sealed, with projection/coercion lemmas retained for consumers. |
| ✅ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics/generality.md) | approved | `codex/gpt-5.6-sol` | The open-set hypothesis is the natural uniform formulation: `ContMDiffAt` does not provide a common smooth neighborhood at `n = ∞`, while the local-diffeomorphism conclusion does. No material generality issue remains. |
| ✅ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics/placement.md) | approved | `codex/gpt-5.6-sol` | The declarations are in their canonical calculus and manifold local-diffeomorphism modules. The imports are topic-specific and directly support the implementations. |
| ✅ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics/naming.md) | approved | `codex/gpt-5.6-sol` | The introduced names accurately describe their conclusions using established Mathlib terminology and adjacent naming patterns; no notation is introduced. |
| ✅ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics/documentation.md) | approved | `codex/gpt-5.6-sol` | Both substantive files have accurate module docstrings, and their public definitions and main theorems are documented clearly without overclaiming. |
| ✅ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics/proof-quality.md) | approved | `codex/gpt-5.6-sol` | The prior definitional-equality concern is resolved: the chart representative and derivative identifications now use explicit, documented intermediate statements and controlled rewriting. The remaining proofs are structured, readable, and appropriately commented. |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/8b876313601c813bcd2435b537d07a2b66ff9b5e...0697ad7bafc3e8713a307fd4a433c31c9be1fa68) at head `0697ad7`. rubrics @ [`f854a7e`](https://github.com/TauCetiProject/TauCetiReview/tree/f854a7e1bf95d7182f672ea06e690d08b07e06a5/rubrics). Review spend: $4.03.</sub>

TauCeti/Geometry/Manifold/LocalDiffeomorph.lean
sqrt-of-2: Bullets 2 and 3 are fixed in `0697ad7b`; bullet 1 is contested — its suggested replacement does not compile.

**Bullet 2 (`PartialDiffeomorph.trans_source`) — fixed.** The wrapper is deleted. The proof now names the two composites and derives each source inline from the generic lemma, exactly as suggested:

```lean
have hΨsource : Ψ.source = Θ.source ∩ Θ ⁻¹' ψ.target := by
  rw [hΨ, PartialDiffeomorph.trans_toPartialEquiv]
  exact OpenPartialHomeomorph.trans_source _ _
have hsource : Φ.source = φ.source ∩ φ ⁻¹' (Θ.source ∩ Θ ⁻¹' ψ.target) := by
  rw [hΦ, PartialDiffeomorph.trans_toPartialEquiv, ← hΨsource]
  exact OpenPartialHomeomorph.trans_source _ _
```

(One detail worth recording for a future round: the literal recipe — one `rw [PartialDiffeomorph.trans_toPartialEquiv]` on the *nested* composite — fails, because after the outer rewrite the inner composite sits under `.toOpenPartialHomeomorph`, and the second rewrite reports `Did not find an occurrence of the pattern (PartialDiffeomorph.trans ?Φ ?Ψ).toPartialEquiv`. Splitting the composite in two, as above, is what makes the generic lemma apply twice.)

**Bullet 3 (`isLocalDiffeomorphAt_of_fderiv_eq`) — fixed.** Deleted, together with its bullet in the module docstring.

**Bullet 1 (`isLocalDiffeomorphAt_of_eqOn`) — contested.** `IsLocalDiffeomorphAt` is *not* `@[expose]`d in Mathlib (`Mathlib/Geometry/Manifold/LocalDiffeomorph.lean:156`: a bare `def ... : Prop := ∃ Φ, x ∈ Φ.source ∧ EqOn f ↑Φ Φ.source`, while its siblings `IsLocalDiffeomorphOn` and `IsLocalDiffeomorph` at lines 253/258 *are* `@[expose] def`). So from another module its body does not unfold and the anonymous constructor is unavailable. Replacing line 203 with `refine ⟨Φ, hxΦ, fun y hy => ?_⟩` as the fix asks gives:

```
error: TauCeti/Geometry/Manifold/LocalDiffeomorph.lean:177:9: Invalid `⟨...⟩` notation:
  The expected type `IsLocalDiffeomorphAt I J n f x` is not an inductive type
Note: This notation can only be used when the expected type is an inductive type with a
single constructor
```

That also answers "merely restates the definition": downstream the definition is inaccessible, and the *only* public introduction rule Mathlib offers is `PartialDiffeomorph.isLocalDiffeomorphAt`, which demands a `PartialDiffeomorph` whose `toFun` is literally `f`. The composite built in the main proof has `toFun = ψ.symm ∘ Θ ∘ φ`, which is only *equal* to `f` on the source — so the structure has to be re-formed with `toFun := f` (all four `PartialEquiv` fields transported along `hf`, and `contMDiffOn_toFun` by `ContMDiffOn.congr`). That is the content of `isLocalDiffeomorphAt_of_eqOn`, it is not available from any existing declaration, and it is the general statement of a step Mathlib itself lacks.

sqrt-of-2: Contested: the proposed pointwise hypothesis is not a generalization, it is a *false* statement for `n = ∞`, which is precisely the case the roadmap consumer needs.

**1. The conclusion already entails `ContMDiffOn` on an open neighbourhood.** `IsLocalDiffeomorphAt I J n f x` is `∃ Φ : PartialDiffeomorph I J M N n, x ∈ Φ.source ∧ EqOn f ↑Φ Φ.source` (`Mathlib/Geometry/Manifold/LocalDiffeomorph.lean:156`), and a `PartialDiffeomorph` carries `open_source` together with `contMDiffOn_toFun : ContMDiffOn I J n ↑Φ Φ.source`. So from the conclusion one gets `ContMDiffOn I J n f Φ.source` with `Φ.source` an *open set containing `x`* — i.e. the theorem as stated is already at the exact strength of its own conclusion. Weakening the hypothesis to `ContMDiffAt I J n f x` can only be sound if `ContMDiffAt` at a point implied `ContMDiffOn` on some open neighbourhood.

**2. It does not, and Mathlib says so in the signature.** The relevant transfer lemma is

```
ContDiffAt.contDiffOn : ContDiffAt 𝕜 n f x → m ≤ n → (m = ↑⊤ → n = ⊤) → ∃ u ∈ 𝓝 x, ContDiffOn 𝕜 m f u
```

(`#check` output against the pinned Mathlib). The third hypothesis `(m = ↑⊤ → n = ⊤)` is exactly what blocks `m = n = ∞`: in `WithTop ℕ∞`, `∞ = ↑(⊤ : ℕ∞)` while `ω = ⊤`, so for `n = ∞` the side condition fails and no `C^∞` neighbourhood is produced. This is not an artifact: `contDiffAt_infty : ContDiffAt 𝕜 ∞ f x ↔ ∀ n : ℕ, ContDiffAt 𝕜 n f x` (`Analysis/Calculus/ContDiff/Defs.lean:943`, manifold version `contMDiffAt_infty` at `Geometry/Manifold/ContMDiff/Defs.lean:643`) — the witnessing neighbourhood is allowed to shrink with the order.

**3. Concrete counterexample to the requested statement.** Take `f : ℝ → ℝ`, `f = id + Σ_k ε_k φ_k`, where `φ_k` is a bump supported in a small interval around `1/k` which is of class exactly `C^k` and not `C^(k+1)`, and `ε_k → 0` fast enough. Then `f` is `C^m` on a neighbourhood of `0` for every finite `m` (any neighbourhood omitting `φ_1,…,φ_(m-1)` works), hence `ContDiffAt ℝ ∞ f 0` by `contDiffAt_infty`, and `fderiv ℝ f 0 = id` is a continuous linear equivalence. But every neighbourhood of `0` contains some `1/k`, so `f` is `C^∞` on *no* neighbourhood of `0`, and therefore `f` is not a `C^∞` local diffeomorphism at `0` by point 1. The `ContMDiffAt`-hypothesis version of `isLocalDiffeomorphAt_of_mfderiv_eq` at `n = ∞` is thus refuted (viewing `ℝ` over `𝓘(ℝ, ℝ)`). The same example refutes the analogous pointwise weakening of `ContDiffOn.exists_openPartialHomeomorph`: its conclusion asserts `ContDiffOn 𝕂 n ↑Θ.symm Θ.target` on the open target, and a `C^∞` inverse with nonvanishing derivative would force `f` to be `C^∞` near each `1/k` by the classical inverse function theorem.

**4. What *is* derivable, and why it is not the right primitive.** For finite `n` (and for `n = ω`) one can indeed obtain an open neighbourhood from `ContDiffAt.contDiffOn` and derive a pointwise corollary. But it can only be stated with that `(m = ↑⊤ → n = ⊤)` side condition dragged into the statement, so it is *not* uniform in `n` and it excludes `n = ∞` — while the roadmap target this PR serves ("apply it to obtain that `exp_p` is a local diffeomorphism at `0`", HopfRinow Layer 1) is a smooth-manifold statement in the `∞` regime, where the input really is smoothness on an open set. Callers in this idiom (`ContMDiffOn f s`, `IsOpen s`, `x ∈ s`) also lose nothing: `ContMDiffOn` on any neighbourhood `s` of `x` gives the hypotheses on `interior s` for free.

I have therefore left both statements in the on-open-set form. If a consumer later needs the finite-order pointwise corollary it can be added then, with the side condition stated explicitly.

sqrt-of-2: Fixed in `0697ad7b`. Both `@[expose]` annotations are removed, so neither structure body is downstream interface any more; the projection lemmas are proved with the sealed-body idiom `:= (rfl)` (693 other occurrences in `TauCeti/`), since a bare `rfl` now fails with `Not a definitional equality: the left-hand side ...` on all three of them.

One consequence worth flagging, since it looks like a regression but is forced by a linter: with `extChartPartialDiffeomorph` sealed, `simpNF` reports that `simp` can prove `coe_extChartPartialDiffeomorph` from `extChartPartialDiffeomorph_toPartialEquiv` (via `extChartAt`/`OpenPartialHomeomorph.extend`/`PartialEquiv.coe_trans`), so keeping it `@[simp]` fails the linter set CI enforces. The lemma is kept — it is the coercion lemma consumers want for `rw` — but without the `@[simp]` attribute. `extChartPartialDiffeomorph_toPartialEquiv` and `PartialDiffeomorph.ofOpenPartialHomeomorph_toPartialEquiv` remain `@[simp]`, and the main proof now goes through them (`hΨsource`/`hsource`) rather than through the bodies.

CBirkbeck:
🔍 Review in progress — `codex` reviewing `0697ad7bafc3`.

TauCeti/Geometry/Manifold/LocalDiffeomorph.lean
CBirkbeck: **Re: your reply on `reuse` —** re-reviewed on `0697ad7`; this clears the finding ✅ — approved on `0697ad7`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3828104714.</sub>

CBirkbeck: **Re: your reply on `api-design` —** re-reviewed on `0697ad7`; this clears the finding ✅ — approved on `0697ad7`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3828110425.</sub>

CBirkbeck: **Re: your reply on `generality` —** re-reviewed on `0697ad7`; this clears the finding ✅ — approved on `0697ad7`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3828109276.</sub>

PR #3135, opened 2026-08-13
fix: test marking avoidance against the covered squares

rudygt:
This PR corrects the marking-avoidance region of grid rectangles so that the fully blocked grid differential can square to zero. `GridRectangle.AvoidsMarkings` tested the markings against `interior`, the open box `cIoo left right ×ˢ cIoo bottom top` of grid points strictly inside the rectangle. Markings live in squares, and a rectangle covers the squares `cIco left right ×ˢ cIco bottom top`, each square named by its lower-left grid point, so the old test undercounted: a rectangle one square wide or tall covers marked squares its open interior cannot see. Following Ozsváth–Stipsicz–Szabó, *Grid Homology for Knots and Links*, §4.4, a rectangle counts for the fully blocked complex `GC̃` only when `r ∩ 𝕆 = r ∩ 𝕏 = ∅` as regions; the juxtaposition pairing proving `∂² = 0` is developed in §4.6.

The defect is fatal to Lane G.3, "The complexes and `∂² = 0`": with the interior test, `∂² ≠ 0` from grid size four onward, so the roadmap's square-zero milestone was unprovable. Explicitly, for `n = 4` take `O c = c`, `X = ![1, 3, 0, 2]`, `x = ![0, 2, 1, 3]`, `z = ![1, 2, 3, 0]`: exactly five pairs of empty, marking-avoiding rectangles lead from `x` through some intermediate state to `z` under the old test, and five is odd, so the `z`-coefficient of `∂ (∂ x)` is `1` in `ZMod 2`. Each of the five pairs involves a thin rectangle that covers a marked square while its open interior is empty. With the corrected squares test the coefficient sums vanish (checked exhaustively over every diagram and state pair for `n ≤ 4`, and by sampling at `n = 5`, against a direct simulation of these definitions).

[...]

PR #4569, opened 2026-08-25
fix: build the merge candidate rather than the pull request head

kim-em:
This PR makes `pr-build` overlay GitHub's merge candidate for a pull request instead of the pull request head, and check the trusted base out at that candidate's own first parent instead of at the moving branch tip.

Overlaying the head's `TauCeti/` onto a base that carries the target branch's Lake pins produces a tree that exists in no commit and that no merge would ever produce. A branch cut before a Mathlib bump therefore compiles pre-bump sources against post-bump Mathlib and fails on code it never touched. On 2026-08-25 the bump to mathlib 618f225 (https://github.com/TauCetiProject/TauCeti/pull/4420 `chore: bump mathlib to 618f225, fix breaking changes`) changed one line of `TauCeti/Analysis/Contour/Argument/Principle.lean`, and 39 of the 40 pull requests then failing failed on that line. Each needed main merged into it by hand before it would build again.

The trust model is unchanged. The candidate is synthesised by GitHub in this repository rather than pushed by a contributor, only its `TauCeti/` is overlaid, lakefile and `scripts/` and manifest and toolchain still come from the trusted base, and its code still executes only under landrun with no credentials in reach. Because the candidate lives here, a fork is no longer contacted at build time at all.

The candidate is resolved once to an immutable SHA and then checked on both sides, because the ref moves and, worse, persists: when a pull request goes conflicted the ref is left pointing at the last clean merge rather than disappearing. Such a stale candidate still names the current head as its second parent, so a head-only check would pass it and build against a base hundreds of commits old, reintroducing the same class of failure through a different door. Conflicted https://github.com/TauCetiProject/TauCeti/pull/3135 and https://github.com/TauCetiProject/TauCeti/pull/1556 currently carry stale candidates whose base parents are 230 and 1376 commits behind main, both with a matching head. `mergeable` excludes those; an ancestry check on the base parent is the backstop; and every lookup lands on a named outcome rather than aborting the step.

A pull request that does not merge cleanly now reports that it needs updating, rather than reporting that the changed files could not be determined and routing to a human.

A green `build` here means this head builds against that base and may enter the queue. It does not mean the head works against main as of now, and it never did at five merges an hour. Only the `merge_group` build establishes that, and it is the one that decides what lands.

`merge_group` is untouched: it already resolves an immutable `base_sha`, and its commit is already a real merge of the batch onto that base.

🤖 Prepared with Claude Code

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -158,40 +158,60 @@ jobs:
           set -uo pipefail
           repo='${{ github.repository }}'
 
-          # `mergeable` is computed asynchronously: null means "ask again", not "conflicted".
-          # `tostring` rather than `// "unknown"` on purpose — false is falsy to jq's `//`, so
-          # the alternative form would report every genuine conflict as "not computed yet".
-          mergeable=null
+          # Mergeability, the candidate, and both endpoints come from ONE response so they
+          # describe one recomputation. Reading `.mergeable` from one call and the merge ref
+          # from another lets a green mergeability be paired with a candidate resolved at a
+          # different moment.
+          #
+          # `tostring` rather than `// "unknown"` on purpose: false is falsy to jq's `//`, so
+          # the obvious form reports every genuine conflict as "not computed yet".
+          snap=""; mergeable=null
           for _ in 1 2 3 4 5 6; do
-            mergeable=$(gh api "repos/$repo/pulls/$NUM" --jq '.mergeable | tostring') || mergeable=null
-            [ "$mergeable" != "null" ] && break
+            snap=$(gh api "repos/$repo/pulls/$NUM" \
+              --jq '[(.mergeable|tostring), (.merge_commit_sha // ""), .head.sha, .base.ref] | join(" ")') || snap=""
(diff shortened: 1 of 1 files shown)


kim-em:
Closing in favour of a simpler approach, which will get its own pull request.

Everything in this branch defends against one property: GitHub's `refs/pull/N/merge` is mutable, and it persists stale when a pull request becomes conflicted rather than disappearing. The mergeability poll, the parent check, the ancestry check, the parent-count check and the retarget check are all consequences of that. So is the second commit here: because `base/` became the candidate's own parent, `check-bump.sh` could run from a commit behind the tip, and had to be re-pinned to the workflow checkout.

Constructing the merge locally instead — check the base out at the current tip, merge the head into it, build the result — removes that whole class rather than guarding it. Conflicts are detected by git rather than by polling an API, identity needs no verification because both sides are chosen rather than discovered, a tree that is built cannot be left over from a previous merge, and `base/` goes back to being the current tip, which makes the validator staleness fixed in 65a9b237 impossible instead of patched.

Left open for the record, since the analysis in the two commit messages is the reason the replacement looks the way it does.

PR #4571, opened 2026-08-25
fix: build the pull request merged into the base, not the head alone

kim-em:
This PR makes `pr-build` construct the merge it is meant to be testing, rather than building a pull request head on its own.

Overlaying the head's `TauCeti/` onto a base that carries the target branch's Lake pins produces a tree that exists in no commit and that no merge would ever produce. A branch cut before a Mathlib bump therefore compiles pre-bump sources against post-bump Mathlib and fails on code it never touched. On 2026-08-25 the bump to mathlib 618f225 (https://github.com/TauCetiProject/TauCeti/pull/4420 `chore: bump mathlib to 618f225, fix breaking changes`) changed one line of `TauCeti/Analysis/Contour/Argument/Principle.lean`, and 39 of the 40 pull requests then failing failed on that line, each needing main merged into it by hand before it would build again.

The target branch is resolved to a commit once, in the step that already resolves the head. The trusted config is checked out at that commit, the same commit is checked out again as the thing to merge into, and the head named by the event is merged into it. So both halves of the build refer to one base even though main moves every few minutes, and the sources built are ones git would produce.

The trust model is unchanged. Only `TauCeti/` and the root `TauCeti.lean` are overlaid from the merged tree; lakefile, `scripts/`, manifest and toolchain still come from the trusted base; and the merged sources still execute only under landrun, offline, with no credentials in reach. Nothing but git runs over the pull request's contents to produce the merge, and the head is fetched from this repository (`refs/pull/N/head` is maintained here for forks too) and its SHA checked against the event before use, so a fork is not contacted at build time at all.

Merging here rather than reading `refs/pull/N/merge` is deliberate. That ref is mutable, and when a pull request conflicts it is left pointing at the last clean merge instead of disappearing: a stale candidate still names the pull request's current head as its second parent, so it passes the obvious check while carrying a base hundreds of commits old. Conflicted https://github.com/TauCetiProject/TauCeti/pull/3135 and https://github.com/TauCetiProject/TauCeti/pull/1556 currently carry such candidates, 230 and 1376 commits behind main. Using that ref safely needs a mergeability poll, a parent check, an ancestry check and a retarget check, all to re-establish facts that are simply known when the merge is made from two commits chosen here. It would also leave the trusted config checked out at the candidate's older parent, which would quietly run `check-bump.sh`, the anchor deciding what may be auto-merged, from a commit behind the tip.

[...]

PR #4617, opened 2026-08-25
fix: build a pull request against its own merge base's Lake pins

kim-em:
This PR takes the Lake pins a pull request is built against from its merge base rather than from the tip of the target branch, and drops the merge introduced in https://github.com/TauCetiProject/TauCeti/pull/4571 in favour of building the pull request's own tree.

Opened as a draft: it partly supersedes #4571, which merged earlier today, and the trade-off in the last section is a judgement call rather than a detail.

A pull request's sources are written against the pins its branch carries. Compiling them against whatever main carries right now is a state no commit is in and no merge would produce. That is what broke on 2026-08-25: the bump to Mathlib 618f225 (https://github.com/TauCetiProject/TauCeti/pull/4420 `chore: bump mathlib to 618f225, fix breaking changes`) needed one line of `TauCeti/Analysis/Contour/Argument/Principle.lean` changed with it, and every branch cut before that bump then failed on that line. Thirty-nine pull requests, none of which had touched it.

#4571 fixed that by merging the target branch into the pull request before building, which makes the sources consistent with the tip's pins. This fixes it from the other side, by leaving the sources alone and taking the pins the sources were written against. Both give a consistent tree; only this one keeps the build cache.

That second part is why it is worth revisiting. Lake artifacts are content-addressed over their whole transitive input, so moving Mathlib gives every module in the library a new hash and nothing already built can be reused. Those artifacts are published only by main's CI, once, after the bump lands, and that build is necessarily a cold one: 2h35m on the day, against a usual 15-25 minutes, and it was never queued, so that is simply what a cold build of this library costs. Pinning every open pull request to the tip means every pull request built during that window builds cold as well. About forty did, at roughly 75 minutes each against a usual twelve, which saturated all twenty runner slots on the free plan and left `Auto-merge` runs queued two hours deep. That is what took the merge rate to 1.8/h against a ~5/h baseline, rather than the bump itself.

Built at its own merge base, a pull request keeps hitting the cache that is already warm for it, and the cold rebuild belongs to main alone. The merge base does not need its own published revision: `lake cache get` backtracks through ancestors until it finds one, and on current main roughly one commit in two to four carries a mapping. What matters is that those ancestors are in the same pin regime as the pull request, which after a bump the tip's ancestors are not.

The pins are read from the merge base rather than from the pull request. For a pull request that does not touch them the two are identical, and the merge base is an ancestor of the target branch, so these are pins that were trusted on it rather than pins the pull request supplied. Everything else still comes from the tip: lakefile, `scripts/`, and the build tooling staged from `gate/`. A pull request that does touch the pins is a bump and is excluded, keeping the tip's config and overlaying its own validated pins as before, because a bump has to be judged against what main carries now or it could move Mathlib forward from a stale base and land something older than main already has. That also keeps `check-bump.sh` running from the tip.

`merge_group` is untouched, and still builds the queue's combined commit against the immutable base it was synthesised on. That remains the only build that decides what lands.

The trade-off: green here no longer predicts green in the merge queue. A pull request sitting at an old merge base can pass here and fail there. That is the right place for it to fail, but finding out costs a queue slot, and the queue admits two entries at a time. This is the same bargain Mathlib makes by building a pull request's own tree as-is.

Also dropped: the conflict detection #4571 added, which came from attempting the merge. Conflicted pull requests remain `merge-sweep`'s job, which already updates or flags them hourly.

🤖 Prepared with Claude Code

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -268,8 +268,8 @@ jobs:
           cp gate/scripts/source-modules.sh base/scripts/source-modules.sh
           cp gate/scripts/HeaderStyle.lean base/scripts/HeaderStyle.lean
 
-      - name: Checkout merge-base config (trusted ancestor — only to scope the PR's pin delta)
-        if: ${{ env.INFRA != '1' && env.BUMP == '1' }}
+      - name: Checkout merge-base config (trusted ancestor — the pins this PR is built against)
+        if: ${{ env.INFRA != '1' && github.event_name != 'merge_group' }}
         uses: actions/checkout@v4
         with:
           ref: ${{ steps.pr.outputs.mergebase }}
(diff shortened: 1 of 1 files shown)


kim-em:
Parking this: as drafted it does not deliver the cache benefit it exists for, and review is right that https://github.com/TauCetiProject/TauCeti/pull/4571 is the better gate.

The lookup root is the problem. `base/` is checked out at the branch tip, and `scripts/lake-cache-get.sh` runs `lake cache get` inside it with no `--rev`, so Lake reads the revision from that checkout's `HEAD`. Lake's revision scope is repository, toolchain and platform; it does not include `lake-manifest.json`. Copying the merge base's pins into `base/` therefore changes what is compiled but not what is looked up.

The consequence is worse than a missed optimisation. After a Mathlib-only bump the toolchain scope is unchanged, so once main publishes a mapping for the tip a stale pull request finds that mapping first, receives artifacts whose input hashes are against the new Mathlib, and cannot use them under the old manifest it was just given. Lake does not keep searching once it has found a mapping. So this appears to work during the publisher window, which is when it was reasoned about, and gets worse the moment publication happens.

Rooting the lookup at the merge base means checking `base/` out there rather than copying two files into it, with enough history to backtrack, which then raises whether `lakefile.toml` and `scripts/` should follow the pins to the merge base or stay on the tip. A toolchain bump can legitimately update lint executables or `HeaderStyle.lean` to syntax the older toolchain cannot parse, so that is a compatibility contract to design rather than a line to move.

Two further defects here regardless: merge-base resolution falls back to the tip on a compare failure, which reproduces exactly the stale-pin combination this work exists to remove and must instead fail closed; and the header still describes #4571's behaviour.

The underlying problem stays real. A bump invalidates every artifact, only main republishes, and that cold build took 2h35m against a usual 15-25 minutes, during which roughly forty pull requests each also built cold. Worth solving, but not like this.

kim-em:
Closing, per the note above: as drafted this does not deliver the cache benefit it exists for, the lookup root in `scripts/lake-cache-get.sh` is the actual problem, and #4571 followed by #4710 (`ci: build the exact pull request head`) covers the failure this was aimed at.

PR #4623, opened 2026-08-26
fix: validate a Lake-pin change even when the PR reaches outside TauCeti/

kim-em:
This PR makes a Lake-pin change get validated whether or not the pull request also reaches outside `TauCeti/`, so the required `bump-guard` status stops asserting a validation that never ran.

The scope guard asked two different questions in one `elif` chain: does this pull request reach outside `TauCeti/`, and does it change the Lake pins. A pull request doing both was only ever answered as out-of-scope, so `BUMP` stayed unset and `scripts/check-bump.sh` never ran. The report step then had nothing to distinguish that from a pull request with no pin change at all, and posted `bump-guard` green with the description "validated forward Lake pin bump (or no pin change)".

That is backwards for the one reader it matters to. Such a pull request is routed to a human precisely because it is out of scope, and the human deciding whether to merge it is told the pin change was validated when nothing validated it.

Asking the two questions separately fixes it at the source. A pin change is now always validated, so `bump-guard` always reports something that actually happened: green when the validator ran and passed, red when it ran and refused. Fixing it this way rather than by reporting the gap also means a pull request that legitimately needs an infrastructure change alongside a bump gets a truthful green and merges on its human review, instead of being blocked by a required check it would have to bypass.

The related case where the changed files could not be listed at all no longer claims "no pin change" either, since that is not known then.

Ordinary bump pull requests are unaffected: they touch only the pins and `TauCeti/`, so they already took the `BUMP` branch. The twelve most recent `chore: bump mathlib` pull requests all touch only in-scope paths.

Nothing else changes: not which pull requests route to a human, not the `scope` status, not what is built, except that a pin change accompanied by an out-of-scope path is now built against its own validated pins rather than the base's.

🤖 Prepared with Claude Code

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -173,6 +173,13 @@ jobs:
           if [ -z "$files" ]; then
             echo "INFRA=1" >> "$GITHUB_ENV"; echo "::warning::no files / unable to list — cannot determine the overlay, routing to human"
           else
+            # The file list was read, so whether the pins changed is now a known fact. Recorded
+            # separately because INFRA is set by seven different things here — an unresolved base,
+            # the merge-group file cap, a failed head fetch, a head that moved, a merge that did
+            # not apply — and only one of them is "could not list the files". Keying the bump
+            # report off INFRA would tell a PR that merely conflicts that its changed files could
+            # not be determined, which is both false and unhelpful.
+            echo "PINS_KNOWN=1" >> "$GITHUB_ENV"
             if echo "$files" | grep -vqE '^TauCeti/|^TauCeti\.lean$|^lake-manifest\.json$|^lean-toolchain$'; then
               if echo "$files" | grep -qE '^lakefile\.toml$'; then
                 echo "HUMAN_LAKEFILE=1" >> "$GITHUB_ENV"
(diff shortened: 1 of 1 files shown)


kim-em:
Reworked from the first version, which detected the gap and failed `bump-guard` closed to report it. That was the wrong shape: it left a pull request needing an infrastructure change alongside a bump unable to merge without bypassing a required check, and bump pull requests should not need that. Validating the pin change in the first place is both simpler and does not block anything.

kim-em:
Review found a further bypass in the pin detection, now fixed, and one adjacent issue I am deliberately leaving out of this pull request.

Fixed: comparing blob IDs alone missed a mode change. Git keeps the file mode beside the object and a symlink is a blob holding its target path, so `lean-toolchain` could be re-pointed at mode `120000` while reusing the existing blob. Its ID would be unchanged, the detection would conclude no pin change, `check-bump.sh` would not run, and main would end up with a dangling symlink where its toolchain should be. Whole tree entries are now compared, mode and type and object, and a pin path that is not an ordinary file at the head makes the state unknown rather than unchanged.

Also fixed: an unresolved merge base falls back to the base commit. That is exact for a merge group, whose head descends from its base, but a guess for a pull request, and a guess cannot answer whether the pull request changed the pins. It now makes the pin state unknown.

Not fixed here: `scope` and the lakefile flag still come from the changed-file list, so they keep the snapshot race that this pull request removed from the pin path. A branch that moves to a clean state while that list is read, and back before the final check, can have its scope decision describe a different commit than the one being statused. That is real and worth fixing, but it is the `scope` status rather than `bump-guard`, the file list also drives the diff-size and new-file guards, and deriving all of it from an immutable diff is a larger change that deserves its own review. Filing separately rather than growing this one.

PR #4647, opened 2026-08-26
refactor(Comodule): derive the additive-hom lemmas from the bundled hom

CBirkbeck:
`Comodule.Hom.toLinearMapAddMonoidHom` is the bundled `Hom R C M N →+ (M →ₗ[R] N)` whose coercion is the action on underlying linear maps. Every lemma saying an additive operation passes to `toLinearMap` is therefore a `map_*` of it, and every pointwise `_apply` lemma follows by `LinearMap.congr_fun`. Four proofs in `Comodule/Hom.lean` re-derived that content instead:

* `nsmul_toLinearMap` and `sum_toLinearMap` were `ext m; simp`; they are `map_nsmul` and `map_sum` of the bundled hom.
* `nsmul_apply` was an induction on `n`, and `sum_apply` a `classical` `Finset.induction_on`; both are one-line consequences of the corresponding `toLinearMap` lemma.

Each `_toLinearMap` lemma now precedes the `_apply` lemma it proves — the order the file already uses for the `zero`/`add`/`smul` group.

In `Comodule/Cat.lean` the six composition lemmas were re-proved by `ext` although `Comodule.Hom` already closes them. The category instance defines `comp f g` as `Comodule.Hom.comp g f` (`Cat.lean:89`), so each categorical lemma **is** the `Hom` lemma with its two arguments swapped — and the names swap with them: `ComoduleCat.add_comp` is `Comodule.Hom.comp_add`, `ComoduleCat.zero_comp` is `Comodule.Hom.comp_zero`, and so on for all six.

No statement changes anywhere, and the `@[simp]` set is unchanged.

### Scope

This is the Comodule half of dedup finding K05. Two things the finding listed need no change, and I left them alone rather than churning them:

* `Comodule/Finite/Preadditive.lean`'s `zero_apply` and `sub_apply` are already `rfl`.
* `ComoduleCat.toLinearMap_nsmul`, `toLinearMap_sum`, `nsmul_apply` and `sum_apply` already delegate to `Comodule.Hom`.

`Comodule.Hom.instAddCommMonoid` is also left as is. Rebuilding it through `Function.Injective.addCommMonoid` would require introducing an `SMul ℕ (Hom R C M N)` instance — an API addition, not a proof dedup, and one that would change the `nsmul` field's definitional behaviour that `nsmul_apply` and downstream files rely on.

[...]

PR #4672, opened 2026-08-26
refactor: bundle the complexification of integral linear maps as an additive hom

CBirkbeck:
Bundle the complexification of integral linear maps as an additive homomorphism, and read the
Hodge-morphism pointwise API off it, so that a single additivity fact replaces twelve hand-rolled
repetitions of it.

## What was duplicated

* In `Hodge/Conjugation.lean`, the six lemmas `integralMapToComplex_{zero,add,neg,sub,nsmul,zsmul}`
  each closed with the *identical* proof `h₁.algHom_ext _ _ fun x ↦ by simp` — six separate runs of
  the same base-change extensionality argument, when only the first two carry any content and the
  remaining four are formal consequences of additivity.
* In `Hodge/Morphism.lean`, the six `{zero,add,neg,sub,nsmul,zsmul}_apply` lemmas were each the
  same four-lemma `simp only` chain
  `[toLinearMap, ⟨op⟩_toIntLinearMap, integralMapToComplex_⟨op⟩, LinearMap.⟨op⟩_apply]`,
  differing only in which operation was substituted.

## What this adds

Two `private` additive homomorphisms, each built with `AddMonoidHom.mk'` so that the
extensionality argument is supplied exactly once, as additivity, and `map_zero` is derived rather
than re-proved:

* `integralMapToComplexAddMonoidHom : (V₁ →ₗ[ℤ] V₂) →+ (W₁ →ₗ[ℂ] W₂)` in `Conjugation.lean`. The
  six lemmas keep their statements and become `map_zero`, `map_add`, `map_neg`, `map_sub`,
  `map_nsmul`, `map_zsmul` of the bundle.
* `Hom.toLinearMapAddMonoidHom : Hom source target →+ (W₁ →ₗ[ℂ] W₂)` in `Morphism.lean`. The six
  `*_apply` lemmas become `LinearMap.congr_fun (map_⟨op⟩ toLinearMapAddMonoidHom …) x`.

Each bundle has a `private` `@[simp]` `coe_*` lemma, and all twelve derived proofs go through it by
`simpa` — so the step that bridges the bundle and the unbundled function is a named rewrite rather
than silent definitional unfolding.

No statement changes, and no use site changes: every existing lemma keeps its exact signature and
`@[simp]` status.

## Module-system surface

**Nothing is added to the public surface.** Both bundles and both `coe_*` lemmas are `private`;
they exist to organise the proofs of the twelve public lemmas in their own files, which remain the
[...]

PR #4677, opened 2026-08-26
refactor(Hodge): bundle the complexification of integral maps

CBirkbeck:
`integralMapToComplex h₁ ι₂` is additive, and its zero, addition, negation, subtraction, `ℕ`-multiple and `ℤ`-multiple laws were six separate theorems carrying the same proof — `h₁.algHom_ext _ _ fun x ↦ by simp` — written out each time.

Bundling it as `integralMapToComplexAddMonoidHom : (V₁ →ₗ[ℤ] V₂) →+ (W₁ →ₗ[ℂ] W₂)` leaves only additivity to check. The six laws then become `map_zero`, `map_add`, `map_neg`, `map_sub`, `map_nsmul` and `map_zsmul` of that map. Statements are unchanged and all six keep `@[simp]`.

Because those six are `@[simp]`, as are the `_toIntLinearMap` lemmas beside them, the six `Hom` evaluation lemmas in `Morphism.lean` no longer need their explicit four-lemma `simp only` chains and close with `simp [toLinearMap]`.

### Scope

This is the Hodge half of dedup finding K05; the Comodule half landed as #4647.

I deliberately did **not** bundle `Hom.toLinearMap` itself. Mirroring #4647 there would have meant three new public declarations (`toIntLinearMapAddMonoidHom`, `toLinearMapAddMonoidHom`, and a bridging `apply` lemma) to replace six already-short two-line proofs, and it fought the module system: with `integralMapToComplex`'s body sealed, `toLinearMapAddMonoidHom f` does not reduce to `f.toLinearMap`, so each evaluation lemma needed an extra bridging rewrite. Bundling pays when the duplicated proofs are *identical*, as in `Conjugation.lean`; here they were merely similar, and the collapse to `simp [toLinearMap]` gets the same reduction with no new API.

### Verification

`lake build` green on both changed modules and all five direct importers (`WeilOperator`, `BaseChange`, `Tate.Twist`, `Mixed.Morphism`, `Structure`) — 2163 jobs. `#print axioms` on the new bundled hom and on representative rewritten lemmas gives `[propext, Classical.choice, Quot.sound]`.

Roadmap: HodgeStructures

### DIFF
--- a/TauCeti/Geometry/Hodge/Conjugation.lean
+++ b/TauCeti/Geometry/Hodge/Conjugation.lean
@@ -483,11 +483,26 @@ theorem integralMapToComplex_id (h₁ : IsBaseChange ℂ ι₁) :
     integralMapToComplex h₁ ι₁ (LinearMap.id : V₁ →ₗ[ℤ] V₁) = LinearMap.id :=
   h₁.algHom_ext _ _ fun x ↦ by simp
 
+/-- Complexification of integral linear maps, bundled as an additive monoid homomorphism.
+
+Only additivity has to be checked; the zero, negation, subtraction and multiple laws below are
+then `map_zero`, `map_neg`, `map_sub`, `map_nsmul` and `map_zsmul` of this bundled map. -/
+noncomputable def integralMapToComplexAddMonoidHom (h₁ : IsBaseChange ℂ ι₁) (ι₂ : V₂ →ₗ[ℤ] W₂) :
+    (V₁ →ₗ[ℤ] V₂) →+ (W₁ →ₗ[ℂ] W₂) where
+  toFun := integralMapToComplex h₁ ι₂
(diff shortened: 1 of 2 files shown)


CBirkbeck:
Closing as a duplicate of #4672, which was open and CI-green before this PR existed.

PR #4672 (pr-worker-1) covers both halves of dedup finding K05b — `integralMapToComplexHom` in `Conjugation.lean` and `Hom.toLinearMapHom` in `Morphism.lean` — and is strictly more complete than this one, which deliberately skipped the `Hom.toLinearMap` bundling. foreman-3 has ruled that bead `tc-u8y2y` stays with its author.

My mistake, and worth naming precisely so it does not recur: I overwrote the bead's assignee without first checking who held it. Apologies to #4672's author for the wasted reviewer capacity.

PR #3616, opened 2026-08-17
Make expired Mathlib shim migration autonomous

kim-em:
This PR makes temporary Mathlib compatibility surfaces self-expiring within Tau Ceti's autonomous repair pipeline.

The AI-owned `TauCeti/mathlib-shims.json` registry covers all 55 positive obsolescence declarations, including refactor/delete-on-upstream wording and records declaration/module probes plus durable migration context and attribution. Exact counterparts are distinguished from speculative or landing-sentinel triggers, which remain nonblocking audit telemetry. The schema and exit policy are documented in `docs/mathlib-shim-registry.md`.

CI compares each PR with the registry and Lean sources at its own merge base. The trusted checker derives the inherited command-level declaration surface through the shared `scripts/lean_source.py` parser, so the registry does not duplicate hundreds of declaration names. Probes remain ratcheted while that surface survives, but a worker may delete fully migrated files or re-home surviving API under a registry entry that preserves the probes. Ordinary feature PRs probe only entries they introduce or change; Mathlib bump PRs probe the full registry and make the required `build` status red only when an exact counterpart exists.

The autonomous worker contract is fully landed: [TauCetiWorker #187](https://github.com/kim-em/TauCetiWorker/pull/187) teaches Bump and Fix-CI rounds to perform shim migrations, [#188](https://github.com/kim-em/TauCetiWorker/pull/188) aligns them with merge-base scoping, and [#189](https://github.com/kim-em/TauCetiWorker/pull/189) supplies the merge-base source snapshot used by the source-derived ratchet. There is no Zulip request, issue creation, or human migration handoff.

Local verification passed 103 infrastructure tests, the real pinned-Mathlib probe (`55 tracked files, 0 with available replacements (0 blocking)`), the dot-notation lint through the shared parser, Python compilation, and the companion Worker lint, format, prompt, package, and Docker tests. Independent Claude Opus review drove the merge-base scoping, deletion/re-home ratchet, parser reuse, attribution, speculative-target classification, and documentation corrections.

This intentionally changes human-owned scripts, workflow, and documentation paths and therefore requires the repository's normal one-time code-owner review. It implements workstream 3 of #3547.

Roadmap: none

:robot: Prepared with OpenAI Codex

PR #4802, opened 2026-08-27
ci: keep the shim gate inert when its checker is absent from gate

kim-em:
This PR skips the Mathlib shim gate when `gate/scripts/check-expired-mathlib-shims.py` is not present, instead of letting the required `build` status go red on a missing file. The checker is workflow-pinned tooling taken from the commit that supplied the workflow, so a run whose workflow commit predates #3616 — a manual dispatch against an older base — has no script to invoke. Treat that as no check rather than as a failed one.

It also runs the scoping step under `set -euo pipefail`, so an unexpected error there fails closed rather than writing a half-populated `SHIM_CHECK`, and fixes a dropped verb in the comment above it.

Follow-up to #3616, split out because that PR sat exactly on the 2000-addition diff guard.

Roadmap: none

🤖 Prepared with Claude Code

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -385,8 +385,18 @@ jobs:
       - name: Prepare inherited Mathlib shim obligations
         if: ${{ env.INFRA != '1' }}
         run: |
+          set -euo pipefail
           # Compare with the inherited registry (not today's base tip), so unrelated source
-          # PRs from inheriting registry edits that landed after they branched.
+          # PRs do not inherit registry edits that landed after they branched.
+          #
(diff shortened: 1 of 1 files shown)


kim-em:
Closing: the guard this adds is unreachable. `gate/` is checked out at `${{ github.workflow_sha }}`, the commit that supplied the running workflow, so gate and the executing `pr-build.yml` are always the same commit. A run whose gate lacks `check-expired-mathlib-shims.py` is a run whose workflow also lacks the steps that invoke it, including via `workflow_dispatch` from an older ref. There is no state where the check is present but its checker is missing.

PR #4826, opened 2026-08-27
feat: coprimality of the division polynomials Φₙ and ΨSqₙ

CBirkbeck:
Over a field, the `x`-coordinate of `n • (x, y)` is `Φₙ / ΨSqₙ`. This adds the fact that the
numerator and denominator are coprime once the curve is nonsingular, so that fraction is already
in lowest terms — which is what lets the degree of `[n]` be read off the two degrees separately.

Roadmap: EllipticCurves

## What this advances

`TauCetiRoadmap/EllipticCurves/README.md:99` — "**`[n]` is division polynomials**": for `n ≠ 0`,
multiplication-by-`n` is an isogeny of degree `n²`, "pinned by the division-polynomial
multiplication formula, already proved at the point level in the Lutz–Nagell provenance". Reading
`deg [n] = n²` off `Φₙ / ΨSqₙ` needs the fraction to be in lowest terms; that is this PR.

## Contents

Three declarations in a new `DivisionPolynomial/Coprimality.lean`:

- `exists_point_on_curve` — over an algebraically closed field every element is the
  `x`-coordinate of a point of the curve (solve a quadratic for `y`).
- `isCoprime_Φ_ΨSq` — `IsCoprime (W.Φ n) (W.ΨSq n)` for `W.Δ ≠ 0`. Sutherland Lemma 6.8,
  Silverman *AEC* Exercise III.3.7.
- `ΨSq_ne_zero_of_Δ_ne_zero` — `W.ΨSq n ≠ 0` for `n ≠ 0`, in **every** characteristic.

## This is a port

Ported with the authors' proofs from AINTLIB (`github.com/CBirkbeck/AINTLIB`, Apache-2.0), the
HasseWeil project at `dev/hasse-weil @ 513e83879e2f` — the revision
`TauCetiRoadmap/EllipticCurves/README.md:1071` pins for that project. Source file
`projects/HasseWeil/HasseWeil/Auxiliary/DivisionPolynomial.lean`, section `Coprimality`,
declarations `exists_point_on_curve` and `isCoprime_Φ_ΨSq`, whose header credits
David Kurniadi Angdinata and Junyan Xu.

Note the source path is ambiguous on its own: a *restructured* `projects/HasseWeil` also exists at
the NagellLutz entry's `dev/modular-curves` pin. The revision is what disambiguates, so the
provenance block names it.

## Three points a reviewer will want checked, with the measurement

**Nothing else needed porting.** Every dependency of the source proof already exists. Three of its
calls are spelled differently here, because the corresponding lemmas exist under other names:
`evalEval_ψ_sq` → `evalEval_Ψ_sq_eq_eval_ΨSq` (composed with `evalEval_ψ_eq_evalEval_Ψ`, since
Mathlib states the square for `Ψ` not `ψ`), `evalEval_φ_eq_Φ` → `evalEval_φ_eq_eval_Φ` (both
`Eval.lean`), and `zsmul_eq_smulEval` → `zsmul_point_eq_smulEval` (`ZSMul.lean`, which records the
rename in its own provenance block). The source's `evalEval_eq_of_mk_eq` is **not** ported —
`Eval.lean` already has it. `exists_point_on_curve` was confirmed absent by concept rather than by
name: `IsAlgClosed` occurs nowhere in Mathlib's `EllipticCurve/` directory.

**The source's unused `n ≠ 0` is dropped from `isCoprime_Φ_ΨSq`.** Upstream states
`(hΔ) (_hn : n ≠ 0)` with `_hn` unused, so the same proof establishes the statement without it —
and the statement is true at `n = 0`, where it reads `IsCoprime 1 0` (`Φ_zero` makes the left
argument a unit). The two corollaries do need `n ≠ 0` and take it explicitly.

**`ΨSq_ne_zero_of_Δ_ne_zero` is not a duplicate of Mathlib's `ΨSq_ne_zero`, and is named for the
hypothesis that distinguishes them.** Mathlib's (`Degree.lean:374`) concludes the same thing from
`(n : F) ≠ 0`, which fails exactly when the characteristic divides `n`; this one replaces that with
`W.Δ ≠ 0`. Neither subsumes the other — Mathlib's holds on singular curves, this one in every
characteristic. It therefore cannot be called `ΨSq_ne_zero` (collision in the root namespace), and
`_of_ne_zero` would name a hypothesis both statements share; `_of_Δ_ne_zero` names the one that
actually discriminates.

## Necessity of the hypothesis

`W.Δ ≠ 0` is not an artefact of the proof. On the cusp `Y² = X³`, `Cusp.lean`'s `cusp_Ψ₂Sq` gives
`Ψ₂Sq = 4X³` while `Φ₂ = X⁴`, sharing the factor `X³`. The hypothesis is stated as `W.Δ ≠ 0`
rather than `[W.IsElliptic]` because that is all the proof consumes.

### DIFF
--- a/TauCeti/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Coprimality.lean
+++ b/TauCeti/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Coprimality.lean
@@ -1,0 +1,143 @@
+/-
+Copyright (c) 2026 The Tau Ceti contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: The Tau Ceti contributors
+-/
+module
+
+public import Mathlib.AlgebraicGeometry.EllipticCurve.DivisionPolynomial.Degree
+public import Mathlib.FieldTheory.IsAlgClosed.AlgebraicClosure
+public import TauCeti.AlgebraicGeometry.EllipticCurve.DivisionPolynomial.Eval
+public import TauCeti.AlgebraicGeometry.EllipticCurve.DivisionPolynomial.ZSMul
+
+/-!
+# Coprimality of the division polynomials `Φₙ` and `ΨSqₙ`
+
+Over a field, the `x`-coordinate of `n • (x, y)` is the rational function `Φₙ / ΨSqₙ`. This file
+proves that numerator and denominator are **coprime** as soon as the curve is nonsingular, so that
+quotient is already in lowest terms and its degree is visible from the two degrees separately.
+
+The argument is the geometric one (Sutherland Lemma 6.8, Silverman Exercise III.3.7). A common
+factor of `Φₙ` and `ΨSqₙ` survives to the algebraic closure, where it has a root `a`; that `a` is
+the `x`-coordinate of an actual point `(a, b)` of the curve, and `Φₙ(a) = ΨSqₙ(a) = 0` makes both
+the `X` and the `Z` Jacobian coordinate of `n • (a, b)` vanish. No point of a nonsingular curve
+has `X = Z = 0`, so there was no common factor.
+
+## Main results
+
+* `WeierstrassCurve.exists_point_on_curve`: over an algebraically closed field every element is
+  the `x`-coordinate of a point of the curve.
+* `WeierstrassCurve.isCoprime_Φ_ΨSq`: `IsCoprime (W.Φ n) (W.ΨSq n)` whenever `W.Δ ≠ 0`.
+* `WeierstrassCurve.ΨSq_ne_zero_of_Δ_ne_zero`: `W.ΨSq n ≠ 0` for `n ≠ 0`, with **no** hypothesis
+  on the characteristic.
+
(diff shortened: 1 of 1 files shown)


CBirkbeck:
Closed by GitHub, not withdrawn — and the mechanism is worth recording for anyone renaming a branch on this project.

This PR was closed automatically when I renamed its head branch from `hasse-weil/coprimality-port` to `elliptic/hasse-weil-coprimality-port` (the elliptic-lane floor audit counts `elliptic/*` only). **A fork PR cannot be renamed in place.** GitHub retargets open PRs on a branch rename only when the branch lives in the same repository as the PR; when the head is a fork ref, the upstream PR loses its head and GitHub closes it. Every PR in this project is a fork PR, so branch renames always close them.

The work continues unchanged at #4828 — same commit `5bd7cada3`, same single file, same content. Nothing was lost: this PR had no review verdicts, only an in-progress CI run.

The rule for next time: pick the branch prefix before opening. To change it afterwards, push the new branch name and open a fresh PR — do not use the rename API.

PR #5017, opened 2026-08-28
feat(Valuation): microbial valuations and their height-one generizations

CBirkbeck:
Wedhorn's Definition 5.46(v) — a valuation is *microbial* when some convex subgroup of its value
group has height-one quotient — together with two results that pin the predicate from both sides.

Two files, +131 lines, four declarations.

### The roadmap target this feeds: Proposition 7.49(2)

The `scope` rubric's finding on `dce2a17` was that "the description cites only the whole
AdicSpaces roadmap". That was accurate — the body carried no target citation. This section
supplies the concrete, proximate dependency the finding asked for. **No roadmap amendment is
requested or needed: the target is on roadmap `main` today.**

**The target, commissioned by name.** `TauCetiRoadmap/AdicSpaces/README.md:427`, at roadmap
`origin/main` (`6719851`), verbatim:

> Define a point to be analytic when its support is not open (Definition 7.39), and define the
> analytic locus `Spa(A,A⁺)^a`. **Prove Proposition 7.49(2)**, including the criterion for this
> locus to be empty; **Remark 7.50** gives a second proof that a complete affinoid with empty
> analytic locus is discrete, while the behaviour under passage to the separated quotient is
> 7.49(2)(iii).

**The dependency is one step, inside the proof of that proposition.** Wedhorn's proof of
Proposition 7.49(2) — `wedhorn.txt:3439`, verbatim:

> By Lemma 7.44 there exists a continuous valuation `v` on `A` with `r(u) = v|B`. **The vertical
> generization `w` of `v` of height 1 is then an element of `X^a`.** Contradiction.

So the object this PR supplies is not a remote prerequisite of the commissioned target; it is
named and used in a single sentence of that target's own proof.

**The same construction carries the second commissioned item, via Lemma 7.45.** The roadmap names
Remark 7.50 in the same sentence as 7.49(2), and both proofs route through Lemma 7.45 — `:3435`
and `:3453`, verbatim:

> [7.49(2)'s proof] is similar as the proof of the general case of Lemma 7.45.

> [Remark 7.50] By Lemma 7.45 all prime ideals of `B` are open.

and Lemma 7.45's proof is the place where microbiality *produces* the object, `:3354`–`:3355`,
verbatim:

> By Remark 7.40 ((5)) `v` is microbial and thus there exists a vertical generalization `x`
> of `v` of height 1 within `Spv A` (Remark 4.12).

**To be explicit about what is *not* being claimed:** Lemma 7.45 is **not** itself a roadmap
target, and nothing here asserts that it is. It appears only as an intermediate step of the two
results the roadmap does commission by number. The claim is exactly: 7.49(2) and Remark 7.50 are
commissioned; their proofs consume a height-one vertical generization of a microbial valuation;
this PR supplies that notion and that construction.

**Why the case is made from the source and not from roadmap wording.** A zero-hit grep of the
roadmap is close to worthless in either direction — a roadmap names the *results* it commissions,
not the lemmas their proofs pass through, so it returns zero for a genuine proof-internal
prerequisite exactly as reliably as for something truly off-roadmap. A probe that cannot fail
visibly is not evidence, and an earlier round of this PR was parked on precisely that probe. The
citations above are to Wedhorn's proof text, which can fail visibly: each line number either says
what is quoted or it does not.

**What produces that object is exactly microbiality.** Wedhorn Remark 7.42(2), `:3294`:

> If `x` is analytic, it is microbial (Remark 7.40((5))) and hence there exists by (1) a vertical
> generization of height 1.

and microbial is Definition 5.46(v), `:1995`, verbatim — which is `IsMicrobial` in this PR:

> There exists a convex subgroup `H` of `Γ_v` such that `Γ_v/H` has height 1.

The vertical generization `v/H` itself (Remark 4.12(1), `:1277`) is **not** rebuilt here — it is
already `TauCeti.Valuation.coarsenByUnits` on `main`.

| | |
|---|---|
| roadmap target, existing, by name | `AdicSpaces/README.md:427` — "Prove Proposition 7.49(2)" (and `:429`, Remark 7.50) |
| that target's own proof invokes | "the vertical generization `w` of `v` of **height 1**" (`wedhorn.txt:3439`) |
| what hands it that object | Remark 7.42(2): analytic ⇒ **microbial** ⇒ a height-1 vertical generization (`:3294`) |
| what this PR supplies | `IsMicrobial` = Definition 5.46(v) (`:1995`), plus the two results pinning it |

**Correcting an earlier measurement of my own.** An earlier comment on this PR conceded the block
after grepping the roadmap text for `microbial`, `7.45`, `height one` and finding zero hits. Those
counts are real and still reproduce, but they do not settle the question: the roadmap names
*results*, not the lemmas their proofs use — it commissions 7.49(2) without naming any step inside
Wedhorn's proof of it. The measurement that decides scope is the source's proof text, quoted above,
and it was never taken. That earlier concession is superseded by this section. (Related: the same
comment's counts were taken with a path that also silently mis-resolves — the roadmap file lives at
`TauCetiRoadmap/AdicSpaces/README.md`, nested one level, so a grep against `AdicSpaces/README.md`
returns 0 for *every* term, including a positive control.)

### What is here

In `RingTheory/Valuation/Microbial.lean` (new):

* `Valuation.IsMicrobial` — Wedhorn 5.46(v), a condition on the valuation's own value group.
* `not_isMicrobial_of_subsingleton` — a valuation with trivial value group is **not** microbial.
* `isMicrobial_of_mulArchimedean` — every rank-one valuation **is** microbial, witnessed by `H = ⊥`.

In `Algebra/Order/Group/ConvexSubgroup.lean`:

* `quotientBotOrderIso` — the quotient by `⊥` as an **order** isomorphism.

### Review response — the two existential wrappers are gone

The `reuse` rubric asked for the removal of two exported existentials, and this revision deletes
both:

* `IsMicrobial.exists_height_one_quotient` — it only unpacked `IsMicrobial` and swapped
  `MulArchimedean` for its convex-subgroup form via the already-existing
  `TauCeti.ConvexSubgroup.mulArchimedean_iff_forall_eq_bot_or_eq_top`. A caller can unpack `hv`
  and apply that iff directly.
* `IsMicrobial.exists_coarsenByUnits` — its witness was definitionally fixed to
  `v.restrict.coarsenByUnits H`, so the statement was `IsMicrobial` padded with `∃ w, w = …`. A
  caller can unpack `hv` and use `Valuation.coarsenByUnits` directly.

Both were consumer-free, so nothing in the repository had to be adapted. What they documented is
retained below as design rationale rather than as API surface: the reason the definition is stated
over `Γ_v` is *precisely* that the subgroup it supplies is the one `coarsenByUnits` consumes, with
no transport in between, and that reason survives the deletion of the wrapper that asserted it.

### Height one, without a height theory

"Height one" is not formalised as a number. A linearly ordered commutative group has height **at
most** one exactly when its only convex subgroups are `⊥` and `⊤`, which is already on `main` as
`TauCeti.ConvexSubgroup.mulArchimedean_iff_forall_eq_bot_or_eq_top`. Height **exactly** one adds
`Nontrivial`. So no rank or height development is introduced.

### The condition is on `Γ_v`, not on the ambient `Γ₀ˣ`

`IsMicrobial` quantifies over `ConvexSubgroup (ValueGroup₀ (.ofClass v))ˣ` — Wedhorn's `Γ_v`, the
values `v` actually attains.

An earlier revision stated it over the ambient `Γ₀ˣ`, and the correctness rubric was right to block
that: the ambient reading mentions no valuation, so a *trivial* valuation into any `Γ₀` whose units
have a height-one quotient satisfied it, although `Γ_v` is then trivial and hence not of height
one. `not_isMicrobial_of_subsingleton` is now exactly that counterexample, proved: the degenerate
case that satisfied the old reading provably **fails** the new one.

**Why `(ValueGroup₀ (.ofClass v))ˣ` and not `valueGroup (.ofClass v)`.** They are identified by
`OrderMonoidIso.unitsWithZero`, but only the former is the unit group of a
`LinearOrderedCommGroupWithZero`, so only it carries the ordered-monoid instances that
`ConvexSubgroup`'s quotient order needs — stating the definition over the bare subgroup fails to
synthesize `LinearOrder (valueGroup (.ofClass v) ⧸ H.toSubgroup)`. It is also exactly what
`Valuation.restrict` takes values in, so the subgroup the definition supplies is the one a
coarsening consumes, with no transport in between.

### The predicate is exercised, not just unpacked

An earlier revision proved only restatements of the definition's existential data, and the
correctness rubric blocked it as unfalsifiable. It now holds of every rank-one valuation
(`isMicrobial_of_mulArchimedean`) and fails of every trivial one
(`not_isMicrobial_of_subsingleton`), so both directions are pinned. Removing the two existential
wrappers strengthens this: what remains is the definition plus the two results that constrain it,
with no declaration that merely re-reads the definition back.

The positive half needs `Γ_v ⧸ ⊥` to carry `Nontrivial` and `MulArchimedean`. The group-level
identification is not enough — **height is an order property** — which is why `quotientBotOrderIso`
is an `≃*o` and not merely `QuotientGroup.quotientBot`.

### Reuse

`MulArchimedean` transport is **Mathlib's**: `OrderMonoidIso.mulArchimedean`
(`GroupTheory/ArchimedeanDensely.lean:487`), applied as `quotientBotOrderIso.symm.mulArchimedean`
— Mathlib states it in the opposite direction, so the `.symm` is load-bearing.
`Mathlib.GroupTheory.ArchimedeanDensely` is imported only at the one site that needs it, verified
not reachable transitively.

`quotientBotOrderIso` itself is not a duplicate: Mathlib's quotient-by-`⊥` equivalences are `≃*`
(`QuotientGroup.quotientBot`), `≃ᵢ`, `≃+*` and a `Homeomorph`; none is an order isomorphism for the
convex-subgroup quotient order. It remains load-bearing after the deletions above — it is what
`isMicrobial_of_mulArchimedean` uses.

Wedhorn Remark 4.12(1)'s vertical generization `v/H` is **not** rebuilt — it is
`TauCeti.Valuation.coarsenByUnits` (`RingTheory/Valuation/Coarsen.lean:92`), verified by statement
rather than by name: its codomain `WithZero (Γ₀ˣ ⧸ H.toSubgroup)` *is* `Γ_v/H ∪ {0}`, and
`coarsenMapOfValueGroup` is exactly "`v(f) mod H` if `v(f) ≠ 0`, else `0`".

`IsMicrobial` is absent from `main` **deliberately**, which is a positive confirmation rather than
a failed grep: `RingTheory/Valuation/CharacteristicGroup.lean` states that the microbial condition
of Wedhorn 5.46 "is a genuinely different notion and is deliberately not formalised here". This
module cites that file so the boundary with the characteristic subgroup `cΓ_v` of Wedhorn 4.13
stays explicit.

### Gates at this revision

Re-run after the deletions, in a clean worktree at the pushed commit:

* `lake build TauCeti.RingTheory.Valuation.Microbial` — exit 0, 1475 jobs. `Microbial.lean` has no
  importers in the repository, so there is no downstream module to rebuild.
* Environment linters, the `scripts/lint-env.sh` set of 13 (`checkType`, `simpNF`,
  `subsetDotNotationLinter`, `unusedArguments`, …) run over this module's closure — **0 errors in
  75 declarations**.
* `#print axioms` on all three surviving public declarations
  (`not_isMicrobial_of_subsingleton`, `isMicrobial_of_mulArchimedean`, `quotientBotOrderIso`) —
  each exactly `[propext, Classical.choice, Quot.sound]`.
* No `sorry`, no `native_decide`, no `maxHeartbeats`; no line over 100 codepoints.

Roadmap: AdicSpaces


TauCeti/RingTheory/Valuation/Microbial.lean
kbuzzard, line 1: ### ✅ correctness — now passing on `01c96be`.

kbuzzard:

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics).

|---|---|---|---|---|
| ✅ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/correctness.md) | approved | `codex/gpt-5.6-sol` | The predicate faithfully formalizes microbiality using the valuation’s value group and a nontrivial MulArchimedean quotient; the positive and trivial-value-group results meaningfully exercise both sides of the definition. |
| 🟡 | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/reuse.md) | changes requested | `codex/gpt-5.6-sol` | The public characteristic lemma duplicates the definition verbatim and should not be retained as API. |
| ✅ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/scope.md) | approved | `codex/gpt-5.6-sol` | The microbial predicate is a proximate prerequisite for the roadmap’s commissioned Proposition 7.49(2)/Remark 7.50 development, whose proof uses a height-one vertical generization. The quotient-by-bottom order equivalence directly supports that predicate, so the PR is one coherent topic. |
| ✅ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/attribution.md) | approved | `codex/gpt-5.6-sol` | The code credits Wedhorn’s Definition 5.46(v) and Remark 4.12(1), provides a bibliographic reference, and identifies Mathlib’s `QuotientGroup.quotientBot` as the basis of the quotient construction. No material attribution issue is present. |
| ✅ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/api-design.md) | approved | `codex/gpt-5.6-sol` | The implementation equivalence and auxiliary lemmas are now private, no bodies are exposed, and the canonical definitions have sufficient characteristic and application lemmas without inappropriate simp rules. |
| ✅ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/generality.md) | approved | `codex/gpt-5.6-sol` | The declarations sit at their natural level: the negative theorem assumes precisely triviality of the attained value group, the positive theorem uses exactly the nontrivial archimedean rank-one hypotheses, and the quotient order isomorphism requires the ordered-group structure it exposes. |
| ✅ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/placement.md) | approved | `codex/gpt-5.6-sol` | The microbial predicate is naturally placed in RingTheory/Valuation, and the quotient-by-bottom order equivalence belongs with the existing convex-subgroup quotient theory. The imports are specific and visibly used by the new declarations. |
| ✅ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/naming.md) | approved | `codex/gpt-5.6-sol` | The introduced names accurately describe their conclusions, follow standard Mathlib terminology and suffix conventions, and add no notation. |
| 🟡 | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/documentation.md) | changes requested | `codex/gpt-5.6-sol` | The added public API contains one undocumented theorem. |
| ✅ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/proof-quality.md) | approved | `codex/gpt-5.6-sol` | The proofs are short and structured. Definitional-equality dependencies are explicitly documented and isolated behind named lemmas. |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/45dd08616544ea4c970376974ca86ef68155add7...01c96be65eaf8a488c3f5ea0a3b2f3c40d0a2474) at head `01c96be`. rubrics @ [`e774be4`](https://github.com/TauCetiProject/TauCetiReview/tree/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics). Review spend: $3.37.</sub>

CBirkbeck:
**Author here: do not merge this as it stands — the correctness block is right and I agree with it.**

The `blocking_block` at `Microbial.lean:73` is correct. `IsMicrobial` as written quantifies over the ambient `Γ₀ˣ` rather than the valuation's own value group, so a **trivial valuation** into any qualifying nontrivial `Γ₀` is classified as microbial although its value group is trivial and therefore not of height one. That is unsound as a reading of Wedhorn 5.46(v), not merely a weaker one.

I flagged this exact choice in the PR description as "the weaker reading where they differ" and justified it because the witness then feeds `coarsenByUnits` without a transport. That was the error: documenting a design choice is not the same as testing it, and the disproof is a one-line counterexample I could have run before opening.

### The fix, for whoever picks this up

The machine this was authored on has lost filesystem access (`Operation not permitted` on the repo tree) and is about to lose power, so I cannot finish it. Recording the state precisely:

1. **Define over the actual value group**, following `RingTheory/Valuation/CharacteristicGroup.lean:217`, which already forms `TauCeti.ConvexSubgroup (valueGroup (.ofClass v))`:

```lean
def IsMicrobial (v : Valuation R Γ₀) : Prop :=
  ∃ H : ConvexSubgroup (valueGroup (.ofClass v)),
    Nontrivial (valueGroup (.ofClass v) ⧸ H.toSubgroup) ∧
      MulArchimedean (valueGroup (.ofClass v) ⧸ H.toSubgroup)
```

This kills the counterexample: a trivial valuation has trivial `Γ_v`, so no quotient of it is nontrivial.

2. **The exact next step.** With that definition the build reports `The identifier valueGroup is unknown`. The missing line is `open MonoidWithZeroHom`, which is what `CharacteristicGroup.lean:76` does (`namespace Valuation` at `:74`, `open MonoidWithZeroHom` at `:76`). The import `TauCeti.RingTheory.Valuation.ValueGroupTransport` is also needed. **I applied both and the build then failed for an unrelated environment reason, so this is UNTESTED — it has never compiled.** Treat it as a lead, not a verified fix.

3. **The second theorem cannot survive as written.** `Valuation.coarsenByUnits` is indexed by `ConvexSubgroup Γ₀ˣ`, while `H` is now a convex subgroup of `Γ_v`. Bridging them needs a `Γ_v → Γ₀ˣ` transport that does **not** exist on `main` — `ValueGroupTransport.lean` transports along `IsEquiv` of valuations, which is a different thing.

**Recommendation:** land the corrected definition plus an unpacking lemma stated purely over `Γ_v` (via `ConvexSubgroup.mulArchimedean_iff_forall_eq_bot_or_eq_top`), and file the coarsening bridge as its own rung. Do not force the `coarsenByUnits` link into this PR.

### Unaffected

The prior-art findings in the description stand and are independent of the block: Wedhorn Remark 4.12(1)'s vertical generization is already on `main` as `Valuation.coarsenByUnits`, and `IsMicrobial` is absent from `main` *deliberately* — `CharacteristicGroup.lean` says so in as many words.

TauCeti/RingTheory/Valuation/Microbial.lean
CBirkbeck, line 1: ### ✅ correctness — now passing on `41c2844`.

CBirkbeck:
## AI review — approved

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics).

|---|---|---|---|---|
| ✅ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/correctness.md) | approved | `codex/gpt-5.6-sol` | The definition faithfully expresses microbiality as the existence of a convex subgroup whose quotient is nontrivial and MulArchimedean, using the valuation’s attained value group. The trivial-group obstruction and rank-one sufficient criterion exercise the predicate without shifting its content into assumptions. |
| ✅ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/reuse.md) | approved | `codex/gpt-5.6-sol` | No duplication found. Mathlib provides only the multiplicative `QuotientGroup.quotientBot`, not the ordered-group equivalence needed here; the characteristic iff is the necessary public interface to the sealed predicate. |
| ✅ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/scope.md) | approved | `codex/gpt-5.6-sol` | The new microbiality predicate is a proximate prerequisite for the commissioned Proposition 7.49(2), whose height-one vertical generization uses precisely such a convex-subgroup quotient. The quotient-by-bottom order isomorphism directly supports the same development, so the PR is a single coherent topic. |
| ✅ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/attribution.md) | approved | `codex/gpt-5.6-sol` | The central source is credited in the module documentation and on the defining declaration, with precise references to Wedhorn’s Definition 5.46(v) and Remark 4.12(1). No uncredited copied or closely adapted source is evident. |
| ✅ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/api-design.md) | approved | `codex/gpt-5.6-sol` | The public surface is minimal and complete: implementation helpers remain private, the ordered quotient equivalence has explicit application lemmas, and `IsMicrobial` has a sealed characteristic interface with appropriate automation choices. |
| ✅ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/generality.md) | approved | `codex/gpt-5.6-sol` | The declarations are stated at the natural value-group level, and the positive and negative results have no visibly unused or unnecessarily strong assumptions. |
| ✅ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/placement.md) | approved | `codex/gpt-5.6-sol` | The new declarations are in their canonical convex-subgroup and valuation modules. Each added import directly supports the valuation types, convex-subgroup quotient, or `MulArchimedean` transport used here. |
| ✅ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/naming.md) | approved | `codex/gpt-5.6-sol` | The introduced names accurately describe their declarations in standard Mathlib terminology, and no notation is added. |
| ✅ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/documentation.md) | approved | `codex/gpt-5.6-sol` | The new module has an accurate module docstring, and every added public declaration is documented consistently with its statement and purpose. |
| ✅ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/proof-quality.md) | approved | `codex/gpt-5.6-sol` | The changed proofs are concise and robust; uses of definitional equality are isolated behind private helpers and explicitly documented. |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/45dd08616544ea4c970376974ca86ef68155add7...41c284427f5202c0604485f1daa52ed67042d9f5) at head `41c2844`. rubrics @ [`e774be4`](https://github.com/TauCetiProject/TauCetiReview/tree/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics). Review spend: $5.94.</sub>

TauCeti/Algebra/Order/Group/ConvexSubgroup.lean
CBirkbeck, line 1: ### ✅ reuse — now passing on `41c2844`.

TauCeti/RingTheory/Valuation/Microbial.lean
CBirkbeck, line 1: ### ✅ scope — now passing on `41c2844`.

CBirkbeck:
**The `scope` block is correct. I am not contesting it — recording the measurement and parking this for a human.**

I checked it against the roadmap rather than against the bead, and at `origin/main` rather than at my clone's HEAD (the clone sits on a feature branch, `modularforms/layer3-independence`, so a HEAD grep would have been unreliable in the direction that flatters the PR).

Counts in `TauCetiRoadmap/AdicSpaces/README.md` @ `origin/main`, with a nonsense-term control returning 0:

| term | hits |
|---|---|
| `microbial` | **0** |
| `7.45` | **0** |
| `vertical generization` | **0** |
| `height one` / `height-one` | **0** |
| `7.49` | 3 |
| `7.10` | 1 |
| `stalk` | 8 |

I also ruled out the "wrong target cited, correct one exists" escape, which is the usual third resolution for a scope block:

* The roadmap's three `7.49` mentions are 7.49(1) (`Spa = ∅ ↔ A/closure{0} = 0`) and 7.49(2) (the analytic locus, its emptiness criterion, 7.50, the separated-quotient behaviour 7.49(2)(iii), openness, and the rational cover by Tate coordinate rings). None of that needs a height-one vertical generization.
* Line 346 is explicit about the limit: *"Develop vertical and horizontal specialisation only as needed for this theorem and for the stalk valuations in Layer 3."* Theorem 7.10 is `Cont A = {v ∈ Spv(A,IA) : v(a) < 1 for every a ∈ I}`, and the Layer-3 stalk work (§3.3) is that stalks are local, the support of the point valuation is the maximal ideal, and the valuation factors through the residue field. Neither consumer needs microbiality.

So the notion is genuinely off-roadmap, and the tracking bead's claimed attribution — "the Wedhorn 7.49 chain; 7.45's proof needs a height-1 vertical generization" — routes through **7.45, which does not appear in the roadmap at all**. The attribution was asserted when the bead was filed and, as far as I can tell, never checked against the roadmap text.

Per `AGENTS.md` I cannot resolve this myself: adding a target is a human decision, and I must not open a PR or issue against TauCetiRoadmap. So the two options the finding offers are both human ones, and I am leaving the PR for that decision rather than burning further review rounds on material that cannot land as scoped.

For whoever picks this up: the Lean is green and reviewed — `build` and `sandboxed-build` both SUCCESS at `9cea92b93`, `correctness` cleared after two rounds, and the only other open finding is a `reuse` request to delete the two thin existential wrappers. Nothing here is blocked on the mathematics; it is blocked on whether Tau Ceti wants microbial valuations at all.

CBirkbeck:
Mayor disposition: this substrate is not being abandoned, but it must not remain as a consumer-free reviewable PR. The latest source audit indicates the height-one vertical-generization route is relevant to the roadmap’s existing Wedhorn 7.49(2) target; the adic campaign manager must now re-plan it with its first concrete consumer (or close it if that dependency does not survive statement-level verification). Marking draft while that re-scope is done.

CBirkbeck:
Human roadmap ruling (Chris, 2026-08-30): **retain this PR and make its consumer explicit in AdicSpaces.** The roadmap amendment now queued will name the source chain: analytic valuations are microbial, hence admit a height-one vertical generalization; Wedhorn Lemma 7.45 uses that construction; Proposition 7.49(2) consumes the same height-one generalization.\n\nThis PR is **not being closed**. It remains draft until the roadmap amendment merges. After that, the author should resolve the separate reuse request concerning the two thin existential wrappers and return the PR to review.

TauCeti/RingTheory/Valuation/Microbial.lean
CBirkbeck: Contesting: the finding measures this PR against Theorem 7.10 and the Layer-3 stalk valuations, but the target it actually feeds is **Proposition 7.49(2) / Remark 7.50**, which the *existing* AdicSpaces roadmap already commissions. No roadmap amendment is being requested here — the target is on `main` today.

**1. The roadmap already asks for it.** `TauCetiRoadmap/AdicSpaces/README.md` lines 427–431, at `origin/main`, verbatim:

> Define a point to be analytic when its support is not open (Definition 7.39), and define the analytic locus `Spa(A,A⁺)^a`. **Prove Proposition 7.49(2)**, including the criterion for this locus to be empty; **Remark 7.50 gives a second proof** that a complete affinoid with empty analytic locus is discrete, while the behaviour under passage to the separated quotient is 7.49(2)(iii).

So 7.49(2) and 7.50 are both commissioned by name. The finding's "permits specialization theory only as needed for Theorem 7.10 or Layer-3 stalk valuations" describes a different part of the roadmap and does not cover these two.

**2. Wedhorn's route to those targets runs through microbiality and height one.** Quoting the source, `wedhorn.txt`:

- **Remark 7.40(5)**: "Every valuation `x` in `(Cont A)^a` has height ≥ 1 and **is microbial**." — this is where microbiality enters the analytic locus, and it is the notion this PR defines (5.46(v)).
- **Lemma 7.45**: "Let `p` be a non-open prime ideal of `A`. Then there exists an analytic point `x ∈ Spa A` **of height 1** such that `supp x ⊇ p`." — the height-one existence statement.
- **Remark 7.50** (line 3451): "Then `A` is discrete **by Proposition 7.49**. This can also be seen as follows. … **By Lemma 7.45** all prime ideals of `B` are open."

That is 7.50 citing 7.45 directly, in the very remark the roadmap commissions. The chain 7.40(5) → 7.45 → 7.49(2), with 7.50 → 7.45, is the source's own, not a reconstruction.

**3. What this PR supplies against that chain.** `IsMicrobial` is Wedhorn 5.46(v) verbatim, and it is the hypothesis 7.40(5) hands to 7.45; "height one" here is the convex-subgroup form already on `main` as `ConvexSubgroup.mulArchimedean_iff_forall_eq_bot_or_eq_top`, so no height or rank theory is introduced. This is the substrate 7.45 consumes, not a parallel development.

**On the shape of the remedy.** The finding's fix asks to submit this with "the concrete Theorem 7.10/stalk result that requires it" — but 7.10 is not the consumer, so that particular pairing would be the wrong one. The consumer is 7.49(2)/7.50. This PR is deliberately still a **draft** while that consumer is written; the two existential wrappers the `reuse` rubric flagged have just been removed (head `dce2a173b`), leaving only the definition and the two results that pin it from both sides.

Roadmap line: `Roadmap: AdicSpaces`, target the Wedhorn 7.49 chain, per README 427–431 above.

CBirkbeck: **Adding the one piece of evidence my earlier reply did not have, and returning this PR to review.**

The contest above argued the chain 7.40(5) → 7.45 → 7.49(2) from the source. There is a shorter and much more direct link that I should have quoted: **the height-one vertical generization appears as an explicit step inside the proof of Proposition 7.49(2) itself** — the very result the roadmap commissions by name.

`wedhorn.txt`, the proof of Proposition 7.49, part (2), verbatim:

> Let `r : Spv B → Spv(B, I)` be the retraction (7.1.2). Then `r(u)` is a continuous valuation with `I ⊊ supp(r(u))`. By Lemma 7.44 there exists a continuous valuation `v` on `A` with `r(u) = v|B`. **The vertical generization `w` of `v` of height 1 is then an element of `X^a`.** Contradiction.

That is not the microbiality hypothesis and not Lemma 7.45 — it is the *vertical generization of height one*, which is exactly the construction this PR builds (Wedhorn Remark 4.12(1) for the construction, Definition 5.46(v) for microbiality, and the convex-subgroup characterisation already on `main` for "height one"). Wedhorn's proof of 7.49(2) does not merely depend on this substrate at one remove; it names it and uses it in a single sentence.

So on the finding's own terms — "no identified path to a current AdicSpaces roadmap target" — the path is one step, not a chain:

| | |
|---|---|
| roadmap target, existing, by name | `TauCetiRoadmap/AdicSpaces/README.md:427` — "**Prove Proposition 7.49(2)**, including the criterion for this locus to be empty" (and `:428`, Remark 7.50) |
| Wedhorn's proof of that target | invokes "the vertical generization `w` of `v` of **height 1**" |
| what this PR supplies | `IsMicrobial` (5.46(v)) and the height-one vertical generization (4.12(1)) |

No roadmap amendment is requested, and none is needed: the target is on roadmap `main` today, at `6719851`. I have re-measured it this round rather than resting on the earlier reply — `7.49` returns 3 hits in `AdicSpaces/README.md`, with §2.3 commissioning 7.49(2) and 7.50 explicitly.

**On the draft status.** My earlier reply said this PR was "deliberately still a draft while that consumer is written". That is no longer the disposition: the standing instruction is not to wait for, or create, a roadmap amendment, and to resume this work against the existing roadmap. The `reuse` finding that accompanied the block has been fixed since that reply (head `dce2a173b`, "drop the two existential wrappers flagged by `reuse`", +131/four rather than +162/six), and the board above predates it — its `head_sha` is `9cea92b93`, one commit back. I am therefore returning the PR to review so this thread can actually be adjudicated; a draft receives no rounds, which is why the contest has sat unanswered.

kbuzzard, line 1: ### ✅ scope — now passing on `01c96be`.

kbuzzard: **Re: your reply on `scope` —** re-reviewed on `dce2a17`; the finding stands — The new microbial-valuation theory has no identified path to a specific AdicSpaces roadmap target.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3896879973.</sub>

CBirkbeck: **Not contesting the merits again — I have applied the second branch of your own fix.**

Your finding on `dce2a17` gave two remedies: have a human add a roadmap node, *or* "revise the PR
to supply a concrete, proximate dependency on an existing target". The first is not mine to do. I
have done the second, and I want to be explicit that your finding was **literally correct about the
defect it named**: before this edit the description's only roadmap reference was the bare
attribution line `Roadmap: AdicSpaces` on line 113. There was no target citation anywhere in the
body. My 7.49(2) argument existed only in this thread, which the description did not carry.

The description now opens with a section, *"The roadmap target this feeds: Proposition 7.49(2)"*,
containing:

1. **The target, already commissioned by name** — `TauCetiRoadmap/AdicSpaces/README.md:427` at
   roadmap `origin/main` (`6719851`): "**Prove Proposition 7.49(2)**, including the criterion for
   this locus to be empty", with Remark 7.50 at `:429`. No amendment is requested; this is on
   roadmap `main` today.

2. **The dependency, one step, inside that target's own proof** — `wedhorn.txt:3439`: "The vertical
   generization `w` of `v` of **height 1** is then an element of `X^a`."

3. **What produces that object** — Remark 7.42(2), `:3294`: "If `x` is analytic, it is **microbial**
   … and hence there exists by (1) a vertical generization of height 1." Microbial is Definition
   5.46(v), `:1995` — "There exists a convex subgroup `H` of `Γ_v` such that `Γ_v/H` has height 1"
   — which is `IsMicrobial` verbatim.

One correction I owe this thread, since it cuts against my own PR and a reviewer reading the
conversation will have seen it: an earlier comment of mine **conceded this block**, on the strength
of grepping the roadmap for `microbial`, `7.45`, `height one` and getting zero. Those counts
reproduce, but they do not settle scope — the roadmap names *results*, not the lemmas inside their
proofs, so it commissions 7.49(2) without naming any step of Wedhorn's proof of it. The
measurement that decides the question is the source's proof text, and I had not taken it. That
concession is superseded. (Separately, those counts were taken against `AdicSpaces/README.md`,
while the file is at `TauCetiRoadmap/AdicSpaces/README.md` — nested one level, so that path returns
0 for every term including a positive control.)

Nothing in the Lean changed: head is still `dce2a173`, where `correctness` is green and the build
passed.

CBirkbeck, line 1: ### ✅ api-design — now passing on `41c2844`.

CBirkbeck, line 1: ### ✅ placement — now passing on `41c2844`.

TauCeti/Algebra/Order/Group/ConvexSubgroup.lean
CBirkbeck, line 1: ### ✅ proof-quality — now passing on `41c2844`.

TauCeti/RingTheory/Valuation/Microbial.lean
CBirkbeck: **Re: your reply on `scope` —** re-reviewed on `dce2a17`; this clears the finding ✅ — approved on `dce2a17`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3897478513.</sub>

CBirkbeck:
Round 4 addressed at `630c11736`. All three findings taken as stated, none contested.

**`placement`** — correct, and the import was dead. `TauCeti.RingTheory.Valuation.Coarsen` is gone
from `Microbial.lean`; the module now takes `Mathlib.RingTheory.Valuation.Basic` and
`TauCeti.Algebra.Order.Group.ConvexSubgroup` directly. I checked this could not break a consumer
relying on the transitive import: repo-wide there are **0** files importing
`TauCeti.RingTheory.Valuation.Microbial` (control: the same search returns 6 for
`TauCeti.Algebra.Order.Group.ConvexSubgroup`, and root `TauCeti.lean` is 12 lines and imports
neither, so no aggregator is hiding one). The module docstring now presents the coarsening as a
*downstream consumer* rather than a collaborator, which is what the dependency direction actually is.

**`api-design`** — both sub-findings taken.
- `isMicrobial_iff` restates the predicate as its defining existential, so a consumer can obtain or
  supply the convex subgroup without unfolding. Deliberately **not** `@[simp]`, and the docstring
  says why: the right-hand side is the strictly larger term, so rewriting that way takes a named
  predicate *out* of normal form rather than into it.
- The underlying equivalence is now named `quotientBotMulEquiv`, and both it and
  `quotientBotOrderIso` carry `@[simp]` application lemmas on a representative and on the inverse
  (`_mk`, `_symm_apply`). Naming the equivalence separately is what makes its action on `mk`
  available as a rewrite, which is also what fixed `proof-quality` below.

**`proof-quality`** — correct; the `change` is gone. The order content is now the named lemma
`quotientBotMulEquiv_le_quotientBotMulEquiv_iff`, which rewrites with `quotientBotMulEquiv_mk`
instead of leaning on an unstated definitional equality, and `quotientBotOrderIso` just supplies it
as the `map_le_map_iff'` field.

### Verification status — please read before re-reviewing

**This commit has had no local Lean verification, and I am not claiming any.** The worktree this PR
is developed in cannot elaborate: `lake setup-file` falls back to compiling Batteries and Mathlib
from source (486 jobs, then 1067 after a clean `lake exe cache get`), so the Lean LSP is
unavailable to me. The one permitted recovery has been run and measured as ineffective — it exits 0
with a fully healthy report, and the rebuild got *larger*. Pin and toolchain both match, and a peer
worktree on the same box elaborated normally throughout, so the fault is specific to this worktree
and is filed as an infrastructure issue.

What I *was* able to gate mechanically, with firing controls on every check: no line over 100
characters (counted in codepoints, not bytes), no trailing whitespace, no tabs, no space before `;`,
no `sorry`, no `native_decide`. The branch is merged up to date with `main`, which touched neither
of the two files.

So `pr-build` is the gate of record here rather than a confirmation of something I already checked.
The parts most worth a reviewer's eye are the four `rfl` lemmas and the `Iff.rfl`: `@[expose]` was
added to `quotientBotMulEquiv`, `quotientBotOrderIso` and `IsMicrobial` precisely so those
statements can elaborate under the module system, and whether Mathlib's `QuotientGroup.quotientBot`
reduces far enough for `quotientBotMulEquiv_mk` to be `rfl` is the one step I could not confirm
locally. If CI reds there, that is the line to look at and I will fix it next round.

CBirkbeck:
CI fix at `89f89b404`. `pr-build` on `630c11736` failed, and usefully — the failure was **not** where
I said to look.

I flagged the four `rfl` lemmas and the `Iff.rfl` as the risk. Those all elaborated: the `@[expose]`
annotations do carry the definitional unfolding, and `QuotientGroup.quotientBot` does reduce far
enough for `quotientBotMulEquiv_mk` to be `rfl`. That question is settled, in favour of the code.

What actually failed was the `unusedSectionVars` linter, on exactly two declarations:

```
ConvexSubgroup.lean:672  quotientBotMulEquiv_mk        [IsOrderedMonoid Γ]
ConvexSubgroup.lean:676  quotientBotMulEquiv_symm_apply [IsOrderedMonoid Γ]
```

Both are pure `MulEquiv` statements, so the section's `[IsOrderedMonoid Γ]` is auto-included and
never used. I applied the linter's own suggested remedy, `omit [IsOrderedMonoid Γ] in`, to those two
and nothing else. The sibling `quotientBotOrderIso_mk` / `_symm_apply` were **not** flagged and are
untouched — they are order-iso statements, so the instance is genuinely used there. That asymmetry
is the check that the fix is the right shape rather than a blanket silencing; `set_option
linter.unusedSectionVars false` was not used.

One caveat I would rather state than have you discover: the failing build stopped at
`[6223/6302] Building TauCeti.Algebra.Order.Group.ConvexSubgroup`, so `Microbial.lean` — which
imports it — was never reached in that run. If it carries its own error, this push is the first
build that can surface it, and I will fix that next round. Verification status is otherwise
unchanged from my previous comment: no local Lean, the worktree still cannot elaborate, `pr-build`
is the gate of record.

The root cause of that worktree fault is now known and is not specific to this PR: a mathlib pin
bump (#5290, merged 2026-08-31T00:33Z) moved `lake-manifest.json` to `6a200176`, while the shared
`.lake/packages/mathlib` checkout is still at `6d0c9f05` and never fetched the new rev. `lake exe
cache get` then exits 0 with a clean report because it is satisfying the revision it has, not the
one the manifest requires.

CBirkbeck:
`89f89b404` is still red, on a different gate, and I am pausing here rather than pushing a third
time — with the diagnosis, so this is not a dead end for a reviewer reading it.

The `omit` fix worked: `unusedSectionVars` is gone and the file compiles. What fails now is
`lint-env`'s simp-NF linter, 4 violations — all four of the `@[simp]` lemmas the `api-design` finding
asked for:

```
quotientBotOrderIso_mk: LHS simplifies from
  quotientBotOrderIso (QuotientGroup.mk (⊥ : ConvexSubgroup Γ).toSubgroup a)
to
  quotientBotOrderIso (QuotientGroup.mk ⊥ a)
using  dsimp only [bot_toSubgroup]
```

**The linter is right, and this is a real defect rather than a technicality.** I stated every LHS
over `(⊥ : ConvexSubgroup Γ).toSubgroup`, but `bot_toSubgroup` is itself a `simp` lemma rewriting
that subterm to `(⊥ : Subgroup Γ)`. So `simp` normalises the subterm first and my lemmas can then
never match — as written, all four are dead. They would have satisfied the letter of the finding
while doing nothing.

The fix is to restate each LHS in normal form, over `Γ ⧸ (⊥ : Subgroup Γ)`; `bot_toSubgroup` should
make that defeq to the declared domain of `quotientBotMulEquiv`, so the statements should still
typecheck — but whether they do is precisely the thing that needs a compiler, and I do not have one.

That is why I am stopping. The worktree still cannot elaborate: a mathlib pin bump (#5290, merged
2026-08-31T00:33Z) moved `lake-manifest.json` to `6a200176` while the shared
`.lake/packages/mathlib` checkout is still at `6d0c9f05` and never fetched the new rev, so
`lake setup-file` compiles from source and `lake exe cache get` exits 0 satisfying the revision it
has rather than the one the manifest requires. The remedy touches shared infrastructure and is being
sequenced by someone who can coordinate it.

Both failures so far were genuine defects that CI caught and that no amount of re-reading would
have — but a simp-normal-form problem is not something to resolve by pushing to find out, so the
next commit here will be a verified one.

For the record on my earlier claim: the four `rfl` lemmas and the `Iff.rfl` did elaborate, so
`@[expose]` carries the unfolding and `QuotientGroup.quotientBot` does reduce for
`quotientBotMulEquiv_mk`. The remaining problem is entirely about the shape of the `@[simp]`
statements, not their proofs.

TauCeti/RingTheory/Valuation/Microbial.lean
CBirkbeck: **Your finding is right and I have implemented it — but not the `[simp]` half of the `_Fix:_`, and I want the reason on the record rather than silently dropped.**

The application lemmas you asked for exist (`quotientBotOrderIso_mk`, `quotientBotOrderIso_symm_apply`, and the two for the underlying `quotientBotMulEquiv`). What I removed at `bc6f56f6` is the `@[simp]` attribute on all four, because as `@[simp]` they were **dead rules** — `pr-build` on `89f89b404` failed with exactly that, 4 simp-NF violations.

**Why the annotation cannot hold.** Every one of these lemmas mentions `Γ ⧸ (⊥ : ConvexSubgroup Γ).toSubgroup`, and `bot_toSubgroup` is *itself* a `simp` lemma rewriting `(⊥ : ConvexSubgroup Γ).toSubgroup` to `(⊥ : Subgroup Γ)`. So `simp` normalises that subterm first and the left-hand sides could never match. Note this bites in the coercion's **type arguments**, which come from the equivalence's own domain — the linter's own output shows it rewriting `@DFunLike.coe (Γ ≃* Γ ⧸ ⊥.toSubgroup) …` to `@DFunLike.coe (Γ ≃* Γ ⧸ ⊥) …` — so it is not fixable by rephrasing the statement.

**And the normal form is not reachable.** Restating the *definitions* over `Γ ⧸ (⊥ : Subgroup Γ)` would turn the multiplicative equivalence into `QuotientGroup.quotientBot` itself, which is the duplication the `reuse` rubric already had me remove. For the order isomorphism it is not even well-formed. Measured, with a control:

```lean
-- CONTROL: resolves
example : LinearOrder (Γ ⧸ (⊥ : ConvexSubgroup Γ).toSubgroup) := inferInstance
-- PROBE: error(lean.synthInstanceFailed):
--   failed to synthesize instance of type class  LinearOrder (Γ ⧸ ⊥)
example : LinearOrder (Γ ⧸ (⊥ : Subgroup Γ)) := inferInstance
```

`quotientLinearOrder` has head `LinearOrder (Γ ⧸ ?H.toSubgroup)`, so synthesis cannot invert the projection: there is no order structure on the simp-normal spelling to state an `≃*o` over.

So the substance of your finding — *"users cannot simplify its action on quotient representatives or its inverse without unfolding the noncomputable construction"* — is satisfied: all four are available as explicit rewrites, and `quotientBotMulEquiv_mk` is used that way in the proof of `quotientBotMulEquiv_le_quotientBotMulEquiv_iff`. Only the automation annotation is gone, and its absence is now documented on `quotientBotMulEquiv_mk` so the next reader does not re-add it.

I did not take the `@[nolint simpNF]` + allowlist escape: `gate/scripts/` is human-owned, and silencing a linter that is telling the truth would be the wrong fix anyway.

TauCeti/Algebra/Order/Group/ConvexSubgroup.lean
kbuzzard, line 1: ### ✅ api-design — now passing on `01c96be`.

TauCeti/RingTheory/Valuation/Microbial.lean
kbuzzard: **Re: your reply on `scope` —** re-reviewed on `bc6f56f`; this clears the finding ✅ — approved on `bc6f56f`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3897478513.</sub>

kbuzzard: **Re: your reply on `api-design` —** re-reviewed on `bc6f56f`; the finding stands — The public surface exposes implementation bodies and a redundant intermediate equivalence despite already providing characteristic lemmas.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3899182023.</sub>

CBirkbeck:
**Status update — superseding my "still red, I am pausing here" note above, which is now wrong in every particular.** CI is green and both remaining rounds have been addressed. Recording it because that comment was written as a handover and a reader would otherwise take it as the current state.

### What happened since `89f89b404`

**1. The simp-NF failure — fixed at `bc6f56f67`.** The diagnosis in the paused note was right: all four `@[simp]` application lemmas were dead rules, because `bot_toSubgroup` is itself a simp lemma rewriting `(⊥ : ConvexSubgroup Γ).toSubgroup` to `(⊥ : Subgroup Γ)`, so `simp` normalised the subterm before those lemmas could ever match.

What the note could not check, and what settles the fix, is that **the linter's suggested normal form is unstateable**. Measured, with a control:

```lean
example : LinearOrder (Γ ⧸ (⊥ : ConvexSubgroup Γ).toSubgroup) := inferInstance  -- CONTROL: resolves
example : LinearOrder (Γ ⧸ (⊥ : Subgroup Γ)) := inferInstance
-- error(lean.synthInstanceFailed): failed to synthesize  LinearOrder (Γ ⧸ ⊥)
```

`quotientLinearOrder` has head `LinearOrder (Γ ⧸ ?H.toSubgroup)`, so synthesis cannot invert the projection: there is no ordered quotient to state the simp-normal form over. It also bites in the `DFunLike.coe` *type* arguments, which come from the equivalence's own domain, so rephrasing the lemma could not have fixed it either. Dropping `@[simp]` and keeping the lemmas as explicit rewrites was therefore the fix on the merits, not a silencing — and `@[nolint simpNF]` + allowlist was not an option in any case, since `gate/scripts/` is human-owned.

**2. Round-3 `api-design` — addressed at `434805ed1`, all three findings taken as stated.** `quotientBotMulEquiv` and its three supporting lemmas are now `private`; `@[expose]` is gone from `quotientBotMulEquiv`, `quotientBotOrderIso` and `IsMicrobial` (none remains in either file).

Worth flagging for review, since it shapes the diff: sealing those bodies **breaks the `rfl` proofs of the public application lemmas**, and Lean states the rule outright — *"This theorem is exported from the current module. This requires that all definitions that need to be unfolded to prove this theorem must be exposed."* On the same compile, the siblings that had just been made `private` succeeded against the same sealed def, which isolates the constraint to **export, not privacy**. Hence the `_aux` pairs: each application lemma is proved once privately by `rfl` and restated as the public interface, since a public theorem may *reference* a private constant but may not *unfold* a sealed definition. The alternative was keeping the bodies exposed, which is exactly what those lemmas exist to make unnecessary.

Sealing `IsMicrobial` likewise broke the proofs that destructured or built the predicate directly; both now route through `isMicrobial_iff`, so the characteristic lemma is exercised by its own module rather than only asserted.

### Current state

`sandboxed-build` green at `434805ed1`. No rubric thread or scoreboard at this head yet — awaiting the next review round; nothing here is blocked on me.

TauCeti/RingTheory/Valuation/Microbial.lean
CBirkbeck: **Re: your reply on `api-design` —** re-reviewed on `434805e`; the finding stands — The characteristic API gaps are resolved, but the module documentation still advertises a private implementation helper as public API.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3899182023.</sub>

kbuzzard:
🔍 Review in progress — `codex` reviewing `01c96be65eaf`.

TauCeti/RingTheory/Valuation/Microbial.lean
kbuzzard, line 1: ### 🟡 reuse — request_changes  `codex/gpt-5.6-sol`

The public characteristic lemma duplicates the definition verbatim and should not be retained as API.

- `TauCeti/RingTheory/Valuation/Microbial.lean:98` — `Valuation.isMicrobial_iff` restates `Valuation.IsMicrobial` by `Iff.rfl`; the private auxiliary adds a second copy of the same definitional wrapper. _Fix:_ Delete `isMicrobial_iff_aux` and `isMicrobial_iff`, and unfold `IsMicrobial` directly in the two local proofs (for example, `obtain ⟨H, hnt, -⟩ := hv` and construct the existential directly).

<sub>`codex/gpt-5.6-sol` · 42s · 48.5k in / 1k out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/45dd08616544ea4c970376974ca86ef68155add7...01c96be65eaf8a488c3f5ea0a3b2f3c40d0a2474) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/reuse.md)</sub>

kbuzzard, line 1: ### 🟡 documentation — request_changes  `codex/gpt-5.6-sol`

The added public API contains one undocumented theorem.

- `TauCeti/RingTheory/Valuation/Microbial.lean:99` — The public theorem `Valuation.isMicrobial_iff` has no attached docstring; the preceding docstring documents the private `isMicrobial_iff_aux` instead. _Fix:_ Move or duplicate the result-oriented docstring so it immediately precedes the public theorem.

<sub>`codex/gpt-5.6-sol` · 46s · 47.7k in / 1.3k out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/45dd08616544ea4c970376974ca86ef68155add7...01c96be65eaf8a488c3f5ea0a3b2f3c40d0a2474) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/e774be4c67efe5a3722863f99f91f314beb9fa51/rubrics/documentation.md)</sub>

CBirkbeck: **Contesting this one, on the ground that `isMicrobial_iff` is not optional API — it is API that the `api-design` rubric required, and that `api-design` is green at this very head *because* it exists.**

First, what I am **not** claiming. I expected your suggested edit to be unimplementable and measured it before writing this. It is not: applying it exactly as written — deleting both lemmas, `obtain ⟨H, hnt, -⟩ := hv`, and building the existential directly — elaborates with no new errors. Your fix compiles, and I am not going to argue otherwise.

### The contradiction

Two rubrics require opposite things about the same declaration.

**`api-design`, round 4, head `dce2a173`** (2026-08-31T19:14:35Z) required it to exist:

> `IsMicrobial` exposes no `isMicrobial_iff`/`IsMicrobial_def` theorem, so consumers must unfold the predicate to obtain or construct its defining subgroup data. _Fix:_ Add a characteristic iff restating the existential condition, with the appropriate normal-form automation annotation.

**`reuse`, round 4, head `01c96be65`** (this thread) requires its deletion:

> _Fix:_ Delete `isMicrobial_iff_aux` and `isMicrobial_iff`, and unfold `IsMicrobial` directly in the two local proofs.

Both cannot hold. And this is not a stale finding I can ignore: on the scoreboard at **this** head, `api-design` is **green** — it went green precisely because the characteristic lemma was added in response to it. Deleting the lemma flips it back to blocking, and the PR cannot go all-green in either configuration.

### Why the premise "should not be retained as API" is the wrong half to keep

The seal and the lemma were installed together, by the same rubric, as a matched pair. **`api-design`, round 3, head `bc6f56f6`** removed `@[expose]` from `IsMicrobial` on exactly this reasoning:

> `IsMicrobial` is marked `@[expose]` even though `isMicrobial_iff` supplies its complete introduction/elimination interface. _Fix:_ Remove `@[expose]`; consumers can rewrite with `isMicrobial_iff` without depending on the definition's body.

So `IsMicrobial` is sealed **on the stated grounds that `isMicrobial_iff` is its interface.** Delete the lemma and that justification is gone: the predicate becomes a sealed `def` with no introduction or elimination form at all, and a downstream module can neither obtain the convex subgroup from it nor supply one to build it.

That the `obtain` form compiles does not answer this. It compiles because both consumers — `not_isMicrobial_of_subsingleton` and `isMicrobial_of_mulArchimedean` — live in the *defining* module, where the body is visible regardless of the seal. Being opaque **outside** that module is the whole point of the seal `api-design` asked for. So "unfold `IsMicrobial` directly in the two local proofs" works for those two proofs and for nothing else, which is what makes it a local convenience rather than a reason to remove the public API.

### What I propose

Keep `isMicrobial_iff`. If `reuse` still judges the `Iff.rfl` restatement redundant, then the coherent alternative is the *other* configuration — restore `@[expose]` on `IsMicrobial` and delete the lemma, so the predicate's body is the interface — but that directly reverses `api-design` round 3, and I am not going to make that change on one rubric's say-so while the other is green at this head.

The one thing I have already fixed, because it is independent of this dispute and correct either way, is the `documentation` finding: the result-oriented docstring now sits on the public theorem rather than on the private helper (`41c284427`).

TauCeti/Algebra/Order/Group/ConvexSubgroup.lean
CBirkbeck: **Re: your reply on `reuse` —** re-reviewed on `41c2844`; this clears the finding ✅ — approved on `41c2844`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3902252767.</sub>

PR #5485, opened 2026-09-02
refactor(DenseGraphLimits): index the cut-norm suprema once, privately, by bundled pairs

cameronfreer:
`cutNormSet` and `cutNormSigned` are suprema indexed by sets and test functions *together with* proof arguments — four binders and six. Those are the roadmap's displayed formulas and they stay. But every argument about either norm had to descend through every binder: nine private "strip one level" lemmas supplied boundedness at each depth, `cutNorm_le` was four nested `Real.iSup_le`, and `abs_testIntegral_le_cutNormSigned` was five nested `le_ciSup_of_le`.

This adds, privately, one supremum per norm over a bundled index — `MeasurableSubset Ω × MeasurableSubset Ω` and `SignedTest Ω × SignedTest Ω` — proves each equal to the public definition once, and routes every consumer through the flat form.

```lean
private theorem cutNormSet_eq_iSup_pair (K) : cutNormSet μ K = ⨆ ST, setIntegrand μ K ST
private theorem cutNormSigned_eq_iSup_pair (K) : cutNormSigned μ K = ⨆ uv, signedIntegrand μ K uv
```

The bridges are `ciSup_prod` plus one `ciSup_subtype` collapse per proof binder, with the `L¹` bound supplying boundedness throughout. `SignedTest` is a two-level subtype rather than one conjunction so that each of the two proof binders per test function collapses by the same step.

## Measured

| | before | after |
|---|---|---|
| private supremum-ladder lemmas | 9 | 2 (one `BddAbove` witness per norm) |
| `le_ciSup_of_le` applications | 8 | 3 |
| `Real.iSup_le` applications | 18 | 7 |
| `cutNorm_le`, `abs_rectIntegral_le_cutNorm`, `cutNorm_nonneg` | 4-deep descents | one step each |
| `cutNormSigned_le`, `abs_testIntegral_le_cutNormSigned`, `cutNormSigned_nonneg` | 6-deep / 5-deep | one step each |
| net lines | | **+16** |

The last row is the honest cost. The two bridges are about as long as the ladders they replace: the signed one still discharges one boundedness and one nonnegativity condition per collapsed binder. What changes is that this is paid **once**, in one place, instead of by every consumer — and any future lemma about either norm is a one-step argument rather than a fresh descent.

A compaction was tried and rejected: an ℝ-specialised private `ciSup_subtype` taking the two bounds as functions makes each collapse one line, but with `f` implicit it sends unification into a deterministic heartbeat timeout at both bridges, and `maxHeartbeats` overrides are not permitted here. It saved one line.

## What does not change

No public statement, name or attribute. The definitions are exactly the roadmap's. `integral_abs_nonneg` and `bddAbove_range_of_forall_le`, two private helpers that existed only to feed the ladders, go with them.

## Roadmap

Roadmap: DenseGraphLimits

Improvement to existing code; no roadmap target is claimed.

🤖 Directed by Cameron Freer, drafted by Fable 5.1

### DIFF
--- a/TauCeti/Combinatorics/DenseGraphLimits/Kernel/CutNorm.lean
+++ b/TauCeti/Combinatorics/DenseGraphLimits/Kernel/CutNorm.lean
@@ -115,10 +115,48 @@ theorem cutNormSet_def (K : SymmKernel Ω μ) :
       ⨆ (S : Set Ω) (_ : MeasurableSet S) (T : Set Ω) (_ : MeasurableSet T),
         |K.rectIntegral μ S T| := (rfl)
 
-omit [IsFiniteMeasure μ] in
-private theorem integral_abs_nonneg (K : SymmKernel Ω μ) :
-    0 ≤ ∫ p, |K p.1 p.2| ∂(μ.prod μ) :=
-  integral_nonneg fun _ => abs_nonneg _
+/-! ### The bundled normal forms
+
+The public definitions index their suprema by sets and test functions *together with* proof
+arguments, which is the roadmap's displayed formula and stays. Internally, every argument about
+either supremum goes through a single supremum over a bundled index, so that introduction and
+elimination are one-step `le_ciSup_of_le` / `Real.iSup_le` applications instead of a descent
+through four or six binders. Nothing here is public. -/
+
+/-- A measurable set, bundled with its measurability. -/
+private abbrev MeasurableSubset (Ω : Type*) [MeasurableSpace Ω] := {S : Set Ω // MeasurableSet S}
+
+/-- The integrand of `cutNormSet` on a bundled pair. -/
+private noncomputable def setIntegrand (K : SymmKernel Ω μ)
+    (ST : MeasurableSubset Ω × MeasurableSubset Ω) : ℝ :=
+  |K.rectIntegral μ ST.1.1 ST.2.1|
+
+private theorem setIntegrand_le (K : SymmKernel Ω μ)
+    (ST : MeasurableSubset Ω × MeasurableSubset Ω) :
+    setIntegrand μ K ST ≤ ∫ p, |K p.1 p.2| ∂(μ.prod μ) :=
+  K.abs_rectIntegral_le_integral_abs μ _ _
+
+private theorem bddAbove_setIntegrand (K : SymmKernel Ω μ) :
+    BddAbove (range (setIntegrand μ K)) :=
+  ⟨_, by rintro _ ⟨ST, rfl⟩; exact setIntegrand_le μ K ST⟩
+
+/-- `cutNormSet` as one supremum over bundled pairs. -/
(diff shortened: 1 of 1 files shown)


cameronfreer:
Converted to draft pending a decision. Measured: the nine supremum-ladder lemmas are gone and every consumer is one step, but the net is **+16 lines** — the two private bridges cost about what the ladders did. That does not meet the stated ship threshold (substantial net deletion), so this should not be adjudicated by review; elaboration timing is being measured and the ledger will be put to Cameron.

cameronfreer:
Closing; discarded on measurement. The full ledger, so the experiment is not repeated:

**What it did.** Replaced the four- and six-binder suprema of `cutNormSet` and `cutNormSigned` by one private supremum per norm over a bundled pair (`MeasurableSubset Ω × MeasurableSubset Ω`, and a two-level `SignedTest Ω × SignedTest Ω`), proved each equal to the roadmap-pinned public definition, and routed every consumer through the flat form. Public statements unchanged.

| | before | after |
|---|---|---|
| private supremum-ladder lemmas | 9 | 2 |
| `le_ciSup_of_le` applications | 8 | 3 |
| `Real.iSup_le` applications | 18 | 7 |
| public consumers (`cutNorm_le`, `abs_rectIntegral_le_cutNorm`, `cutNorm_nonneg` and the three signed counterparts) | 4-to-6-deep descents | one step each |
| elaboration of `Kernel/CutNorm.lean` (`lake build` of the module) | 6.75 s | 6.65 s |
| **net lines** | | **+16** |

**Why it is discarded.** The two private bridges back to the raw-binder formulas cost about what the nine ladders did: the signed bridge discharges one boundedness and one nonnegativity side condition per collapsed binder. The descent is paid once instead of per consumer, but the file has six consumers already written, so the saving is hypothetical while the bridge cost is permanent. `TauCetiRoadmap/DenseGraphLimits/Suggested.lean` pins the raw-binder forms as *definitions*, so the bridge is required under both "bundle the definitions and state the raw forms as theorems" and "keep the definitions and bridge privately" — the cost is structural, not a choice of implementation.

**Compaction tried and rejected.** An ℝ-specialised private `ciSup_subtype` wrapper taking the two bounds as functions (`∀ i, f i ≤ C` and `∀ i, 0 ≤ f i`) makes each collapse one line, but with `f` and `p` implicit it sends unification into a deterministic heartbeat timeout at both bridges in this nested setting, while the explicit `ciSup_subtype ⟨C, by …⟩ (by …)` form is cheap. `maxHeartbeats` overrides are not permitted. It saved one line.

**What does transfer.** `ciSup_prod` needs only `BddAbove (range f)`; `ciSup_subtype` needs that plus `sSup ∅ ≤ iSup f`, which in `ℝ` is `Real.sSup_empty` followed by `Real.iSup_nonneg`. Each raw proof binder collapses in one such step. Anyone needing a flat index for a *new* argument about these norms can build the bridge from those two lemmas in about thirty lines; it is not worth carrying in the file for the consumers that exist.

PR #5600, opened 2026-09-04
feat(MordellWeil): the canonical Néron–Tate height by Tate's telescoping limit

CBirkbeck:
Roadmap: EllipticCurves

## The target

`TauCetiRoadmap/EllipticCurves/README.md:805-811` names the canonical (Néron–Tate) height as a
Layer 6 milestone, separate from and later than Mordell–Weil, "consumed by regulators, isogeny
compatibility, and BSD". `STATUS.md:36` lists it as open. This PR is the **first item** on that
list: the construction, its defining convergence, and its bounded difference from the naïve height.

## What it adds

```lean
noncomputable def Point.canonicalHeight (P : W.Point) : ℝ :=
  limUnder atTop fun n : ℕ ↦ ((2 ^ n) • P).naiveHeight / (2 * 4 ^ n)

theorem Point.tendsto_naiveHeight_two_pow_nsmul_div_four_pow [W.toAffine.IsElliptic] (P : W.Point) :
    Tendsto (fun n : ℕ ↦ ((2 ^ n) • P).naiveHeight / (2 * 4 ^ n)) atTop (𝓝 P.canonicalHeight)

@[simp] theorem Point.canonicalHeight_zero [W.toAffine.IsElliptic] :
    (0 : W.Point).canonicalHeight = 0

@[simp] theorem Point.canonicalHeight_neg [W.toAffine.IsElliptic] (P : W.Point) :
    (-P).canonicalHeight = P.canonicalHeight

theorem Point.abs_canonicalHeight_sub_naiveHeight_le [W.toAffine.IsElliptic] :
    ∃ D, ∀ P : W.Point, |P.canonicalHeight - P.naiveHeight / 2| ≤ D
```

## The normalisation, since it was wrong in the first version

`h` here is the height of the `x`-coordinate, and `x` has a **double** pole at the point at
infinity — so `h_x` is the height attached to the divisor `2(O)`. Heights scale linearly in the
divisor, so the height attached to `(O)`, which is the one the Néron–Tate pairing, the regulator
and BSD are stated with, is **half** of `lim 4⁻ⁿ h(2ⁿ P)`. The first version of this PR omitted
that factor and so defined twice the conventional height; every invariant derived from it later
would have been scaled wrongly. `correctness` caught it and the definition now carries the `2`.
The module docstring records the divisor argument so the next reader need not rederive it.

## How it works

- **The doubling bound.** `approx_parallelogram_law` at `Q = P`, where `P - Q = 0` and `h 0 = 0`,
[...]

PR #5708, opened 2026-09-05
feat(FunctionField): evaluate a function on a divisor, for Weil reciprocity

CBirkbeck:
Roadmap: EllipticCurves

## The target

`TauCetiRoadmap/EllipticCurves/README.md` §Layer 2, "The Weil pairing — the divisor construction",
lists the pairing's prerequisites as named milestones:

> Its named prerequisites, each a milestone: existence of a function with divisor `N(P) − N(O)`;
> moving a divisor within its class to obtain disjoint support; **evaluation of a function on a
> divisor of disjoint support**; **Weil reciprocity** `f(div g) = g(div f)`; and independence of
> every function and divisor choice.

This PR is the third of those — the quantity Weil reciprocity compares.

## What it adds

Two modules, split along the line between place-local and divisor material.

`FieldTheory/FunctionField/Place/Residue.lean` — the local factors, generic place API:

```lean
noncomputable def Place.residueUnit (P : Place k F) (f : Fˣ) (hf : P.ord (f : F) = 0) :
    P.ResidueFieldˣ
noncomputable def Place.normResidue (P : Place k F) (f : Fˣ) (hf : P.ord (f : F) = 0) : kˣ
noncomputable def Place.normResidueOrOne (P : Place k F) (f : Fˣ) : kˣ
```

### The order hypotheses are derived, not demanded

`residueUnit f hf` carries its admissibility proof as an argument, so a lemma about `f * g`,
`f⁻¹` or `1` has to name a proof for the *composite*. An earlier revision made that the caller's
problem — `residueUnit_one` and `normResidue_one` took a hypothesis `P.ord ((1 : Fˣ) : F) = 0`
that is simply true, and the `mul`/`inv` lemmas asked for the composite's order alongside the
factors'. Three public lemmas now discharge all of it:

```lean
theorem ord_mul_eq_zero (hf : P.ord (f : F) = 0) (hg : P.ord (g : F) = 0) :
    P.ord ((f * g : Fˣ) : F) = 0
theorem ord_inv_eq_zero (hf : P.ord (f : F) = 0) : P.ord ((f⁻¹ : Fˣ) : F) = 0
theorem ord_div_eq_zero (hf : P.ord (f : F) = 0) (hg : P.ord (g : F) = 0) :
    P.ord ((f / g : Fˣ) : F) = 0
```

There is no `one` member: `((1 : Fˣ) : F)` is `1` definitionally, so `TauCeti.Place.ord_one`
already has the type the identity law needs.

so `residueUnit_one` and `normResidue_one` are hypothesis-free and `@[simp]`, and the residue and
norm lemmas state only what is genuinely assumed — the orders of the *inputs*.

They have to be named declarations rather than inlined tactic proofs, because `residueUnit`'s
admissibility argument is *part of the left-hand side*: `residueUnit_mul`'s statement is about
`P.residueUnit (f * g) (ord_mul_eq_zero hf hg)`, and a tactic block in that position would
make the statement unreadable and leave a caller nothing to state the composite's admissibility
with.

They do not re-derive the group laws, though. `TauCeti.Place.ordAddMonoidHom` already bundles
`ord_P` as an additive homomorphism on `Additive Fˣ`, so all four are `map_add`, `map_neg`,
`map_sub` and `map_zero` read through `ordAddMonoidHom_apply`. That homomorphism was defined in
`Divisor/Principal.lean` even though it is pure `Place` API mentioning no divisor, and nothing
outside that file used it; this PR **moves it to `Place/Basic.lean`**, next to the `ord_one`,
`ord_mul`, `ord_inv` and `ord_div` it bundles — and the four closure lemmas land there too, since
they mention no residue field and would otherwise make generic order API depend on the
residue/norm layer. `Divisor.eval`'s own `IsUnitAtSupport.mul` now calls `ord_mul_eq_zero`
instead of repeating the same calculation a third time.

`mem_unitGroup_of_ord_eq_zero` is `private`: its only consumer is the private `unitGroupMk`, so it
is implementation, not interface. The detour it exists to avoid is still recorded below.

The relocation costs one adjustment in `Principal.lean`: `orderSystem.finite_support` closed its
goal by definitional unfolding of `ordAddMonoidHom`, which worked only while the two were in the
same module. It now goes through `ordAddMonoidHom_apply`.

The quotient laws — `ord_div_eq_zero`, `residueUnit_div`, `normResidue_div` and
`normResidueOrOne_div` — complete the row, matching `Divisor.eval_div` and `IsUnitAtSupport.div`
on the divisor side.

with `normResidueOrOne_of_ord_eq_zero` and `normResidueOrOne_of_ord_ne_zero`, the two `@[simp]`
branch rules; `ord_mul_eq_zero`, `residueUnit_mul`, `normResidue_mul`,
`normResidueOrOne_mul`, `normResidueOrOne_one`, `normResidue_one` and `normResidueOrOne_inv`, the
local group laws in the function variable; and
`coe_residueUnit` / `coe_normResidue`, the `@[simp]` field values of the two units, named after the adjacent `TauCeti.Algebra.coe_normUnits` — so a consumer relates them to
`IsLocalRing.residue` and `Algebra.norm` without unfolding either definition.

Neither value lemma is `rfl`, and that is not a stylistic choice: an **exported** `rfl` theorem
requires every definition it unfolds to be `@[expose]`d —

```
Not a definitional equality: the left-hand side
  ↑(P.residueUnit f hf)
is not definitionally equal to the right-hand side
  (IsLocalRing.residue ↥P.integers) ⟨↑f, ⋯⟩
Note: This theorem is exported from the current module. This requires that all definitions that
need to be unfolded to prove this theorem must be exposed.
```

— whereas a tactic proof (`simp [residueUnit]`) exports a proof term instead of a defeq
obligation. So the bodies stay unexposed and the values are still named. Checked downstream: a
probe module importing only `Divisor/Eval.lean` closes both value goals by `simp` alone.

`FieldTheory/FunctionField/Divisor/Eval.lean` — the product over a divisor:

```lean
noncomputable def Divisor.eval (D : Divisor k F) (f : Fˣ) : kˣ
def Divisor.IsUnitAtSupport (D : Divisor k F) (f : Fˣ) : Prop
```

`eval` is `MonoidHom`-backed, but that homomorphism is **private**: `eval_zero`/`eval_add`/
`eval_neg`/`eval_sub` being hypothesis-free is the whole of what it buys a consumer, so the bundled
map is not public surface. It can be made public when something outside the module needs it.

with `eval_zero`, `eval_add`, `eval_neg`, `eval_sub`, `eval_single`, `eval_eq_finsuppProd`,
`eval_eq_prod_normResidue`, `isUnitAtSupport_iff` and `isUnitAtSupport_iff_disjoint`.
`eval_zero`, `eval_add`, `eval_neg`, `eval_sub`, `eval_single` and `isUnitAtSupport_iff` are all
`@[simp]`; `eval_eq_finsuppProd` deliberately is not (below).

## The other half of the bilinearity

`evalHom` is a homomorphism in the *divisor*, so `eval_add`/`eval_neg`/`eval_sub`/`eval_zero` are
hypothesis-free. The function variable is not automatic, and the group laws there split into two
kinds — which is the mathematical content, not boilerplate:

```lean
@[simp] theorem eval_one (D : Divisor k F) : eval D (1 : Fˣ) = 1
@[simp] theorem eval_inv (D : Divisor k F) (f : Fˣ) : eval D f⁻¹ = (eval D f)⁻¹

theorem eval_mul (hf : IsUnitAtSupport D f) (hg : IsUnitAtSupport D g) :
    eval D (f * g) = eval D f * eval D g
theorem eval_div (hf : IsUnitAtSupport D f) (hg : IsUnitAtSupport D g) :
    eval D (f / g) = eval D f / eval D g
```

**`eval_one` and `eval_inv` need no hypothesis.** `ord_P f⁻¹ = -ord_P f` vanishes exactly when
`ord_P f` does, so admissibility is *invariant* under inversion, and at the places where `f` is
not a unit both sides are `1`. `isUnitAtSupport_inv_iff` records the invariance as an iff.

**`eval_mul` and `eval_div` genuinely need both.** At a place where `f` and `g` have opposite
nonzero orders their product is a unit while neither factor is, so the left side sees a real norm
there and the right side sees `1` twice. Admissibility is a subgroup condition in **each** variable, which is what makes the laws
composable:

| | zero/one | closure | invariance | inverse/difference |
|---|---|---|---|---|
| function | `isUnitAtSupport_one` | `IsUnitAtSupport.mul` | `isUnitAtSupport_inv_iff` | `IsUnitAtSupport.div` |
| divisor | `isUnitAtSupport_zero` | `IsUnitAtSupport.add` | `isUnitAtSupport_neg_iff` | `IsUnitAtSupport.sub` |

The function row is what Weil reciprocity and the independence-of-choices milestone (comparing two
functions with the same divisor) are stated against; the divisor row is what moving a divisor
within its class to obtain disjoint support needs. Both are named roadmap prerequisites of
III.8.1.

`IsUnitAtSupport.add` takes its support inclusion from `Finsupp.support_add`, under a local
`classical` — that supplies the `DecidableEq (Place k F)` the union mentions without putting it in
the statement.

### `eval_zsmul`, and the partial laws are `@[simp]`

`eval` is a homomorphism in the divisor, so `eval_zsmul : eval (n • D) f = eval D f ^ n` comes
from `map_zpow` — and since `evalHom` is private, a consumer cannot get it any other way. It is
the law the `N(P) − N(O)` divisors of the Weil pairing are evaluated through.

The eight partial group laws (`residueUnit_mul`, `_inv`, `_div`, their `normResidue` counterparts,
and `normResidueOrOne_mul` / `_div`) all carry `@[simp]`. That is not obvious for a lemma whose
left-hand side contains a dependent admissibility proof, so it was checked rather than assumed:
with all eight tagged, `#lint in TauCeti` over the full linter set reports "All linting checks
passed", 0 errors in 644 declarations — no redundancy and no simp-normal-form violation.

`eval_mul` and `eval_div` are the exception, and deliberately so. A private round asked twice for
`@[simp]` on them by analogy with the local laws; `#lint` refuses it. Their hypothesis
`IsUnitAtSupport D f` is itself rewritten by the `@[simp]` `isUnitAtSupport_iff`, so simp would
never be able to apply the rule — the linter says exactly that: *"Left-hand side does not simplify,
when using the simp lemma on itself … The simp lemma may be invalid because hypothesis hf
simplifies"*. The eight local laws pass because their hypothesis `P.ord f = 0` is already a simp
normal form. The reason is recorded above `eval_mul` in the file.

### Two further simp decisions, linter-verified rather than assumed

`normResidueOrOne_one` is **not** `@[simp]`: `normResidueOrOne_of_ord_eq_zero` is, and `simp`
discharges `ord_P 1 = 0` on its own, so the total form is rewritten to `normResidue` before the
`one` rule could fire. `normResidue_one` is the `@[simp]` rule for that normal form. This surfaced
as a real failure — `simp [eval_eq_finsuppProd]` left `∏ normResidue 1 ⋯ ^ n = 1` open — not as a
guess.

The four "invariance" closure lemmas (`isUnitAtSupport_one`, `isUnitAtSupport_inv_iff`,
`isUnitAtSupport_zero`, `isUnitAtSupport_neg_iff`) are **not** `@[simp]` either: with
`isUnitAtSupport_iff` tagged, `simp` proves both outright, and the environment linter reports them
as redundant simp lemmas when they carry the attribute (`simp can prove this`). They stay as named
lemmas because they are what a term-mode proof reaches for.

### On the size of this PR

The module grew over four `api-design` rounds, each asking for the next slice of the surrounding
API: the unit value lemmas, then multiplicativity in the function, then the remaining group laws
there, then the divisor-side closure. Every request was correct and none contradicted an earlier
one, so all were implemented rather than contested — and the result is a genuinely complete API
rather than the minimum that compiles. Both closure directions are now saturated — zero/one, product, inverse, difference **and integer
power** in each variable:

| | zero/one | product | inverse | difference | power |
|---|---|---|---|---|---|
| function | `isUnitAtSupport_one` | `IsUnitAtSupport.mul` | `isUnitAtSupport_inv_iff` | `.div` | `.zpow` |
| divisor | `isUnitAtSupport_zero` | `.add` | `isUnitAtSupport_neg_iff` | `.sub` | `.zsmul` |

with `eval_zpow` and `eval_zsmul` the two evaluation laws, and `normResidueOrOne_zpow` the local
one. Like `normResidueOrOne_inv`, the power laws need **no** admissibility hypothesis: where `f`
is not a unit neither is any nonzero power of it, so both sides are `1`.

`f(D) = ∏_{P ∈ supp D} N_{F_P/k}(f(P)) ^ (coeff D P)`. The norm is not decoration: `f(P)` lives in
the residue field `F_P`, which varies with `P`, so without pushing each value down to `k` the
factors would live in different fields and the product would not typecheck.

## Three load-bearing design choices

**Functions in `Fˣ`, not `F`.** The admissibility condition is `P.ord f = 0`, and `P.ord 0 = 0` by
the junk-value convention on `ord` — so over `F` the condition would declare `f = 0` admissible at
*every* place. `Divisor.principal` is already stated for `Fˣ`, so this also matches the API it will
be used with.

**Values in `kˣ` with neutral value `1`**, not in `k` with `0`. Coefficients are integers, so each
local factor is raised to a possibly negative power and the divisor law has to survive that. With
`0` in `k` it does not: at a place where `f` is not a unit,
`eval (single P 1) * eval (single P (-1))` would be `0` while `eval 0 = 1`. In `kˣ` every `zpow`
identity is unconditional and the law is automatic.

**A total definition, with admissibility on the theorems.** Making the hypothesis an argument would
put a proof term inside `f(D)`, to be transported through every rewrite of `D` or `f`, and would
block packaging `f(-)` as a homomorphism at all. Instead `normResidueOrOne` is total, `evalHom` is a
`MonoidHom` out of `Multiplicative (Divisor k F)`, so `eval_zero`/`eval_add`/`eval_neg`/`eval_sub`
are hypothesis-free, and `eval_eq_prod_normResidue` recovers the textbook product exactly where
admissibility holds.

## Reuse

Everything rests on existing API rather than new machinery:

- `TauCeti.freeAbelianCharEquiv` (`Algebra/Group/FreeAbelianCharacter.lean`) supplies
  multiplicativity: a divisor is a finitely supported integer combination of places, so a
  homomorphism out of it is exactly a choice of value per place. Its `@[simp]`
  `freeAbelianCharEquiv_symm_apply_ofAdd` gives the Finsupp product directly.
- `ValuationSubring.unitGroupToResidueFieldUnits` is Mathlib's residue map on unit groups, so
  `residueUnit` is that at the unit-group element `f` names — not a residue paired with a separate
  nonzero proof. **This is what makes the whole group law free**: `residueUnit_one`/`_mul`/`_inv`
  are `map_one`/`map_mul`/`map_inv`.
- Landing in the unit group needs one detour worth recording. `Valuation.mem_unitGroup_iff` would
  ask unification to see `P.integers` as `P.valuation.valuationSubring`, and **`Place.integers` is
  not an exposed definition** — Lean says so outright, `integers ↦ 105` — so that unfolding is
  unavailable outside `Place/Basic.lean` and the attempt exhausts the heartbeat budget at `whnf`.
  `mem_unitGroup_of_ord_eq_zero` therefore routes through `IsUnit`, via
  `ValuationSubring.mem_unitGroup_iff` and `ValuationSubring.valuation_eq_one_iff`, both generic in
  the subring. No `@[expose]` and no new lemma in `Place/Basic.lean` were needed.
- `TauCeti.Algebra.normUnits` (`RingTheory/Norm/Units.lean`) is the norm as a homomorphism of unit
  groups, so `normResidue` is that applied to the residue rather than a hand-rolled
  `Units.map (Algebra.norm k)`, and `normResidue_mul` is `map_mul`. Its docstring also states the
  finiteness policy argued for below — "No finiteness hypothesis is needed here, because
  `Algebra.norm` itself has none … A consumer that cares about the value supplies its own
  finiteness where the value is computed" — so this module cites it rather than restating it.
- `mem_support_principal_iff` gives `isUnitAtSupport_iff_disjoint`.
- No thin specialisations are added: an earlier revision carried
  `mem_integers_of_ord_eq_zero` (a special case of `mem_integers_iff_ord_nonneg`, now inlined as
  `hf.ge`) and `normResidue_eq_one_of_not_finite` (a `kˣ`-valued restatement of
  `Algebra.norm_eq_one_of_not_module_finite`, now cited from the docstring instead). Both are gone.

**On finiteness.** An earlier version of this description claimed no finiteness hypothesis was
needed, because `Units.map` sends units to units for any `MonoidHom`. That is true of
well-typedness and false of meaning, and `correctness` was right to block it: `Algebra.norm` is
`LinearMap.det ∘ lmul`, so on a residue field that is not module-finite over `k` it is the junk
value `1` (`Algebra.norm_eq_one_of_not_module_finite`), and `eval` would silently drop that factor.

The norm is therefore the classical field norm *exactly* when `Module.Finite k P.ResidueField`,
which `Place.finiteDimensional_residueField` supplies at every place of a function field — so the
junk regime never arises where this API is used. Both the module docstring and `normResidue`'s own
docstring say this and name the Mathlib lemma that bounds the junk regime.

The hypothesis is deliberately *not* in the signatures: `Algebra.norm` consumes no finiteness
argument, so it would be unused in every one of `normResidue`, `normResidueOrOne`, `evalHom` and
`eval`, and the `unusedArguments` linter rejects unused instance arguments — that linter failed CI
on #5600 for exactly this pattern. This mirrors `TauCeti.Place.degree`, an un-hypothesised
definition in this same development whose docstring likewise records that finiteness "guards the
junk value of `Module.finrank`".

`IsFunctionField` appears only in `isUnitAtSupport_iff_disjoint`, where `principal` requires it.

`eval_eq_finsuppProd` is deliberately **not** `@[simp]`, for the reason already recorded on
`WeierstrassCurve.Affine.Point.naiveHeight_eq_logHeight`: tagging a defining equation dissolves
`eval` on sight and puts the structural lemmas out of simp-normal form. The `#lint` environment
linter reports exactly `eval_zero`, `eval_neg` and `eval_single` when it does carry `@[simp]` —
that is how the decision was checked, not assumed. The rationale lives in a comment above the
theorem, not in its docstring.

## The subtype representation appears once

`residueUnit` is `unitGroupToResidueFieldUnits` at an element of `𝒪_P`'s unit group, and that
group is a `Subgroup Fˣ`, so its operations are computed on values. An earlier revision let each
group law close on a trailing `rfl` that depended on that representation — five of them, none
saying so. The construction is now a private `unitGroupMk` with `unitGroupMk_mul`, `_one`, `_inv`,
`_zpow` and one coercion lemma, all `rfl`, and every `residueUnit` and `normResidue` law is a
`map_*` applied to one of them — except the quotient laws, which compose a product with an
inverse and need no representation lemma at all. One `rfl` remains, in `coe_residueUnit`, and it unfolds
`IsLocalRing.residue` to `Ideal.Quotient.mk` — its definition, not a representation fact — with a
comment saying so.

## `IsUnitAtSupport` is not exposed

The predicate is an ordinary opaque `def`, and `isUnitAtSupport_iff` (`Iff.rfl`) is its whole
interface: introduction and elimination in one lemma. `eval_eq_prod_normResidue` therefore states
its membership proofs as `isUnitAtSupport_iff.1 h P.1 P.2` rather than applying `h` directly, which
is what previously forced `@[expose]`. Checked with a downstream probe module rather than assumed —
a separate file importing only `Divisor/Eval.lean` applies the product formula, introduces the
predicate from the pointwise condition, and eliminates it, all with the body unexposed.

## Placement

The residue-and-norm layer mentions no divisor: it is generic place API, so it lives in
`FieldTheory/FunctionField/Place/Residue.lean` next to `Basic`, `Degree`, `Zeros` and the rest, and
a consumer who wants the local norm no longer has to import `Divisor/Principal.lean` to reach it.
`Divisor/Eval.lean` imports it and keeps only the divisor material, alongside `Basic`, `Principal`,
`ProductFormula`, `Conorm` and `AffineModel`.

There is no elliptic-curve content in either module: this is general function-field material,
motivated by the elliptic roadmap (hence the `Roadmap:` line) but stated for any `F/k`. The
elliptic function field reaches it through the existing `WeierstrassCurve.Affine.isFunctionField`.

## Provenance

**Original work**, following the classical construction the roadmap names: Silverman, *The
Arithmetic of Elliptic Curves*, III.8 (the divisor construction of the Weil pairing, III.8.1), now
cited in both module docstrings under `## References` via the `[silverman2009]` bibliography key.

Absence checked before writing: `grep -rniE 'weilReciprocity|weil_reciprocity'` over TauCeti and
the pinned Mathlib returns nothing (only quadratic, Artin and Frobenius reciprocity, in unrelated
developments), and `evalDivisor|eval_divisor|divisorEval` returns nothing in either.

## Gates

- `lake build`: green, 0 errors, 0 warnings under `warningAsError = true`, no Mathlib recompiled.
- `scripts/lint-style.sh`: exit 0, all 4324 headers conforming; no line over 100 characters.
- **`#lint in TauCeti` — the full environment-linter set, not just simpNF**: "All linting checks
  passed", 0 errors in 633 declarations. The narrower simpNF-only probe is what let an
  `unusedArguments` violation reach CI on #5600, so this runs the full set; it is also what
  confirmed every `@[simp]` decision in the two modules, in both directions: which tags are safe
  and which two would be redundant.
- `scripts/lint-dot-notation.py`: 0 new.
- No `sorry`, no `native_decide`, no `maxHeartbeats`; 2 files, both new (new-file guard is 1000
  additions each).
- Claim at `refs/tauceti-claims/author/EllipticCurves/divisor-eval`; no open PR carries an
  overlapping marker and none touches `FunctionField/Divisor/` or `FunctionField/Place/`.

### DIFF
--- a/TauCeti/FieldTheory/FunctionField/Place/Residue.lean
+++ b/TauCeti/FieldTheory/FunctionField/Place/Residue.lean
@@ -110,10 +110,10 @@ This is the classical field norm precisely when `Module.Finite k P.ResidueField`
 `TauCeti.Place.finiteDimensional_residueField` supplies for every place of a function field.
 Absent that, `Algebra.norm` is the junk value `1`
 (`Algebra.norm_eq_one_of_not_module_finite`), exactly as `TauCeti.Place.degree` is junk `0`
-absent the same hypothesis. The hypothesis is not in the signature because `Algebra.norm` does
-not consume one, so requiring it here would leave it unused — the policy
-`TauCeti.Algebra.normUnits` states for itself, and this is that homomorphism applied to the
-residue. -/
+absent the same hypothesis. -/
+-- Finiteness is documented rather than assumed: `Algebra.normUnits` consumes no such argument, so
+-- a hypothesis here would be unused, which `unusedArguments` rejects. `TauCeti.Place.degree`
+-- records the same convention for `Module.finrank`.
 noncomputable def normResidue (P : Place k F) (f : Fˣ) (hf : P.ord (f : F) = 0) : kˣ :=
   Algebra.normUnits k (P.residueUnit f hf)
 
--- a/TauCeti/FieldTheory/FunctionField/Place/Residue.lean
+++ b/TauCeti/FieldTheory/FunctionField/Place/Residue.lean
@@ -41,10 +41,8 @@ not consume one — so the guarantee is recorded here rather than in the signatu
 ## Main results
 
 * `TauCeti.Place.coe_residueUnit` and `TauCeti.Place.coe_normResidue`: the underlying field values
-  of the two units, as `@[simp]` normal forms, named after `TauCeti.Algebra.coe_normUnits` — the
-  adjacent lemma of exactly this shape. They are the whole reason a consumer never has to unfold
-  either definition: they name the `IsLocalRing.residue` and `Algebra.norm` the units are built
-  from.
+  of the two units: `IsLocalRing.residue` and the `Algebra.norm` of it. Both are `@[simp]`, and
+  are named after `TauCeti.Algebra.coe_normUnits`, the adjacent lemma of the same shape.
 * `TauCeti.Place.normResidueOrOne_of_ord_eq_zero` and
   `TauCeti.Place.normResidueOrOne_of_ord_ne_zero`: the two branches of the total extension.
 * The group laws in the function: `residueUnit_one`, `residueUnit_mul`, `residueUnit_inv` and
--- a/TauCeti/FieldTheory/FunctionField/Place/Residue.lean
+++ b/TauCeti/FieldTheory/FunctionField/Place/Residue.lean
@@ -92,6 +92,7 @@ noncomputable def residueUnit (P : Place k F) (f : Fˣ) (hf : P.ord (f : F) = 0)
     P.ResidueFieldˣ :=
   P.integers.unitGroupToResidueFieldUnits ⟨f, P.mem_unitGroup_of_ord_eq_zero hf⟩
 
+/-- The residue field element underlying `residueUnit`: the residue of `f` in `𝒪_P / 𝔪_P`. -/
 @[simp]
 theorem coe_residueUnit (P : Place k F) (f : Fˣ) (hf : P.ord (f : F) = 0) :
     (P.residueUnit f hf : P.ResidueField)
@@ -114,6 +115,7 @@ residue. -/
 noncomputable def normResidue (P : Place k F) (f : Fˣ) (hf : P.ord (f : F) = 0) : kˣ :=
   Algebra.normUnits k (P.residueUnit f hf)
 
+/-- The element of `k` underlying `normResidue`: the norm of the residue. -/
 @[simp]
 theorem coe_normResidue (P : Place k F) (f : Fˣ) (hf : P.ord (f : F) = 0) :
     (P.normResidue f hf : k) = Algebra.norm k (P.residueUnit f hf : P.ResidueField) := by
@@ -126,11 +128,13 @@ survives that. -/
 noncomputable def normResidueOrOne (P : Place k F) (f : Fˣ) : kˣ :=
   if hf : P.ord (f : F) = 0 then P.normResidue f hf else 1
 
+/-- Where `f` is a unit at `P`, the total local factor is the genuine norm of the residue. -/
 @[simp]
 theorem normResidueOrOne_of_ord_eq_zero {P : Place k F} {f : Fˣ} (hf : P.ord (f : F) = 0) :
     P.normResidueOrOne f = P.normResidue f hf := by
   simp [normResidueOrOne, hf]
 
+/-- Where `f` is not a unit at `P`, the total local factor is the neutral value `1`. -/
 @[simp]
 theorem normResidueOrOne_of_ord_ne_zero {P : Place k F} {f : Fˣ} (hf : P.ord (f : F) ≠ 0) :
     P.normResidueOrOne f = 1 := by
--- a/TauCeti/FieldTheory/FunctionField/Place/Residue.lean
+++ b/TauCeti/FieldTheory/FunctionField/Place/Residue.lean
@@ -159,10 +159,6 @@ theorem normResidue_mul {P : Place k F} {f g : Fˣ} (hf : P.ord (f : F) = 0)
       = P.normResidue f hf * P.normResidue g hg := by
   rw [normResidue, normResidue, normResidue, residueUnit_mul hf hg, map_mul]
 
-/-- The order of the constant `1` is zero at every place, so `1` is admissible everywhere. -/
-theorem ord_one_units (P : Place k F) : P.ord ((1 : Fˣ) : F) = 0 := by
-  rw [Units.val_one, P.ord_one]
-
 /-- `ord_P f⁻¹ = -ord_P f`, so admissibility is invariant under inversion. -/
 theorem ord_inv_eq_zero {P : Place k F} {f : Fˣ} (hf : P.ord (f : F) = 0) :
     P.ord ((f⁻¹ : Fˣ) : F) = 0 := by
@@ -205,7 +201,8 @@ theorem normResidueOrOne_mul {P : Place k F} {f g : Fˣ} (hf : P.ord (f : F) = 0
 theorem normResidueOrOne_one (P : Place k F) : P.normResidueOrOne (1 : Fˣ) = 1 := by
   -- read off multiplicativity at `f = g = 1` rather than from the residue: `a = a * a` in a
   -- group forces `a = 1`, which avoids all of the subtype-coercion work
-  have h := normResidueOrOne_mul (P := P) (f := 1) (g := 1) P.ord_one_units P.ord_one_units
+  have h1 : P.ord ((1 : Fˣ) : F) = 0 := by simp
+  have h := normResidueOrOne_mul (P := P) (f := 1) (g := 1) h1 h1
   rw [one_mul] at h
   exact right_eq_mul.1 h
 
--- a/TauCeti/FieldTheory/FunctionField/Place/Residue.lean
+++ b/TauCeti/FieldTheory/FunctionField/Place/Residue.lean
@@ -46,6 +46,14 @@ not consume one — so the guarantee is recorded here rather than in the signatu
   built from.
 * `TauCeti.Place.normResidueOrOne_of_ord_eq_zero` and
   `TauCeti.Place.normResidueOrOne_of_ord_ne_zero`: the two branches of the total extension.
+* The group laws in the function: `residueUnit_one`, `residueUnit_mul`, `residueUnit_inv` and
+  their `normResidue` counterparts, together with `normResidueOrOne_one`, `normResidueOrOne_mul`
+  and `normResidueOrOne_inv` for the total form. Because `residueUnit` is the image of a
+  `MonoidHom`, the first three are `map_one`, `map_mul` and `map_inv`; the `normResidue` ones
+  compose that with `Algebra.normUnits`, itself a `MonoidHom`. Only the *total* form needs work,
+  and only for the product: `normResidueOrOne_mul` needs both factors admissible, while
+  `normResidueOrOne_inv` needs nothing, since `ord_P f⁻¹ = -ord_P f` vanishes exactly when
+  `ord_P f` does.
 
 ## References
 
@@ -59,21 +67,36 @@ namespace TauCeti.Place
 
 variable {k F : Type*} [Field k] [Field F] [Algebra k F]
 
+/-- A function of order zero at `P` lies in the unit group of `𝒪_P`: order zero is valuation one,
+which is exactly `ValuationSubring.mem_unitGroup_iff`. -/
+theorem mem_unitGroup_of_ord_eq_zero (P : Place k F) {f : Fˣ} (hf : P.ord (f : F) = 0) :
+    f ∈ P.integers.unitGroup :=
+  -- Routed through `IsUnit` rather than through `Valuation.mem_unitGroup_iff`, which would ask
+  -- unification to see `P.integers` as `P.valuation.valuationSubring`: `Place.integers` is not
+  -- an exposed definition, so that unfolding is unavailable here and the attempt exhausts the
+  -- heartbeat budget at `whnf`. Both lemmas used below are generic in the valuation subring.
+  (ValuationSubring.mem_unitGroup_iff _ f).2 <|
+    (ValuationSubring.valuation_eq_one_iff _ _).1
+      ((P.isUnit_iff_ord_eq_zero (x := ⟨(f : F), P.mem_integers_iff_ord_nonneg.2 hf.ge⟩)
+        (Units.ne_zero f)).2 hf)
+
 /-- **The residue `f(P)` of a function that is a unit at `P`**, as a unit of the residue field.
 Being a unit is what makes the residue nonzero, which is what lets it be raised to a negative
-power in `TauCeti.Divisor.eval`. -/
+power in `TauCeti.Divisor.eval`.
+
+This is Mathlib's `ValuationSubring.unitGroupToResidueFieldUnits` at the unit-group element `f`
+names, rather than a residue paired with a separate proof that it is nonzero. Being the image of
+a `MonoidHom` is what makes the group laws below `map_one`, `map_mul` and `map_inv`. -/
 noncomputable def residueUnit (P : Place k F) (f : Fˣ) (hf : P.ord (f : F) = 0) :
     P.ResidueFieldˣ :=
-  -- `f` lies in `𝒪_P` because `0 ≤ ord_P f`, and the coercion back to `F` of the element of
-  -- `𝒪_P` so named is `rfl`, which is why `hf` and `Units.ne_zero f` apply to it verbatim.
-  Units.mk0 (IsLocalRing.residue P.integers ⟨(f : F), P.mem_integers_iff_ord_nonneg.2 hf.ge⟩)
-    (by rw [Ne, P.residue_eq_zero_iff_ord_pos (Units.ne_zero f), hf]; omega)
+  P.integers.unitGroupToResidueFieldUnits ⟨f, P.mem_unitGroup_of_ord_eq_zero hf⟩
 
 @[simp]
 theorem val_residueUnit (P : Place k F) (f : Fˣ) (hf : P.ord (f : F) = 0) :
     (P.residueUnit f hf : P.ResidueField)
       = IsLocalRing.residue P.integers ⟨(f : F), P.mem_integers_iff_ord_nonneg.2 hf.ge⟩ := by
-  simp [residueUnit]
+  rw [residueUnit, ValuationSubring.coe_unitGroupToResidueFieldUnits_apply]
+  rfl
 
 /-- **The norm to `k` of the residue of a function that is a unit at `P`.** The residue field
 varies with `P`; the norm is what puts the value in `k`, the one field all the local factors of
@@ -124,8 +147,7 @@ theorem residueUnit_mul {P : Place k F} {f g : Fˣ} (hf : P.ord (f : F) = 0)
     (hg : P.ord (g : F) = 0) :
     P.residueUnit (f * g) (ord_mul_eq_zero hf hg)
       = P.residueUnit f hf * P.residueUnit g hg := by
-  ext
-  rw [Units.val_mul, val_residueUnit, val_residueUnit, val_residueUnit, ← map_mul]
+  rw [residueUnit, residueUnit, residueUnit, ← map_mul]
   rfl
 
 /-- **The norm of the residue is multiplicative in the function**, at a place where both factors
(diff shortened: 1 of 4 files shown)


CBirkbeck:
🔍 Review in progress — `codex` reviewing `74bccc39db1c`.

TauCeti/FieldTheory/FunctionField/Place/Residue.lean
CBirkbeck, line 1: ### ✅ reuse — now passing on `5560334`.

CBirkbeck:
## AI review — approved

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics).

|---|---|---|---|---|
| ✅ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/correctness.md) | approved | `codex/gpt-5.6-sol` | The definitions and theorem statements faithfully formalize function evaluation on an admissible divisor, including normed residue factors, disjoint-support admissibility, and the edge cases for inversion and integer powers. |
| ✅ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/reuse.md) | approved | `codex/gpt-5.6-sol` | The new evaluation API reuses the existing free-abelian character, residue-unit, norm-unit, and principal-divisor infrastructure. The earlier redundant `unitGroupMk_div` scaffolding has been removed, and the quotient law now composes the established multiplication and inversion laws. |
| ✅ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/scope.md) | approved | `codex/gpt-5.6-sol` | The PR directly advances the EllipticCurves Layer 2 named milestone for evaluating a function on a divisor of disjoint support. The local residue/norm machinery, divisor evaluation API, admissibility laws, and relocation of supporting order infrastructure form one coherent topic, with the preceding divisor-calculus and disjoint-support foundations already present. |
| ✅ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/attribution.md) | approved | `codex/gpt-5.6-sol` | Both new modules credit Silverman III.8 for the central classical construction, and the code identifies the existing TauCeti and Mathlib declarations it builds on. No uncredited copied or closely adapted formalization was found. |
| ✅ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/api-design.md) | approved | `codex/gpt-5.6-sol` | The canonical public surface is appropriately placed and complete: local residue/norm definitions and divisor evaluation have characteristic value, branch, algebraic, admissibility, and disjoint-support lemmas with suitable automation, while representation-specific helpers and the internal evaluation homomorphism remain private. |
| ✅ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/generality.md) | approved | `codex/gpt-5.6-sol` | The declarations are stated at the natural generality: formal place and divisor laws require only fields and an algebra, while `IsFunctionField` appears only for the principal-divisor characterization that needs it. No materially excessive assumptions, duplicated special cases, or unjustified abstraction were found. |
| ✅ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/placement.md) | approved | `codex/gpt-5.6-sol` | The new declarations are placed in the canonical Place and Divisor topic modules, the generic order API is correctly located in Place/Basic, and each added import has a direct use. |
| ✅ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/naming.md) | approved | `codex/gpt-5.6-sol` | The introduced names accurately describe their conclusions in standard Mathlib terminology, use appropriate namespaces, and add no notation. |
| ✅ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/documentation.md) | approved | `codex/gpt-5.6-sol` | The new modules and public declarations document their mathematical purpose, hypotheses, totalization convention, and finite-residue-field caveat accurately. |
| ✅ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/proof-quality.md) | approved | `codex/gpt-5.6-sol` | The proofs use stable homomorphism and finite-product APIs, with no undocumented `change`/`show` or substantial brittle rewrite chains. Representation-dependent definitional equalities are isolated in private helper lemmas and explicitly documented. |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/d6214142a5e6403ba4f977246f1fc16418d561e4...556033426a6e7c7895f6afd5fea1e9bf7a8cd7de) at head `5560334`. rubrics @ [`afb424e`](https://github.com/TauCetiProject/TauCetiReview/tree/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics). Review spend: $6.35.</sub>

CBirkbeck:
🔍 Review in progress — `codex` reviewing `556033426a6e`.

PR #5207, opened 2026-08-30
feat(FunctionField): the Riemann–Roch theorem

mrdouglasny:
Prepared by Claude Opus 5, reviewed by @mrdouglasny.

`Divisor.IsRiemannRochDivisor` already records the Riemann–Roch identity and everything that follows from it — that `g₀` is the genus, that `deg W = 2g - 2` and `ℓ(W) = g`, that any two such divisors are linearly equivalent. Its own docstring names what is missing: *"the Riemann–Roch theorem itself is the assertion that one exists, and exhibits the divisor of a nonzero Weil differential as such a `W`."* This supplies that, completing Stichtenoth Section I.5.

## The argument

The divisors bounding a fixed nonzero `ω` are stable under suprema, because `A_F(D ⊔ E) = A_F(D) + A_F(E)` — a repartition is assignable place by place to whichever bound it already respects — and their degrees are bounded above, because past the threshold of Riemann's theorem the index of specialty vanishes and with it `Ω_F(D)`. A divisor of maximal degree in that family is therefore its greatest element, places having positive degree.

Multiplication by a function is then an injective `k`-linear map carrying `L(W - D)` onto `Ω_F(D)`: a nonzero `x` has `x · ω ∈ Ω_F(D)` exactly when `ω ∈ Ω_F(D - div x)`, which by maximality says `x ∈ L(W - D)`; surjectivity is Proposition 1.5.9, already available as `exists_repartitionDualMul_eq`. Hence `ℓ(W - D) = dim_k Ω_F(D) = i(D)`, which rearranges to the Riemann–Roch identity.

Essentially all of the work was already on `main` — the local-to-global quotient engine, Riemann's theorem, `dim_k Ω_F(D) = i(D)` (1.5.7), `dim_F Ω_F = 1` (1.5.9), and all of `RiemannRoch/Uniqueness.lean`. What this adds is the last structural step.

## The route does not pass through local components

Stichtenoth builds the divisor of `ω` place by place from its local components, and `Differential/LocalComponent.lean` accordingly records the remaining statements of Section I.7 as blocked for want of that divisor:

> The remaining statements of Section I.7 — that no local component of a nonzero Weil differential vanishes, that `v_P (ω)` is the largest bound its local component respects, and the explicit generator of `Ω_{k(x)}` — need the divisor of a Weil differential, and are not proved here.

[...]

PR #5903, opened 2026-09-06
feat: the Riemann–Roch theorem

sqrt-of-2:
This PR proves the **Riemann–Roch theorem** for an algebraic function field `F/k` with exact constant field, closing the summit of Layer 4 of the AlgebraicCurves roadmap (Stichtenoth I.5): it constructs the divisor of a nonzero Weil differential, proves the duality theorem `L(W − D) ≃ Ω_F(D)` (Theorem 1.5.14), and concludes that this divisor is a Riemann–Roch divisor for the genus (Theorem 1.5.15), so that `TauCeti.Divisor.IsRiemannRochDivisor` — which already carries `ℓ(W) = g`, `deg W = 2g − 2`, the equality of `g₀` with the genus, and linear equivalence of any two such divisors — is inhabited and the canonical class is well defined.

The witness is `(ω)` for `ω ≠ 0`. Enlarging a divisor shrinks `Ω_F(D)`, so the divisors bounding a fixed `ω` are downward closed; the family is stable under suprema, because `A_F(D ⊔ E) = A_F(D) + A_F(E)` (a repartition bounded by `D ⊔ E` respects one of the two bounds at each place separately, so assigning entries to sides splits it), and its degrees are bounded above, because past the threshold of Riemann's theorem the index of specialty vanishes and with it `Ω_F(D)`. A divisor of maximal degree in that family is therefore its greatest element, places having positive degree. Multiplication by a function is then an injective `k`-linear map `F → Ω_F` carrying `L(W − D)` onto `Ω_F(D)`: a nonzero `x` has `x · ω ∈ Ω_F(D)` exactly when `ω ∈ Ω_F(D − div x)`, which by maximality says `x ∈ L(W − D)`; surjectivity is Proposition 1.5.9, already on `main` as `exists_repartitionDualMul_eq`. Hence `ℓ(W − D) = dim_k Ω_F(D) = i(D)`, which rearranges to the Riemann–Roch identity. Essentially all of the machinery was already on `main` — the local-to-global quotient engine, Riemann's theorem, `dim_k Ω_F(D) = i(D)` and `dim_F Ω_F = 1`; what is added is the maximality argument, the duality isomorphism, and the rearrangement.

[...]

PR #5940, opened 2026-09-07
feat(FunctionField): the Riemann–Roch theorem

mrdouglasny:
Prepared by Claude Opus 5, reviewed by @mrdouglasny.

Reopens #5207, which the queue housekeeping retired on 6 September. The three round-one findings were addressed and pushed on 30 August, but the heavy CI never fired on that head — `a0a1c833c` has only `zulip-pr` and `label` check-runs, no `sandboxed-build` and no `build` — so the review never re-ran and the 7-day timer closed it. This is that work, rebased onto current `main` and re-verified.

`Divisor.IsRiemannRochDivisor` already records the Riemann–Roch identity and everything that follows from it — that `g₀` is the genus, that `deg W = 2g - 2` and `ℓ(W) = g`, that any two such divisors are linearly equivalent. Its own docstring names what is missing: *"the Riemann–Roch theorem itself is the assertion that one exists, and exhibits the divisor of a nonzero Weil differential as such a `W`."* This supplies that, completing Stichtenoth Section I.5.

## The argument

The divisors bounding a fixed nonzero `ω` are stable under suprema, because `A_F(D ⊔ E) = A_F(D) + A_F(E)` — a repartition is assignable place by place to whichever bound it already respects — and their degrees are bounded above, because past the threshold of `exists_forall_indexOfSpecialty_eq_zero` the divisor is nonspecial and so bounds no nonzero Weil differential. A divisor of maximal degree in that family is therefore its greatest element, places having positive degree.

Multiplication by a function is then an injective `k`-linear map carrying `L(W - D)` onto `Ω_F(D)`: a nonzero `x` has `x · ω ∈ Ω_F(D)` exactly when `ω ∈ Ω_F(D - div x)`, which by maximality says `x ∈ L(W - D)`; surjectivity is Proposition 1.5.9, already available as `exists_repartitionDualMul_eq`. Hence `ℓ(W - D) = dim_k Ω_F(D) = i(D)`, which rearranges to the Riemann–Roch identity.

Essentially all of the work was already on `main` — the local-to-global quotient engine, Riemann's theorem, `dim_k Ω_F(D) = i(D)` (1.5.7), `dim_F Ω_F = 1` (1.5.9), and all of `RiemannRoch/Uniqueness.lean`. What this adds is the last structural step.

## The route does not pass through local components

Stichtenoth builds the divisor of `ω` place by place from its local components, and `Differential/LocalComponent.lean` accordingly records the remaining statements of Section I.7 as blocked for want of that divisor:

> The remaining statements of Section I.7 — that no local component of a nonzero Weil differential vanishes, that `v_P (ω)` is the largest bound its local component respects, and the explicit generator of `Ω_{k(x)}` — need the divisor of a Weil differential, and are not proved here.

Getting the greatest element from sup-stability and bounded degree instead means `LocalComponent` is **not** in this module's import cone. So I.7 is unblocked by this rather than prerequisite to it — the dependency the module structure anticipated runs the other way. If a local-components-first design was actually intended, that is worth saying now.

## Round-one findings from #5207, carried forward

- **reuse** — the bounded-degree proof reconstructed eventual vanishing of the index of specialty and converted a zero finrank to `Ω_F(D) = ⊥`. It now uses `exists_forall_indexOfSpecialty_eq_zero` and `weilDifferentialFiltration_eq_bot_iff_indexOfSpecialty_eq_zero`; thirteen lines became five.
- **proof-quality** — the inverse-action rewrite chain was duplicated at `Weil.lean:237`. Factored as `repartitionDualMul_inv_repartitionDualMul`, stated where `repartitionDualMul` lives and used in both places.
- **api-design** — `adeleFiltration_sup` now carries `@[simp]`.

## Verification

- Full-library `lake build` on this exact head: **10691 jobs, green** — so neither new `@[simp]` disturbs an existing proof.
- 0 sorries.
- `#print axioms` on `exists_isRiemannRochDivisor`, `isRiemannRochDivisor_of_isGreatest_mem_weilDifferentialFiltration`, `exists_isGreatest_mem_weilDifferentialFiltration`, `adeleFiltration_sup`: `[propext, Classical.choice, Quot.sound]` only.
- `lake exe lint-style` clean on all three files.
- The roadmap's Layer 4 target compiles verbatim against it:

```lean
obtain ⟨W, hW⟩ := exists_isRiemannRochDivisor hF hex
exact ⟨W, hW.degree_eq hF hex, hW.dim_eq hF hex, hW⟩
```

## Roadmap

Advances [`AlgebraicCurves`](https://github.com/TauCetiProject/TauCetiRoadmap/blob/main/TauCetiRoadmap/AlgebraicCurves/README.md), **Layer 4: repartitions, Weil differentials, and Riemann–Roch** — specifically the target at `TauCetiRoadmap/AlgebraicCurves/Suggested.lean:397`, discharged verbatim above. Layer 5's `deg ≥ 2g − 1` target (Stichtenoth 1.5.17) becomes a short derivation from this but is not included; one topic per PR.

Reference: H. Stichtenoth, *Algebraic Function Fields and Codes*, 2nd ed., GTM 254, Springer 2009, Section I.5 — Proposition 1.5.11 and Theorem 1.5.15.

Roadmap: AlgebraicCurves

### DIFF
--- a/TauCeti/FieldTheory/FunctionField/Differential/CanonicalDivisor.lean
+++ b/TauCeti/FieldTheory/FunctionField/Differential/CanonicalDivisor.lean
@@ -1,0 +1,236 @@
+/-
+Copyright (c) 2026 The Tau Ceti contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: The Tau Ceti contributors
+-/
+module
+
+public import TauCeti.FieldTheory.FunctionField.Differential.Dimension
+public import TauCeti.FieldTheory.FunctionField.RiemannRoch.Uniqueness
+
+/-!
+# The divisor of a Weil differential, and the Riemann–Roch theorem
+
+`TauCeti.Divisor.IsRiemannRochDivisor` records what it means for a divisor `W` to satisfy the
+Riemann–Roch identity `ℓ(D) = deg D + 1 - g₀ + ℓ(W - D)`, and everything about such a `W` —
+that `g₀` is the genus, that `deg W = 2g - 2` and `ℓ(W) = g`, that any two are linearly
+equivalent — is proved there.  What is missing, and is supplied here, is that **one exists**.
+That is the Riemann–Roch theorem.
+
+The witness is the divisor of a nonzero Weil differential.  Enlarging a divisor shrinks the space
+`Ω_F(D)` of Weil differentials it bounds, so the divisors bounding a fixed `ω` form a downward
+closed family; the two facts that make it have a *greatest* element are that the family is stable
+under suprema — because `A_F(D ⊔ E) = A_F(D) + A_F(E)`, which is
+`TauCeti.adeleFiltration_sup` — and that its degrees are bounded above, because past the
+threshold of Riemann's theorem the index of specialty vanishes and with it `Ω_F(D)`.
+A divisor of maximal degree in the family is then the greatest, since places have positive degree.
+
+Writing `W` for that greatest divisor, multiplication by a function is an injective `k`-linear map
+`F → Ω_F` carrying `L(W - D)` onto `Ω_F(D)`: a nonzero `x` has `x · ω ∈ Ω_F(D)` exactly when
+`ω ∈ Ω_F(D - div x)`, which by maximality says `D - div x ≤ W`, that is `x ∈ L(W - D)`; and it is
+onto because every Weil differential is a multiple of `ω` (Proposition 1.5.9, already available as
+`TauCeti.exists_repartitionDualMul_eq`).  So `ℓ(W - D) = dim_k Ω_F(D) = i(D)`, which rearranges to
+the Riemann–Roch identity.
+
+This completes Section I.5 of Stichtenoth.  It also supplies the divisor whose absence is noted in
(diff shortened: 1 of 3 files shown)


mrdouglasny:
Closing as superseded by #5903, which merged today.

#5903 does everything this PR does and more — `weilDifferentialDivisor` and `canonicalClass` as actual definitions, the order `v_P(ω)`, the transformation law, the duality exported as a named `k`-linear equivalence, and Proposition 1.6.2 — and it closes Layer 4 outright. Our `adeleFiltration_sup` (with the `@[simp]`) and `repartitionDualMul_inv_repartitionDualMul` are both on `main` as part of it. Nothing here is left to add.

The attribution is correct and I want to say so on the record: #5903 credits the retired #5207 in its PR body, its commit trailer, and an inline comment at the construction it adapts. That is exactly right, and appreciated.

The mistake was ours. #5903 was opened on 6 September, roughly eight hours after #5207 was auto-closed; we opened this PR on the 7th having checked only whether Riemann–Roch had *landed* on `main`, not whether anyone had an open PR for it. Checking open pull requests for the same roadmap target before starting would have caught it, and we will do that in future.

PR #6517, opened 2026-09-13
refactor(Isogeny): restrict any valuation along an isogeny, not just v_∞

CBirkbeck:
Roadmap: EllipticCurves

Layer 0, **Places**. Restricting a valuation along an isogeny's function-field pullback preserves
triviality on the base field. This PR states that once, generally, and has the two places that were
proving it locally cite it.

Provenance: no port.

> **Note.** This description was stale for several rounds and I have rewritten it. It previously
> described an earlier, wider version of this PR — one that added a general computation rule
> `Isogeny.comap_fieldPullback_apply_algebraMap` and deleted the specialization
> `comap_infinityPlace_apply_algebraMap`. **Neither of those is true of this PR any more**: the
> general rule was removed at `3a5a26b3` after `reuse` asked for it twice, and the specialization
> was restored untouched. Rubric findings that referred to a "promised arbitrary-valuation
> computation rule", or to a specialization surviving "alongside the generalized computation
> theorem", were reading that stale text, and the fault is mine.

## What this changes

* **New:** `TauCeti/RingTheory/Valuation/IsTrivialOn.lean` — `Valuation.IsTrivialOn.comap`, an
  instance: the restriction of a valuation trivial on `A` along an `A`-algebra map is trivial on
  `A`. Nothing about isogenies or curves enters.
* `Isogeny/InfinityPlace.lean` **deletes** its bespoke local instance
  `((infinityPlace W₁).comap φ.fieldPullback.toRingHom).IsTrivialOn F`, which the general instance
  now supplies, and imports the new module directly, proof-only.
* `Affine/FunctionField/InfinityPlace/Unique.lean` likewise instantiates the general instance
  instead of re-proving it, and already imported the module directly.
* `Isogeny/FunctionField.lean` only loses the module-doc sentence that advertised a re-export.
  **It does not import the new module.** An earlier revision gave it a `public import` so consumers
  would inherit the instance; `placement` objected that the file uses `IsTrivialOn` in none of its
  own declarations, and that import was removed at `7cdcb527`. Any finding that says this file
  "newly imports" the module, or that a direct import elsewhere is redundant because of a
  re-export here, is reading this description from before that commit — my fault, now corrected.

`Isogeny.comap_infinityPlace_apply_algebraMap` is **not touched**: it is byte-identical to `main`.

## Diff

Four files, +56/−14 against `main`. The only new public declaration is `Valuation.IsTrivialOn.comap`;
everything else is a deletion or a one-line import. No existing statement changes.


TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/InfinityPlace.lean
CBirkbeck, line 1: ### ✅ reuse — now passing on `7cdcb52`.

CBirkbeck, line 1: ### ✅ api-design — now passing on `7cdcb52`.

CBirkbeck, line 1: ### ✅ placement — now passing on `7cdcb52`.

CBirkbeck:
## AI review — approved

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics).

|---|---|---|---|---|
| ✅ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/correctness.md) | approved | `codex/gpt-5.6-sol` | `Valuation.IsTrivialOn.comap` faithfully expresses restriction along an algebra homomorphism: `f.commutes` ensures base elements retain their valuation. Its two uses are non-vacuous, and the removed bespoke instance is semantically subsumed by the general one. |
| ✅ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/reuse.md) | approved | `codex/gpt-5.6-sol` | The generic `Valuation.IsTrivialOn.comap` replaces the two bespoke proofs, and the remaining local instance is a necessary instantiation through `IsScalarTower.toAlgHom`, not a duplicate proof or parallel API. |
| ✅ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/scope.md) | approved | `codex/gpt-5.6-sol` | This is a single-topic extraction and modest generalization of two already-merged valuation-restriction arguments, so roadmap fit is automatic under the refactor policy. |
| ✅ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/attribution.md) | approved | `codex/gpt-5.6-sol` | The generic valuation result credits Wedhorn in the code, and the remaining changes are routine reuse of existing in-repository arguments with no additional attribution requirement. |
| ✅ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/api-design.md) | approved | `codex/gpt-5.6-sol` | The canonical public surface is appropriately minimal: `Valuation.IsTrivialOn.comap` is reusable general infrastructure, while the two specialized users retain no duplicate compatibility API. Its instance form directly supplies the characteristic behavior needed by downstream instance search. |
| ✅ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/generality.md) | approved | `codex/gpt-5.6-sol` | The new comap instance is stated at the natural generality: arbitrary algebra homomorphisms, commutative-semiring base, ring-valued domains, and the monoid structure required by valuations. The affected specialization is removed rather than duplicated. |
| ✅ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/placement.md) | approved | `codex/gpt-5.6-sol` | The generic instance is in its natural valuation module, and both downstream files import it directly where used; the prior unrelated re-export from `Isogeny.FunctionField` is absent. |
| ✅ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/naming.md) | approved | `codex/gpt-5.6-sol` | The introduced `Valuation.IsTrivialOn.comap` name accurately describes the conclusion, and its namespace matches the first explicit argument `v : Valuation C Γ₀`. No notation or other naming issues are introduced. |
| ✅ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/documentation.md) | approved | `codex/gpt-5.6-sol` | The new module and public instance are documented accurately, with clear mathematical motivation and no stale process narrative or unsupported claims. |
| ✅ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics/proof-quality.md) | approved | `codex/gpt-5.6-sol` | The new proof is concise and robust, and the documented local instance cleanly reuses the generalized result without brittle rewriting or undocumented definitional equality. |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/6660c16a29f69a54c1745602722a3bbbd7e2080b...7cdcb5273e8373de03a97de3ceb43cb1bfea1bbe) at head `7cdcb52`. rubrics @ [`afb424e`](https://github.com/TauCetiProject/TauCetiReview/tree/afb424eda89e8ac96d9eb69f6a88972055a4cd1b/rubrics). Review spend: $5.81.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/InfinityPlace.lean
CBirkbeck: Contesting, with the measurement — and the placement half of the earlier round is fixed, which is why `placement` and `reuse` are green this round.

`comap_fieldPullback_apply_algebraMap` cannot be `@[simp]`: both ingredients of its proof already are, so `simp` carries its left-hand side to its right-hand side unaided. On the head under review:

```
theorem comap_fieldPullback_apply_algebraMap … :
    (v.comap φ.fieldPullback.toRingHom) (algebraMap W₂.CoordinateRing W₂.FunctionField c)
      = v (φ.pullback c) := by
  simp          -- closes the goal
```

The two rules doing it are `Valuation.comap_apply` (Mathlib, `@[simp]`) and `fieldPullback_algebraMap` (this file, line 173, `@[simp]`). Tagging this theorem would add a rule whose left-hand side `simp` never sees in that form, which is what `simpNF` rejects.

I have taken the half of the observation that is right: the proof is now the one-line `simp` rather than a four-lemma `rw` chain, and the reason for the missing annotation is recorded beside the declaration in the house idiom. The name is kept because `InfinityPlace.lean` rewrites by it three times and the module docstring advertises it as the computation rule the places API is stated through.

CBirkbeck: **Re: your reply on `api-design` —** re-reviewed on `11944ea`; this clears the finding ✅ — approved on `11944ea`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3998680776.</sub>

CBirkbeck: **Contesting in part: one of the three consumers rewrites `←`, which is the case `simp` cannot cover.**

The finding (rubric `reuse`, round 3, head `11944eacf`) reads:

> `TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean:330` — `comap_fieldPullback_apply_algebraMap` merely bundles two existing `@[simp]` lemmas and adds a parallel rewrite API. _Fix:_ Delete the theorem and let its three consumers use `simp`/`rw` with `Valuation.comap_apply` and `fieldPullback_algebraMap` directly.

The premise is right and I established it myself last round — the proof *is* just `by simp`, which is exactly why I declined the `api-design` request to mark it `@[simp]` (that would be a simpNF violation) and recorded the reason in the comment above it. But "its proof is `simp`" and "its consumers can use `simp`" come apart here, because of the direction of use.

Of the three consumers in `Isogeny/InfinityPlace.lean`:

* `:114` — `rwa [hu, comap_fieldPullback_apply_algebraMap] at h` — forward. `simp` would do.
* `:138` — forward, inside a `rw` chain. `simp` would do.
* `:81` — **`rw [← comap_fieldPullback_apply_algebraMap φ _, …]`** — backward.

That third site is the load-bearing one. `one_lt_infinityPlace_pullback_X` states `1 < infinityPlace W₁ (φ.pullback x₂)` and rewrites it *into* the comap form precisely so the next line, `set u := (infinityPlace W₁).comap φ.fieldPullback.toRingHom with hu`, can abstract the restricted valuation and run the rest of the proof (`hpoly`, `hy`, `hcr`) against an opaque `u`. `simp` has no backward mode, so the fix line as written does not reach it; the replacement would have to be

```lean
rw [← Valuation.comap_apply, ← fieldPullback_algebraMap]
```

which is two backward rewrites whose order matters, the first of which (`← Valuation.comap_apply`) matches `v (f x)` for *any* valuation and ring hom and so is markedly more fragile than the single named rewrite it replaces. Bundling two `@[simp]` lemmas under a name is standard practice exactly when the bundle is needed in the non-simp direction.

There is also a design point: this file's module doc (`InfinityPlace.lean:50`) names this theorem as *the computation rule the other results are stated through*, and the file is deliberately ordered computation-rule-first, equivalence-second. It is the API the places layer rewrites by, which is what the comment on the declaration records.

**What I will do if you still want it gone:** I am happy to delete it and inline at all three sites, taking the two backward rewrites at `:81`. Say so and I will push that. What I do not want to do is delete it on the stated ground that the consumers can use `simp`, because at `:81` they cannot — and if that site is then written with `← Valuation.comap_apply`, the "parallel rewrite API" has not been removed, only spelled out in a more brittle form at the call site.

CBirkbeck: **Withdrawing my contest — I measured it and I was wrong. Fixed in `1d8dd13`.**

My previous reply argued that the consumer at `InfinityPlace.lean:81` rewrites backwards, that `simp` has no backward mode, and that the replacement would therefore need "two backward rewrites whose order matters, the first of which (`← Valuation.comap_apply`) matches `v (f x)` for *any* valuation and ring hom". I then actually tried it. Results, each a full `lake build` of the module:

| replacement at `:81` | result |
|---|---|
| `rw [← Valuation.comap_apply, ← fieldPullback_algebraMap, …]` | fails — *did not find an occurrence of the pattern* |
| `rw [← fieldPullback_algebraMap, ← Valuation.comap_apply, …]` | fails — same |
| `rw [← Valuation.comap_apply, …]` | fails — same |
| **`rw [← fieldPullback_algebraMap, …]`** | **builds** |

So it is *one* backward rewrite, not two, and the ordering problem I described does not arise: `Valuation.comap_apply` is definitional at that position, so no rewrite by it is needed — which is also why every spelling that mentions it fails. The premise of my objection was simply false.

`1d8dd13` therefore does what the finding asked: deletes `comap_fieldPullback_apply_algebraMap` and inlines at all three consumers.

One detail worth recording, since it is the grain of truth in the original bundling. The two forward sites could *not* be done with a hand-picked `simp only [Valuation.comap_apply, fieldPullback_algebraMap]` — that leaves `φ.fieldPullback.toRingHom (…)` unreduced and reports `fieldPullback_algebraMap` as unused, because the coercion layers `AlgHom.toRingHom_eq_coe` and `RingHom.coe_coe` have to normalize first. Plain `simp` handles them, so those sites are now `simpa [hu, Valuation.mem_integer_iff] using h` and `simpa [IsScalarTower.algebraMap_apply …] using one_lt_infinityPlace_pullback_X φ`. That is exactly your fix line's "use `simp`", and it reads better than what it replaced.

Net −16 lines. Full `lake build` of the library green on the pin `30a58f79` (11281 jobs, zero Mathlib recompiles); no line over 100 characters. `instIsTrivialOnComapFieldPullback` is untouched — that generalization is the half of this PR the places API needs, and nothing here disturbs it.

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean
CBirkbeck, line 1: ### ✅ generality — now passing on `7cdcb52`.

CBirkbeck, line 1: ### ✅ proof-quality — now passing on `7cdcb52`.

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/InfinityPlace.lean
CBirkbeck: **Re: your reply on `reuse` —** re-reviewed on `11944ea`; the finding stands — The new computation theorem duplicates the existing simp API rather than adding a distinct result.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3998724253.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean
CBirkbeck: **Accepted — fixed in `e80d088`.**

The finding (rubric `generality`, round 1, head `11944eac`) is right:

> `…/Isogeny/FunctionField.lean:321` — `instIsTrivialOnComapFieldPullback` uses no property of `Isogeny` or `fieldPullback` beyond being an `F`-algebra hom, so its API is at an unnecessarily specialized level. _Fix:_ Prove the generic result for restricting a valuation along an arbitrary `F`-algebra hom (at the weakest compatible ring assumptions), then obtain this instance by specialization.

`Valuation.IsTrivialOn.comap` now lives in a new `TauCeti/RingTheory/Valuation/IsTrivialOn.lean`:

```lean
instance IsTrivialOn.comap {A B C : Type*} [CommSemiring A] [Ring B] [Ring C] [Algebra A B]
    [Algebra A C] (f : B →ₐ[A] C) (v : Valuation C Γ₀) [v.IsTrivialOn A] :
    (v.comap f.toRingHom).IsTrivialOn A where
  eq_one a ha := by simpa using Valuation.IsTrivialOn.eq_one (v := v) a ha
```

On the ring assumptions: this is at `LinearOrderedCommMonoidWithZero Γ₀`, which is what the `IsTrivialOn` class itself requires, rather than the `LinearOrderedCommGroupWithZero` the isogeny version had assumed — the group structure was never used.

**One deviation from the fix line:** I did not keep the specialization. Instance search finds the general instance directly at all three former call sites in `InfinityPlace.lean` (the build confirms it), so a specialized restatement would be exactly the duplication `reuse` objects to elsewhere on this PR. The elliptic instance is therefore deleted outright rather than re-derived.

I checked Mathlib first: it defines `Valuation.IsTrivialOn` (`RingTheory/Valuation/Basic.lean:682`) and `Valuation.comap`, and its only `IsTrivialOn` instances are for finite fields and `inftyValuation`. There is no instance relating the two, so this is new rather than a restatement.

Worth recording: two further places in this repository derive the same fact inline — `FieldTheory/FunctionField/Place/Extension/Basic.lean:252` and `.../InfinityPlace/Unique.lean:132`. Both restrict along a bare `algebraMap` rather than an `AlgHom`, so they do not match this instance's shape and I have deliberately left them alone rather than widen this PR; the file's `## Provenance` section records it.

Full `lake build` green (11282 jobs, zero Mathlib recompiles); `scripts/lint-env.sh` PASS, no new violations.

CBirkbeck: **Accepted — fixed in `e80d088`, by the same change that answered `generality`.**

The finding (rubric `proof-quality`, round 1, head `11944eac`):

> `…/Isogeny/FunctionField.lean:325` — `instIsTrivialOnComapFieldPullback.eq_one` depends on four representation-level rewrites, including two coercion lemmas, making the proof sensitive to Mathlib normalization changes. _Fix:_ Replace the rewrite chain with a robust `simpa` from `Valuation.IsTrivialOn.eq_one c hc`; if simplification cannot bridge the algebra-map coercions, add or use a dedicated lemma.

Correct, and the brittleness had a cause worth naming: the chain

```lean
rw [Valuation.comap_apply, AlgHom.toRingHom_eq_coe, RingHom.coe_coe, AlgHom.commutes]
```

is three representation-level steps wrapped around one mathematical step (`AlgHom.commutes`, that an algebra map fixes the base). It was only needed because the statement had been pinned to `φ.fieldPullback`, which forces the `toRingHom`/`coe` layers into the goal.

Generalizing to an arbitrary `f : B →ₐ[A] C` removes the reason for all three, and the proof becomes exactly the `simpa` you asked for:

```lean
  eq_one a ha := by simpa using Valuation.IsTrivialOn.eq_one (v := v) a ha
```

No dedicated bridging lemma was needed — `simp` normalizes the algebra-map coercions on its own. The declaration this finding is against no longer exists; see the `generality` thread for why the specialization was deleted rather than re-derived from the general instance.

TauCeti/RingTheory/Valuation/IsTrivialOn.lean
CBirkbeck, line 1: ### ✅ documentation — now passing on `7cdcb52`.

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/InfinityPlace.lean
CBirkbeck: **Re: your reply on `reuse` —** re-reviewed on `e80d088`; this clears the finding ✅ — approved on `e80d088`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3998763163.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean
CBirkbeck: **Re: your reply on `generality` —** re-reviewed on `e80d088`; this clears the finding ✅ — approved on `e80d088`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3998800682.</sub>

CBirkbeck: **Re: your reply on `proof-quality` —** re-reviewed on `e80d088`; this clears the finding ✅ — approved on `e80d088`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3998800720.</sub>

TauCeti/RingTheory/Valuation/IsTrivialOn.lean
CBirkbeck: **Accepted — fixed in `270901b`.**

> `TauCeti/RingTheory/Valuation/IsTrivialOn.lean:26` — The `## Provenance` section records port status, Mathlib API gaps, prior duplicate implementations, and a scope decision; this is coordination history that will become stale rather than mathematical usage documentation. _Fix:_ Remove the provenance section from the module docstring; retain only the mathematical motivation, main result, and reference.

Removed. The docstring now carries only the motivation, `## Main results` and `## References`.

You are right about the content, and specifically about the staleness: the section named two other files as carrying inline duplicates of this instance, which is a claim that stops being true the moment either is touched, and it recorded a scope decision about *this PR* rather than anything a reader of the module needs.

For the record, since it is the reason the section existed: I added it to satisfy the submission checklist I work from, which asks for "Main results + References + Provenance" on every new module. Having now looked at the sibling modules, `## Provenance` appears in only 5 of the 17 files in `TauCeti/RingTheory/Valuation/`, and in each of those it names an external source that was ported — `Descent.lean`'s is the clearest example. That is a genuine provenance claim; mine was repository history wearing the same heading. I will keep the section for ports and leave it off original modules.

utensil:
🔍 Review in progress — `codex` reviewing `270901b486ee`.

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean
utensil, line 1: ### 🟡 reuse — request_changes  `codex/gpt-5.6-sol`

The generic `IsTrivialOn.comap` instance leaves an existing specialization reimplemented by hand.

- `TauCeti/AlgebraicGeometry/EllipticCurve/Affine/FunctionField/InfinityPlace/Unique.lean:132` — The private instance re-proves the new public `Valuation.IsTrivialOn.comap` for the rational-function inclusion. _Fix:_ Import `TauCeti.RingTheory.Valuation.IsTrivialOn`, delete this instance, and instantiate the generic result locally with `IsScalarTower.toAlgHom F (RatFunc F) W.FunctionField` where needed.

<sub>`codex/gpt-5.6-sol` · 170s · 492.1k in / 5.7k out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/65a04559f0cc692abebf14d055494cda36cc14e7...270901b486eee32be93b9025143570f4548b53ca) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/reuse.md)</sub>

TauCeti/RingTheory/Valuation/IsTrivialOn.lean
utensil, line 1: ### 🟡 naming — request_changes  `codex/gpt-5.6-sol`

The new instance is placed under `Valuation` but exposes an algebra homomorphism as its first explicit argument, violating the namespace/dot-notation rule.

- `TauCeti/RingTheory/Valuation/IsTrivialOn.lean:36` — `Valuation.IsTrivialOn.comap` has `f : B →ₐ[A] C` before `v`, so its first explicit argument is an `AlgHom` despite the `Valuation` namespace; this prevents the namespace from matching the receiver as required by the naming convention. _Fix:_ Swap the explicit arguments to `(v : Valuation C Γ₀) (f : B →ₐ[A] C)`, preserving the name while making the valuation the receiver and matching the expression order `v.comap f`.

<sub>`codex/gpt-5.6-sol` · 138s · 342.6k in / 4.5k out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/65a04559f0cc692abebf14d055494cda36cc14e7...270901b486eee32be93b9025143570f4548b53ca) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/naming.md)</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/InfinityPlace.lean
utensil, line 1: ### 🟡 proof-quality — request_changes  `codex/gpt-5.6-sol`

The main proof still silently relies on definitional equality across valuation and homomorphism wrappers.

- `TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/InfinityPlace.lean:81` — `set u := …` abstracts the rewritten expression only because `Valuation.comap` unfolds through `toRingHom` coercions; this undocumented representation-level dependency is fragile. _Fix:_ Introduce `u` and an explicit local evaluation equality proved using `Valuation.comap_apply` and `fieldPullback_algebraMap`, then rewrite through that equality without restoring the deleted public theorem.

<sub>`codex/gpt-5.6-sol` · 132s · 182.6k in / 4.3k out tokens · reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/65a04559f0cc692abebf14d055494cda36cc14e7...270901b486eee32be93b9025143570f4548b53ca) · [rubric](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/proof-quality.md)</sub>

utensil:

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics).

|---|---|---|---|---|
| ✅ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/correctness.md) | approved | `codex/gpt-5.6-sol` | The generalized `Valuation.IsTrivialOn.comap` instance faithfully proves preservation of base-field triviality, and the revised infinity-place proofs retain their original mathematical content. |
| 🟡 | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/reuse.md) | changes requested | `codex/gpt-5.6-sol` | The generic `IsTrivialOn.comap` instance leaves an existing specialization reimplemented by hand. |
| ✅ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/scope.md) | approved | `codex/gpt-5.6-sol` | This is a coherent refactor and modest generalization of already-merged place-restriction machinery; the generic valuation instance directly replaces the specialized isogeny instance and remains within one topic. |
| ✅ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/attribution.md) | approved | `codex/gpt-5.6-sol` | The generalized instance is routine and credits Wedhorn; the relocated isogeny material retains the existing Silverman, roadmap, and AINTLIB attribution. No missing central source was found. |
| ✅ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/api-design.md) | approved | `codex/gpt-5.6-sol` | The canonical surface is appropriately generalized, publicly placed, and sufficient for valuation restriction without retaining redundant isogeny-specific compatibility declarations. |
| ✅ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/generality.md) | approved | `codex/gpt-5.6-sol` | The new `Valuation.IsTrivialOn.comap` instance is stated at the natural generality: arbitrary commutative-semiring base, ring algebras, algebra homomorphism, valuation, and ordered value monoid, with no visibly unused or excessive assumptions. |
| ✅ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/placement.md) | approved | `codex/gpt-5.6-sol` | The generic comap instance is in the canonical valuation module, and the isogeny function-field re-export directly supports valuation restriction along `fieldPullback`. No misplaced declaration or evidently wrong import remains. |
| 🟡 | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/naming.md) | changes requested | `codex/gpt-5.6-sol` | The new instance is placed under `Valuation` but exposes an algebra homomorphism as its first explicit argument, violating the namespace/dot-notation rule. |
| ✅ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/documentation.md) | approved | `codex/gpt-5.6-sol` | The new module and public instance have accurate module-level and declaration-level documentation, and the updated existing module docstring accurately describes the re-exported API. |
| 🟡 | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics/proof-quality.md) | changes requested | `codex/gpt-5.6-sol` | The main proof still silently relies on definitional equality across valuation and homomorphism wrappers. |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/65a04559f0cc692abebf14d055494cda36cc14e7...270901b486eee32be93b9025143570f4548b53ca) at head `270901b`. rubrics @ [`153ea8a`](https://github.com/TauCetiProject/TauCetiReview/tree/153ea8a6febe59f492c2f3d48471733e8b550848/rubrics). Review spend: $4.40.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/InfinityPlace.lean
utensil: **Re: your reply on `reuse` —** re-reviewed on `270901b`; the finding stands — The generic `IsTrivialOn.comap` instance leaves an existing specialization reimplemented by hand.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3998763163.</sub>

utensil: **Re: your reply on `api-design` —** re-reviewed on `270901b`; this clears the finding ✅ — approved on `270901b`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3998680776.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean
utensil: **Re: your reply on `generality` —** re-reviewed on `270901b`; this clears the finding ✅ — approved on `270901b`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3998800682.</sub>

TauCeti/RingTheory/Valuation/IsTrivialOn.lean
utensil: **Re: your reply on `documentation` —** re-reviewed on `270901b`; this clears the finding ✅ — approved on `270901b`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3998903165.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean
utensil: **Re: your reply on `proof-quality` —** re-reviewed on `270901b`; the finding stands — The main proof still silently relies on definitional equality across valuation and homomorphism wrappers.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3998800720.</sub>

TauCeti/RingTheory/Valuation/IsTrivialOn.lean
CBirkbeck: **Accepted — fixed in `6ccee70`.**

> `TauCeti/RingTheory/Valuation/IsTrivialOn.lean:36` — `Valuation.IsTrivialOn.comap` has `f : B →ₐ[A] C` before `v`, so its first explicit argument is an `AlgHom` despite the `Valuation` namespace … _Fix:_ Swap the explicit arguments to `(v : Valuation C Γ₀) (f : B →ₐ[A] C)`.

Swapped. You are right on both counts: the first explicit argument now matches the namespace, and it also matches the expression order `v.comap f`, which the old order inverted for no reason.

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean
CBirkbeck: **Accepted — fixed in `6ccee70`, and this corrects an error of mine.**

> `.../InfinityPlace/Unique.lean:132` — The private instance re-proves the new public `Valuation.IsTrivialOn.comap` for the rational-function inclusion. _Fix:_ Import `TauCeti.RingTheory.Valuation.IsTrivialOn`, delete this instance, and instantiate the generic result locally with `IsScalarTower.toAlgHom F (RatFunc F) W.FunctionField` where needed.

Done, and thank you for the `IsScalarTower.toAlgHom` pointer specifically. When I introduced the generic instance I looked at this site and at `FieldTheory/FunctionField/Place/Extension/Basic.lean:252`, concluded that both "restrict along a bare `algebraMap` rather than an `AlgHom`, so they do not match this instance's shape", and deliberately left them alone. That was wrong — `IsScalarTower.toAlgHom` is exactly the bridge, and the hand-rolled proof was redundant.

One deviation from the fix line, made explicit in a comment: the declaration does not disappear, it becomes

```lean
local instance : (v.comap (algebraMap (RatFunc F) W.FunctionField)).IsTrivialOn F :=
  Valuation.IsTrivialOn.comap v (IsScalarTower.toAlgHom F (RatFunc F) W.FunctionField)
```

because instance search will not see the `AlgHom` through a bare `algebraMap` on its own. So what remains is a one-line instantiation of the general result rather than a second proof of it, which is what the finding was about. `Place/Extension/Basic.lean:252` is the same shape and the same fix; I have left it out of this PR only to keep the diff to the module under review, and noted it.

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/InfinityPlace.lean
CBirkbeck: **Accepted — fixed in `6ccee70`.**

> `.../Isogeny/InfinityPlace.lean:81` — `set u := …` abstracts the rewritten expression only because `Valuation.comap` unfolds through `toRingHom` coercions; this undocumented representation-level dependency is fragile. _Fix:_ Introduce `u` and an explicit local evaluation equality proved using `Valuation.comap_apply` and `fieldPullback_algebraMap`, then rewrite through that equality without restoring the deleted public theorem.

Done exactly as written. `u` is introduced first, then

```lean
have hueval : ∀ c : W₂.CoordinateRing,
    u (algebraMap W₂.CoordinateRing W₂.FunctionField c) = infinityPlace W₁ (φ.pullback c) :=
  fun c ↦ by rw [hu, Valuation.comap_apply, AlgHom.toRingHom_eq_coe, RingHom.coe_coe,
    fieldPullback_algebraMap]
```

and both places that previously leaned on the definitional unfolding now rewrite through `hueval` — the opening rewrite of the goal, and `hmem`, which had been `simpa [hu, Valuation.mem_integer_iff]` and is now an explicit `rw` through it. Nothing public is restored; `hueval` is local to the one proof that needs it, which is the distinction the earlier `reuse` finding drew.

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean
CBirkbeck, line 1: ### ✅ correctness — now passing on `7cdcb52`.

CBirkbeck: **Accepted — restored in `69a31fa`.**

> `.../Isogeny/FunctionField.lean:34` — `Isogeny.comap_fieldPullback_apply_algebraMap` is not declared; the computation rule is instead recreated as the local `hueval` inside one infinity-place proof, leaving arbitrary restricted valuations without the promised API. _Fix:_ Add the general theorem in `FunctionField.lean` for any valuation `v`, then rewrite the infinity-place proof through it.

Done exactly that. The theorem is back in `FunctionField.lean`, general in the valuation, and both `one_lt_infinityPlace_pullback_X` and `hmem` rewrite through it; the local `hueval` is gone.

I should say plainly where the deletion came from, because it was not an oversight: `reuse` (round 3, head `11944eac`) asked for it — *"Delete the theorem and let its three consumers use `simp`/`rw` with `Valuation.comap_apply` and `fieldPullback_algebraMap` directly"* — and I complied after checking that the consumers really could. So the two findings point opposite ways at the same declaration, and I have taken yours. Reasons, in order of weight:

1. A block outranks a request-changes.
2. Your framing is about the **API**, not about this PR's consumers, and that is the stronger reading. `reuse` was right that the three call sites here can inline; it does not follow that an arbitrary restricted valuation should have no computation rule.
3. It matches where this file is going. The next module in this chain needs precisely this rule for a *point* valuation rather than the valuation at infinity — the `infinityPlace`-specialized version it replaces cannot serve that, which is the whole reason this PR generalizes.

One measurement, since it bears on how much was actually at stake: on `main`, all five occurrences of `comap_infinityPlace_apply_algebraMap` are inside `Isogeny/InfinityPlace.lean` itself — one doc bullet, the declaration, and three uses. So nothing outside that file lost access when it was deleted. That does not make the deletion right, but it does mean the regression was latent rather than breaking, which is worth recording accurately.

Restoring it also preserves what `proof-quality` asked for in the same round: the representation-level dependency is now a named theorem proved once, rather than a definitional unfolding relied on silently at `set u := …`.

CBirkbeck: **Reversing the fix I made for this finding, because `correctness` blocked on its absence. Flagging it here rather than quietly undoing it.**

This thread (rubric `reuse`, round 3, head `11944eac`) asked:

> `.../Isogeny/FunctionField.lean:330` — `comap_fieldPullback_apply_algebraMap` merely bundles two existing `@[simp]` lemmas and adds a parallel rewrite API. _Fix:_ Delete the theorem and let its three consumers use `simp`/`rw` with `Valuation.comap_apply` and `fieldPullback_algebraMap` directly.

I did that, after measuring that the backward-rewriting consumer really could manage with `rw [← fieldPullback_algebraMap]`.

`correctness` (rubric `correctness`, round 1, head `6ccee70a`, ⛔ **block**) then said the opposite about the same declaration:

> `.../Isogeny/FunctionField.lean:34` — `Isogeny.comap_fieldPullback_apply_algebraMap` is not declared … leaving arbitrary restricted valuations without the promised API. _Fix:_ Add the general theorem in `FunctionField.lean` for any valuation `v`, then rewrite the infinity-place proof through it.

I have restored it (`69a31fa`) and taken `correctness`, on three grounds: a block outranks a request; the two findings are asking different questions, yours about this PR's three call sites and `correctness`'s about the API available to *any* restricted valuation; and the next module in this chain needs the rule for a point valuation, which is what this PR generalizes it for.

Your underlying observation stands and I am not disputing it — the proof is still just `simp`, the theorem still is not `@[simp]` (that would be a simpNF violation, as established earlier in this PR), and the comment above it now says so and says what it is for. If you still want it gone, the thing to resolve is the disagreement with `correctness` rather than the code, and I would rather that be visible than have me oscillate between the two rounds.

CBirkbeck: **Re: your reply on `correctness` —** re-reviewed on `69a31fa`; this clears the finding ✅ — approved on `69a31fa`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3999098337.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/InfinityPlace.lean
CBirkbeck: **Re: your reply on `reuse` —** re-reviewed on `69a31fa`; the finding stands — The generalized triviality instance is reusable, but the new valuation computation theorem duplicates existing rewrite machinery.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3999098397.</sub>

TauCeti/RingTheory/Valuation/IsTrivialOn.lean
CBirkbeck: **Re: your reply on `naming` —** re-reviewed on `69a31fa`; this clears the finding ✅ — approved on `69a31fa`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3999043024.</sub>

CBirkbeck: **Re: your reply on `documentation` —** re-reviewed on `69a31fa`; this clears the finding ✅ — approved on `69a31fa`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3998903165.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean
CBirkbeck: **Re: your reply on `proof-quality` —** re-reviewed on `69a31fa`; this clears the finding ✅ — approved on `69a31fa`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3999043116.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/InfinityPlace.lean
CBirkbeck: **Holding the code as it is, because `correctness` blocks on the opposite. This needs adjudication between the two rubrics, not another push.**

The finding re-raised on `69a31fa` is:

> The generalized triviality instance is reusable, but the new valuation computation theorem duplicates existing rewrite machinery.

`correctness` (round 1, head `6ccee70a`, ⛔ **block**) says the opposite about the same declaration:

> `Isogeny.comap_fieldPullback_apply_algebraMap` is not declared … leaving arbitrary restricted valuations without the promised API. _Fix:_ Add the general theorem in `FunctionField.lean` for any valuation `v`, then rewrite the infinity-place proof through it.

I have now been round this loop once in each direction: deleted the theorem for `reuse` at `11944eac`, restored it for `correctness` at `69a31fa`. Pushing again would only move the block from one rubric to the other, so I am stopping rather than oscillating.

Where I think the disagreement actually lies, for whoever resolves it: your objection is that the proof is `simp` and this PR's three call sites can inline it, which I verified is true. `correctness`'s objection is about what an *arbitrary* restricted valuation has available, independent of this PR's call sites. Both can be right at once; they are not answering the same question.

Two facts that may help settle it:

* On `main`, all five occurrences of the specialization this replaces, `comap_infinityPlace_apply_algebraMap`, are inside `Isogeny/InfinityPlace.lean` itself. So nothing outside that file ever consumed it, and `correctness`'s regression is latent rather than breaking.
* The next module in this chain needs the rule for a **point** valuation rather than the valuation at infinity. That consumer does not exist yet, so it cannot be cited as a present consumer — but it is the reason the PR generalizes rather than merely moves.

If the decision is that it should go, say so here and I will delete it and take the `correctness` block instead; I would rather that be an explicit call than something I flip on my own.

CBirkbeck: **Re: your reply on `reuse` —** re-reviewed on `69a31fa`; the finding stands — The generalized triviality instance is reusable, but the valuation computation theorem remains a duplicate composite rewrite API.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3999186743.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean
CBirkbeck: **Accepted — fixed in `abca8a7`.**

> The generic triviality instance is at the natural level, but the generalized computation theorem still imposes unnecessary group structure on the valuation codomain.

Right, and it was an inconsistency of mine rather than a considered choice: I weakened `Valuation.IsTrivialOn.comap` to `[LinearOrderedCommMonoidWithZero Γ₀]` when that finding was raised earlier in this PR, but left `comap_fieldPullback_apply_algebraMap` at `[LinearOrderedCommGroupWithZero Γ]` — which it inherited from the `infinityPlace`-specialized version it replaced. `Valuation` itself only requires the monoid, and the proof uses nothing more. Both declarations now sit at the same level.

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/InfinityPlace.lean
CBirkbeck: **Still holding, and the picture has changed in a way worth recording: `correctness` now explicitly approves this code.**

Round 4, on the same head `69a31fa`:

* `correctness` — **approve**. It had been a ⛔ block on this theorem's *absence*; restoring it satisfied it.
* `reuse` — this finding, restated: "the valuation computation theorem remains a duplicate composite rewrite API."

So the two rubrics have now each evaluated the same declaration on the same commit and reached opposite conclusions in the same round. That is no longer a sequencing artefact of my having pushed in between; it is a live disagreement about whether a named composite rewrite rule should exist when its proof is `simp`.

I am not pushing on it a third time. My position, for whoever adjudicates:

* Your objection is sound on its own terms — the proof is `simp`, and I verified earlier in this PR that each of the three call sites here can inline it, including the backward one via `rw [← fieldPullback_algebraMap]`.
* `correctness`'s objection is about the API an arbitrary restricted valuation is left with, not about these three call sites, and deleting the theorem is what triggered its block.
* Neither of you is wrong about the facts. What is needed is a ruling on whether "provable by `simp`" is sufficient grounds to deny a rule a name, given that the specialization it replaces (`comap_infinityPlace_apply_algebraMap`) *did* have a name on `main` and this PR's stated purpose is to generalize it.

I have taken the other finding you raised this round — the codomain is now `LinearOrderedCommMonoidWithZero`, in `abca8a7`.

CBirkbeck: **This is already done — the declaration was deleted three commits ago, and it no longer exists anywhere in the repository.**

> **`reuse` — request_changes, round 6**:
> `TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean:330` — `comap_fieldPullback_apply_algebraMap` merely bundles existing `Valuation.comap_apply` and `fieldPullback_algebraMap`; its proof is `by simp`, creating a parallel normalization theorem. _Fix:_ Delete the theorem and use `simp` at forward consumers and `rw [← fieldPullback_algebraMap]` at the reverse consumer.

I agreed with this in an earlier round and carried it out. `3a5a26b3` ("narrow this PR to the uncontested generalization") deleted the theorem and restored `main`'s `comap_infinityPlace_apply_algebraMap` and its call sites untouched, exactly as this fix asks — the reverse consumer now uses `rw [← fieldPullback_algebraMap]`, which I measured as sufficient before making the change.

Three checks on the current head `3a5a26b`, any one of which settles it:

1. `grep -rn "comap_fieldPullback_apply_algebraMap" TauCeti/` — **no matches anywhere in the repository.**
2. `TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean` is **321 lines long**, so the cited `:330` does not exist. Line 321 is `end TauCeti`.
3. Grepping the very diff this finding links (`65a04559...3a5a26b3`) for that identifier returns nothing.

This PR's whole diff against `main` is now 4 files, +56/-14: the new `TauCeti/RingTheory/Valuation/IsTrivialOn.lean`, one public import in `Isogeny/FunctionField.lean`, and the two call sites in `Isogeny/InfinityPlace.lean` and `InfinityPlace/Unique.lean` that instantiate `Valuation.IsTrivialOn.comap` instead of re-proving it. Nothing in it bundles `Valuation.comap_apply`.

One likely cause worth flagging, since it would affect other rounds too: this round's `base_sha` is `65a04559`, which is a stale `main`. Diffing against it pulls in unrelated churn — `TauCeti/Probability/Distributions/Wishart.lean` alone shows 344 deletions, and `MeasureTheory/Measure/SymmetricMatrix/Rank.lean` 70 — none of which this PR touches. The `merge_base_sha` recorded alongside it is a different commit, `8e673113`.

Nothing to change here. If something in the current 4-file diff does bundle existing rewrites, point at it and I will fix that instead.

CBirkbeck: **Re: your reply on `reuse` —** re-reviewed on `fd939e0`; this clears the finding ✅ — approved on `fd939e0`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3999451191.</sub>

CBirkbeck: **Contesting: the replacement this fix names does not exist, and it does not exist because `reuse` required its deletion — which `reuse` approves on this very head.**

> **`api-design` — request_changes, round 7**, head `fd939e0c`:
> `.../Isogeny/InfinityPlace.lean:78` — `Isogeny.comap_infinityPlace_apply_algebraMap` survives even though `comap_fieldPullback_apply_algebraMap` is its fully general canonical replacement. _Fix:_ Delete the specialization and rewrite its three in-repository uses with `comap_fieldPullback_apply_algebraMap`.

The fix is not executable as written. On this head, `grep -rn "comap_fieldPullback_apply_algebraMap" TauCeti/` returns **no matches anywhere in the repository** — there is nothing to rewrite the three uses with. (The three uses are real: `InfinityPlace.lean:90`, `:123`, `:147`.)

It is absent because a sibling rubric required that:

> **`reuse` — request_changes, round 6**:
> `.../Isogeny/FunctionField.lean:330` — `comap_fieldPullback_apply_algebraMap` merely bundles existing `Valuation.comap_apply` and `fieldPullback_algebraMap`; its proof is `by simp`, creating a parallel normalization theorem. _Fix:_ **Delete the theorem** and use `simp` at forward consumers and `rw [← fieldPullback_algebraMap]` at the reverse consumer.

I carried that out in `3a5a26b3`, and **`reuse` is ✅ approved on this same round-7 board**. So `api-design` is asking me to keep a declaration that `reuse` has twice asked me to delete and has now approved the absence of.

And the other half of the fix is blocked from the third side:

> **`correctness` — ⛔ block, round 1**, head `6ccee70a`:
> `.../Isogeny/FunctionField.lean:34` — `Isogeny.comap_fieldPullback_apply_algebraMap` is not declared; the computation rule is instead recreated as the local `hueval` inside one infinity-place proof, leaving arbitrary restricted valuations without the promised API.

That is why `comap_infinityPlace_apply_algebraMap` is present and byte-identical to `main`: this PR does not touch it. Deleting it is what `correctness` blocked on at round 1, and `correctness` is ✅ on this head too.

So satisfying this finding requires re-adding the theorem `reuse` demanded be deleted, and then deleting the one `correctness` blocked on the absence of — closing a three-rubric loop. I am not going to push either half and let the other two re-fire.

**What this PR actually is**, for the record — 4 files, +56/−14 after the merge of current `main`: the new `TauCeti/RingTheory/Valuation/IsTrivialOn.lean` with the general `Valuation.IsTrivialOn.comap`; one `public import` in `Isogeny/FunctionField.lean` so consumers get it; and two call sites (`Isogeny/InfinityPlace.lean`, `InfinityPlace/Unique.lean`) that now instantiate that instance instead of re-proving it locally. It **deletes** a bespoke `IsTrivialOn` instance rather than adding a parallel one. `comap_infinityPlace_apply_algebraMap` is untouched.

If `api-design` is upheld over both `reuse` and `correctness`, say so explicitly and name which of the two you want reopened, and I will make that change in one commit.

CBirkbeck: **Re: your reply on `api-design` —** re-reviewed on `d3f1a6f`; this clears the finding ✅ — approved on `d3f1a6f`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3999564384.</sub>

CBirkbeck: **Withdrawing my diagnosis from the previous round. You were reading a stale PR description, and that was my fault, not a tooling fault.**

Last round I argued this finding cited a declaration that "does not exist anywhere in the repository" and suggested the cause was a stale `base_sha` in the review metadata. The first half was true — `comap_fieldPullback_apply_algebraMap` really is absent — but the explanation was wrong, and I have found the actual cause.

**The PR description still described the original, wider version of this PR.** It promised

> `Isogeny.comap_fieldPullback_apply_algebraMap` — **any** valuation of the source function field, restricted along an isogeny and evaluated on the image of the target's coordinate ring, is that valuation of the pullback.

and asserted

> `Isogeny.comap_infinityPlace_apply_algebraMap` is **deleted**

Neither has been true since `3a5a26b3`, where I narrowed the PR — removing the general rule after `reuse` asked for it twice, and restoring the specialization untouched. I changed the code and never updated the description. So this round's ⛔ `correctness` ("the promised arbitrary-valuation computation rule is absent") is **correct on its own terms**: the PR did promise it and does not deliver it. And your finding here — the specialization surviving "alongside the generalized computation theorem" — follows from the same text, which is why it named a theorem that is not in the diff.

I have rewritten the description to say what the PR actually is: one new instance `Valuation.IsTrivialOn.comap`, one `public import` so consumers get it, and two call sites that now instantiate it instead of re-proving it locally — including a bespoke `IsTrivialOn` instance in `Isogeny/InfinityPlace.lean` that the PR **deletes**. `comap_infinityPlace_apply_algebraMap` is byte-identical to `main`.

No code change accompanies this, because I do not believe any is needed: on the actual diff there is no generalized computation theorem for the specialization to duplicate. If you disagree once the description matches the code, say so against the current text and I will act on it.

My apologies for the misdirection — I should have checked the description against the diff before blaming the metadata.

CBirkbeck: **Re: your reply on `api-design` —** re-reviewed on `d3f1a6f`; this clears the finding ✅ — approved on `d3f1a6f`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3999638227.</sub>

TauCeti/AlgebraicGeometry/EllipticCurve/Isogeny/FunctionField.lean
CBirkbeck: **Re: your reply on `generality` —** re-reviewed on `d3f1a6f`; this clears the finding ✅ — approved on `d3f1a6f`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 3999204839.</sub>

PR #7243, opened 2026-09-17
refactor: generalize diagonal_mul_single_mul_diagonal to non-unital entries

harahu:
`TauCeti.diagonal_mul_single_mul_diagonal` bound `[Semiring A]`:

```lean
diagonal v * single i j c * diagonal w = single i j (v i * c * w j)
```

Neither side mentions `1`, and both the matrix product and the entry product
are written left-associated, so nothing in the statement needs associativity
either. The `Semiring` was entirely an artifact of *how the proof was written*:
it opened with `Matrix.mul_assoc`, which requires `NonUnitalSemiring` entries,
and then needed `mul_assoc` again on the entries to finish.

Reordering the rewrite removes both. Applying `Matrix.mul_diagonal` before
`Matrix.diagonal_mul` produces `v a * (single i j c) a b * w b`, already
left-associated exactly as the statement writes it, so the off-diagonal case
closes by `mul_zero`/`zero_mul` and the diagonal case by `rfl`-level `simp`.
The lemma then holds over any `NonUnitalNonAssocSemiring`.

### What was checked beyond the binder

* This is the natural level, reached by fixing the proof rather than by
  stepping down a class at a time. Weakening only to `NonUnitalSemiring` also
  compiles against the original proof, but that still asks for an associativity
  the statement never uses. `NonUnitalNonAssocSemiring` is the weakest class in
  which `diagonal`, `single` and matrix multiplication are all available.
* `[Fintype m]` and `[Fintype n]` both remain load-bearing — one for each of the
  two matrix products — as do the two `DecidableEq`s, needed by `diagonal` and
  `single`.
* Placement is unchanged; the file exists for this single lemma.
* Both consumers, `LinearAlgebra/Matrix/GeneralLinearGroup/Transvection.lean`
  and `Algebra/AlgebraicGroup/GeneralLinear/Adjoint/Basic.lean`, were built and
  are unaffected.

### The import narrowing, and the `placement` finding

The `placement` rubric asked for the file's `public import
Mathlib.LinearAlgebra.Matrix.Transvection` to be narrowed. That change was
deliberately split off as #7253 under the one-topic rule, which is what the
first reply in that thread said. The rubric's answer was that a separate open PR
does not fix *this* head, which is fair — the gate judges the head commit.

#7253 has since merged, and this branch is now rebased onto a `main` that
contains it, so the narrow `public import Mathlib.Data.Matrix.Basis` is in place
and the diff here is once again only the binder and the proof. Nothing was
duplicated between the two PRs.

Roadmap: ReductiveGroups


TauCeti/LinearAlgebra/Matrix/Diagonal.lean
sqrt-of-2, line 1: ### ✅ generality — now passing on `0be823a`.

sqrt-of-2, line 1: ### ✅ placement — now passing on `4aabbbd`.

sqrt-of-2:
## AI review — approved

Each rubric is judged independently by multiple review agents; the PR merges only once **every** rubric is green — any rubric that is not green (changes requested, blocked, errored, stale, or not yet run) blocks the merge. See the [rubrics](https://github.com/TauCetiProject/TauCetiReview/tree/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics).

|---|---|---|---|---|
| ✅ | [correctness](https://github.com/TauCetiProject/TauCetiReview/blob/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics/correctness.md) | approved (carried from `0be823a`, patch unchanged) | `codex/gpt-5.6-sol` | The generalized assumptions remain sufficient, and the reordered rewrites establish the stated left-associated product without using a unit or associativity. |
| ✅ | [reuse](https://github.com/TauCetiProject/TauCetiReview/blob/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics/reuse.md) | approved (carried from `0be823a`, patch unchanged) | `codex/gpt-5.6-sol` | The PR only generalizes an existing TauCeti declaration and introduces no new duplicated API or re-derived scaffolding. |
| ✅ | [scope](https://github.com/TauCetiProject/TauCetiReview/blob/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics/scope.md) | approved (carried from `0be823a`, patch unchanged) | `codex/gpt-5.6-sol` | This is a single-topic generalization of an already-merged lemma, so roadmap fit is automatic under the refactor policy. |
| ✅ | [attribution](https://github.com/TauCetiProject/TauCetiReview/blob/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics/attribution.md) | approved (carried from `0be823a`, patch unchanged) | `codex/gpt-5.6-sol` | The change is a routine proof and typeclass generalization with no identifiable external formal or informal source requiring attribution. |
| ✅ | [api-design](https://github.com/TauCetiProject/TauCetiReview/blob/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics/api-design.md) | approved (carried from `0be823a`, patch unchanged) | `codex/gpt-5.6-sol` | The generalized typeclass bound preserves a minimal, canonical public theorem statement without exposing implementation details or introducing duplicate or compatibility-only API. |
| ✅ | [generality](https://github.com/TauCetiProject/TauCetiReview/blob/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics/generality.md) | approved (carried from `0be823a`, patch unchanged) | `codex/gpt-5.6-sol` | The declaration now sits at the natural Mathlib matrix-API level; weakening it further would require locally re-deriving the multiplication lemmas rather than improving the usable abstraction. |
| ✅ | [placement](https://github.com/TauCetiProject/TauCetiReview/blob/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics/placement.md) | approved | `codex/gpt-5.6-sol` | The declaration remains in its natural matrix-diagonal module, and the previously overbroad transvection import is no longer present in the current diff. |
| ✅ | [naming](https://github.com/TauCetiProject/TauCetiReview/blob/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics/naming.md) | approved (carried from `0be823a`, patch unchanged) | `codex/gpt-5.6-sol` | The PR introduces no new declaration names or notation; the existing theorem name continues to describe its equality accurately, and its namespace remains appropriate. |
| ✅ | [documentation](https://github.com/TauCetiProject/TauCetiReview/blob/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics/documentation.md) | approved (carried from `0be823a`, patch unchanged) | `codex/gpt-5.6-sol` | The existing theorem documentation remains accurate under the generalized typeclass assumptions; the PR adds no misleading or process-oriented documentation. |
| ✅ | [proof-quality](https://github.com/TauCetiProject/TauCetiReview/blob/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics/proof-quality.md) | approved (carried from `0be823a`, patch unchanged) | `codex/gpt-5.6-sol` | The revised proof is short, readable, and robust: it avoids unnecessary associativity rewrites and does not rely on undocumented definitional-equality manipulation. |

<sub>Reviewing [this diff](https://github.com/TauCetiProject/TauCeti/compare/eff57fb0b2b24121fd95b6cbad1a1d2e84a2635f...4aabbbdceb567ff29825480240e86793f9a2b827) at head `4aabbbd`. rubrics @ [`9e06df9`](https://github.com/TauCetiProject/TauCetiReview/tree/9e06df966f1e3d476f8b8035518d386ecd1fad0f/rubrics). Review spend: $1.09.</sub>

TauCeti/LinearAlgebra/Matrix/Diagonal.lean
harahu: Contesting this one, on the reuse cost rather than on the class being wrong.

**The suggested class is right about the floor.** Both halves were tested, and both are load-bearing:

* `[AddMonoid A]` in place of `[AddCommMonoid A]` — the *statement* stops elaborating (`failed to synthesize`). Matrix multiplication is a `Finset.sum`, and Mathlib's instance is `instance [Fintype n] [Mul α] [AddCommMonoid α] : Mul (Matrix n n α)`. Commutativity is not optional.
* `[Mul A] [Zero A]` in place of `[MulZeroClass A]` — the statement elaborates but the identity becomes *false*: the off-diagonal terms are `0 * single i j c k b`, and without `zero_mul` they do not vanish.

So `[AddCommMonoid A] [MulZeroClass A]` is a genuine, non-vacuous weakening and is itself minimal. That is not what I am contesting.

**What it costs is the reuse this PR was approved for.** Mathlib develops the whole of matrix multiplication one class higher:

* `Matrix.diagonal_mul` and `Matrix.mul_diagonal` sit under `variable [NonUnitalNonAssocSemiring α]` in `Mathlib/Data/Matrix/Mul.lean`;
* the entire `single`-times-multiplication section of `Mathlib/Data/Matrix/Basis.lean` sits under `variable [Fintype m] [NonUnitalNonAssocSemiring α]`;
* there is no `Matrix.diagonal_mul_single` or `single_mul_diagonal` to fall back on.

Neither of those two lemmas actually uses distributivity, so this is an upstream over-ask — but it is upstream, and TauCeti cannot restate them without duplicating Mathlib API, which `CLAUDE.md` forbids.

I tried the suggested direct proof. Every version that typechecks at `[AddCommMonoid A] [MulZeroClass A]` has to reduce the double sum by hand and, in doing so, reprove `diagonal_mul` and `mul_diagonal` inline — literally `have : (diagonal d * M) a b = d a * M a b` and its mirror — because those are exactly the steps that are missing below `NonUnitalNonAssocSemiring`. That turns a six-line proof which reuses two Mathlib lemmas into a dense local re-derivation of them.

That is the scaffolding the `reuse` rubric approved this PR for not introducing: "The PR only generalizes an existing TauCeti declaration and introduces no new duplicated API or re-derived scaffolding." Satisfying this finding as written would falsify that sentence. The two rubrics cannot both be satisfied here, and I would rather keep the reuse.

**The mathematical case is thin in the other direction too.** An `AddCommMonoid` carrying a zero-annihilating multiplication that does *not* distribute is not a structure matrix theory is developed over; matrices over it are not a semiring, so even `diagonal v * single i j c * diagonal w` is only well-defined by the left-association convention. `NonUnitalNonAssocSemiring` is where Mathlib's matrix API lives and where callers of this lemma actually are.

Happy to be overruled — if the preference is to take the weaker class and accept the inline re-derivation, say so and I will push it.

harahu: Contesting: the fix is right, but it is already a separate open PR, deliberately.

This finding asks for exactly the change made by **#7253, "refactor: narrow the matrix-unit import to Data.Matrix.Basis"** — which is open and mergeable, and whose entire diff is:

```diff
--- `Matrix.diagonal`, `Matrix.single`, and matrix multiplication occur in the statement below.
--public import Mathlib.LinearAlgebra.Matrix.Transvection
+public import Mathlib.Data.Matrix.Basis
```

That PR was split off on purpose, and this PR's description says so under "Noted, deliberately not bundled":

> The file carries `public import Mathlib.LinearAlgebra.Matrix.Transvection` with a comment saying it is there for `Matrix.diagonal`, `Matrix.single` and matrix multiplication — but `Matrix.single` actually lives in `Mathlib.Data.Matrix.Basis`, and Transvection is a much heavier, field-oriented module. Narrowing that import looks right, especially now that the lemma is non-unital, but it is a separate topic and is left for its own PR.

`CLAUDE.md` requires it: "One topic per PR. Ship a prerequisite refactor as its own PR." This PR changes a typeclass binder and a proof; the import narrowing is an independent change to an independent line, with its own measurement (the dependency closure drops from 1559 `lake build` jobs to 1201) and its own hazard analysis of the two downstream importers. Bundling it here would also put two open PRs in direct textual conflict on the same import line, so whichever merged second would need a rebase for no benefit.

If #7253 merges first, this finding disappears from #7243 on rebase. If #7243 merges first, #7253 still applies cleanly. Either order works; merging the change into both does not.

I would rather not duplicate it here — but if the preference is to land the import narrowing in this PR and close #7253, say so and I will do that instead.

sqrt-of-2: **Re: your reply on `generality` —** re-reviewed on `0be823a`; this clears the finding ✅ — approved on `0be823a`.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 4039322756.</sub>

sqrt-of-2: **Re: your reply on `placement` —** re-reviewed on `0be823a`; the finding stands — The declaration is naturally placed, but the current PR head still exposes an unnecessarily broad, higher-level dependency; a separate open PR does not correct this revision.

<sub>`codex/gpt-5.6-sol` · addresses your replies through comment 4039324712.</sub>

PR #7722, opened 2026-09-20
refactor: generalize valuation centres and isolate Dedekind comparisons

harahu:
`Valuation.centerIdeal_ne_bot` no longer requires an explicit `IsDomain R` or `Nontrivial R`: its existing algebra map into the field `K` supplies nontriviality inside the proof. The centre API accepts any linearly ordered commutative group with zero as its valuation target; its proofs do not require discreteness. The generic centre construction and its lemmas now live in `TauCeti/RingTheory/Valuation/Center.lean`; `DedekindDomain/AdicValuation/Basic.lean` retains the four Dedekind-specific comparison theorems.

The place-centre API uses the same weaker assumptions: `Place.center`, both membership characterizations, the residue-kernel formula, and the lies-over theorem drop unnecessary Dedekind instances. Membership is proved directly using the generic centre lemmas. Assumptions are introduced in appropriate sections; the three affected affine-model files use no `omit` directives. Module and declaration documentation distinguish a nonzero prime ideal bundled as `HeightOneSpectrum R` from the height-one interpretation available for Dedekind domains.

The pinned Mathlib's `HeightOneSpectrum` takes only `R` and `CommRing R`. Removing the unused surrounding instances therefore requires no `nolint`, allowlist change, or dependency bump. This PR does not depend on #7738 or #7791.

Validation on the latest commit:
- The changed modules and downstream callers through `Different.Localization` and `Different.Tower` build with the repository's standard options.
- All 15 default environment linters pass on all five changed modules.
- The full-library source style/header check and `git diff --check` pass.
- Full sandboxed CI is pending on this revision; it passed on revision `675b74166`. No self-review was run for this revision.

Roadmap: none

PR #7738, opened 2026-09-20
refactor: drop IsDomain from centerIdeal_ne_bot entirely

harahu:
**Needs human review: this PR intentionally touches `scripts/`.** The allowlist line is part of
the deliverable, not an obstacle — please do not strip it to make the PR auto-mergeable.

This is the complete version of the weakening in #7722. That PR only got halfway, because
finishing it requires a human-owned file; the reviewers there asked for the full thing, so here it
is separately rather than blocking the partial improvement behind human review.

## The change

`Valuation.centerIdeal_ne_bot` asked for `[IsDomain R]`. The surrounding section already fixes
`[CommRing R] [Field K] [Algebra R K]`, so `algebraMap R K` is a ring hom into a nontrivial ring
and `RingHom.domain_nontrivial` supplies the `Nontrivial R` the proof actually uses. The binder
comes out entirely rather than being weakened to `Nontrivial R` as in #7722.

## Two knock-on effects, and why each needs what it needs

Both come from `IsDedekindDomain.HeightOneSpectrum` being declared `@[nolint unusedArguments]`
upstream — the Dedekind instance is a *phantom* parameter of the type. Until now
`[IsDedekindDomain R]` was doing real work discharging `centerIdeal_ne_bot`'s `IsDomain R`.

1. `unusedSectionVars` reports it unused in `asIdeal_heightOneSpectrum`. **`omit` cannot help
   here**: the statement names `HeightOneSpectrum R`, so without the instance in scope it does not
   elaborate at all. The linter's other suggestion, restructuring the `variable` declarations, is
   the applicable one, so the instance moves off the `Comparison` section line onto its six
   declarations.
2. `unusedArguments` then reports it on `heightOneSpectrum` itself — the same phantom parameter
   one level up, and this time genuinely unremovable. That is what `@[nolint unusedArguments]`
   is for, mirroring exactly what Mathlib does for the structure, plus the declaration-level entry
   in `scripts/lint-nolints-allowlist.txt` that `lint-env` requires for every `nolint` under
   `TauCeti/`.

## Relationship to #7722

The two overlap on the same line: #7722 weakens `[IsDomain R]` to `[Nontrivial R]` and stays
auto-mergeable; this one removes it. Whichever lands first, the other needs a trivial rebase. If
you would rather have only one, this is the better end state and #7722 can be closed — it exists
so that the partial improvement is not blocked on human review turnaround.

## Verification

Full `lake build`, `scripts/lint-style.sh`, and `lint-env` all pass — the last reporting
`PASS — no new violations (58 grandfathered, 12 ratchetable)`.

A note on that: `scripts/lint-env.sh` cannot run on macOS as written. Lines 358 and 363 use
GNU-only `\|` BRE alternation, so BSD `sed` returns an empty match and the script aborts with
`internal error: unrecognized selected linter set ''`. I verified the above with a local throwaway
copy using `-E` and `(base|tacticAlt)`, which I did not commit. There is already an upstream branch
named `fix/lint-env-posix-bre`, so this is presumably known, but it does mean the gate is
CI-only for macOS contributors — which is how #7722's failure got past me in the first place.

Roadmap: none


harahu:
Superseded by #7722, which removes the redundant assumptions and generalizes the valuation-centre API without a nolint annotation or allowlist change. The alternate approach in this PR is no longer needed.

